Issue #21, opened 2022-05-27
Functions should have a section about their hooks


tellyworth:
### Discussed in https://github.com/WordPress/wporg-developer/discussions/13

<div type='discussions-op-text'>

<sup>Originally posted by **dd32** May 23, 2022</sup>
The only way to see a list of filters/actions present within a function is to expand the code block and read through.

There should be a dedicated list of filters (and their short description) present on the function page, linking to the hooks info page.</div>

Ian also notes:

> It'd be really helpful if that section also listed the earliest hook at which you can call the function (when applicable). Also any other pre-conditions.

StevenDufresne:
**Huddle Chat Notes:**
- We have the data for hooks (since we have a "hooks" section in the docs).
- Getting the order is a bit more challenging.

dd32:
> The only way to see a list of filters/actions present within a function is to expand the code block and read through.

It turns out this is wrong. the page DOES display the hooks, just not in a super obvious way. They're included in the "Related" section, the highlighted items here are a filter and action:

<img width="874" alt="Screen Shot 2022-05-30 at 4 56 17 pm" src="https://user-images.githubusercontent.com/767313/170934773-2c670bb6-7f37-4e6f-8948-be6f474b10b8.png">

That section is a bit noisy, we don't need to know that the function uses `__()` or `do_action` surely (See: all the X'd items).. but I never realised hooks were listed in there, as it's not exactly clear at all.

iandunn:
>  I never realised hooks were listed in there, as it's not exactly clear at all

Doh, me neither. 

> That section is a bit noisy, we don't need to know that the function uses __() or do_action surely (See: all the X'd items)

💯

dd32:
It might be worth closing this, and opening a new issue for the remaining part of the issue:

> It'd be really helpful if that section also listed the earliest hook at which you can call the function (when applicable). Also any other pre-conditions.

iandunn:
Opened #115 for the remaining task

PR #37, opened 2022-05-31
Hide commonly-used functions from the uses section.

dd32:
As noted in #21, the "Uses" section can get rather noisy when it includes functions that are used by, in some cases, 1,000 other functions.

This PR removes any function from that section that is used by more than 50 other functions. An arbitrary number I chose.

Example with https://developer.wordpress.org/reference/functions/wp_authenticate/

| Before | After |
| --- | --- |
| Screen Shot 2022-05-31 at 1 11 49 pm  | Screen Shot 2022-05-31 at 1 12 09 pm  | 

Leaving this as a PR as I'm not 100% sure on it. It might make more sense to keep them or sort it differently.. or.. make the functions clickable within the code source-view instead (like GitHub)


tellyworth:
I think as long as the hooks themselves are displayed somewhere (whether in this list or a separate one) then this is a definite improvement.

I guess as an alternative the commonly-used functions could be pushed to the bottom below a Show More link.

dd32:
> I think as long as the hooks themselves are displayed somewhere (whether in this list or a separate one) then this is a definite improvement.

Well, this isn't for removing the hooks from the table (That's already happened), just common WordPress functions - such as `__()` or `apply_filters()` which is used by nearly every function.

> I guess as an alternative the commonly-used functions could be pushed to the bottom below a Show More link.

That seems like a good middle-ground. Move all of the commonly-used functions to the bottom of the list, and hide them by default. Let me play with that.

tellyworth:
My last commit partially implements the move-to-the-bottom idea. It returns a count of the number to display, but the JS still needs to be fixed (restored?) for the show/hide part to work.

dd32:
> but the JS still needs to be fixed (restored?) for the show/hide part to work.

@tellyworth I've updated the JS to respect the count from PHP, appears to work now. 

| Default | Expanded |
| --- | --- |
| Screen Shot 2022-06-09 at 3 51 24 pm | Screen Shot 2022-06-09 at 3 51 28 pm |

Issue #20, opened 2022-05-27
Function parameter data types should be easier to see.


tellyworth:
### Discussed in https://github.com/WordPress/wporg-developer/discussions/9

<div type='discussions-op-text'>

<sup>Originally posted by **StevenDufresne** May 23, 2022</sup>
Parameter data types are not given enough visual priority.

There are plenty of ways to up the visual priority, here's just one example.
| Before | After |
| --- | --- |
| ![](https://d.pr/i/J91AcZ.png) | ![](https://d.pr/i/c0HYE4.png) |</div>

@ryelle gave a great example from MDN:
![](https://user-images.githubusercontent.com/541093/170580231-ade6fa27-d537-4eeb-9abe-45861f864cfc.png)

outdoor2kode:
Another docs example that might be worth referring to. 
https://docs.docker.com/engine/api/v1.41/#operation/ImageList

StevenDufresne:
I'll leave this issue open, in case anyone wants to make some changes or improve on what I merged in #31.

ryelle:
Now that the parameter names use `<code>`  it can get a little visually repetitive/confusing if the parameter description also has code formatting in it:

<img width="681" alt="Screen Shot 2022-05-31 at 6 51 55 PM" src="https://user-images.githubusercontent.com/541093/171296753-005b960b-506c-4e89-850f-f372b8cc9829.png">

On the huddle today, we talked about maybe using the blue styling of the parameter from the heading

<img width="237" alt="Screen Shot 2022-05-31 at 6 53 46 PM" src="https://user-images.githubusercontent.com/541093/171296911-0469f709-128c-49ba-87a4-b7dbda18173a.png">

StevenDufresne:
For me, the screenshot isn't too visually confusing... I think I've become accustomed to `<code>` heavy texts over the years of reading developer documentation. There are probably more dramatic ones lurking in the code reference though.

Give it a try!

dd32:
I'm closing this as I think we've achieved the goal intended.

PR #39, opened 2022-05-31
Add return syntax highlighting

dd32:
See #20 

This:
 - adds highlighting of types to the return data
 - adds the return-type to the function declaration, mimicking the PHP docs.

| Before | After |
| --- | --- |
| Screen Shot 2022-05-31 at 3 23 44 pm  | Screen Shot 2022-05-31 at 3 23 37 pm  |


iandunn:
Can we rely on `WP_User` always being linked in the `@return` description? It might be safer to keep the link in the `WP_User|false` bit

StevenDufresne:
Does moving the string on to a new line make it more readable?

Issue #40, opened 2022-05-31
Improve search UX


dd32:
While there's some [ideas to improve the search](https://github.com/WordPress/wporg-developer/discussions/14), there probably a few smaller things we can do without replacing it entirely.

- [x]  When typing, the auto-complete dropdown offers exact matches, selecting one of these however just fills the search area, it doesn't submit the form nor does it go to the selected definition.
- [x]  The search button is not immediately obvious, with the search-menu-open icon being larger and more obvious. (This could be just me..)
- [x]  The search icon used on developer is the reverse of the global menu icon direction.

Search clicking (bonus included confusion of where to click) [(mp4)](https://user-images.githubusercontent.com/767313/171106834-8aef4ce7-8ab9-44c8-a51d-bb408292fb57.mov)
<img src="https://user-images.githubusercontent.com/767313/171107964-c3a9b0f5-113c-4cf9-b33e-784df8f20e7d.gif">

Menu icons:
<img width="346" alt="Screen Shot 2022-05-31 at 4 28 55 pm" src="https://user-images.githubusercontent.com/767313/171107227-2989a442-d225-412a-a9b1-1e0ee510d255.png">

iandunn:
> it doesn't submit the form 

➕ , that's always bugged me

> The search button is not immediately obvious

➕ . I'm not sure what a better location would be though; the sidebar?

dd32:
> I'm not sure what a better location would be though; the sidebar?

I'm not sure either. The Global header uses a search button/icon in the same location, but has a focus state that makes it clear the input ends before the icon, which makes it more obvious it might be the button:
<img width="311" alt="Screen Shot 2022-06-01 at 2 17 20 pm" src="https://user-images.githubusercontent.com/767313/171326587-4f1a0670-791f-42de-9bc4-ca02165667f5.png">

outdoor2kode:
Changes made:
* Search Icon is aligned with the global menu icon direction.
* Make the focus state clear - the input ends before the icon.
* Change the placeholder from `Search...` to `Search in Code Reference...`.

Alternative ways to make the search button in the search bar obvious:
a) Always show the search bar and remove the bigger search icon (search-menu-open icon)

![image](https://user-images.githubusercontent.com/18050944/171497821-e0d46f83-cd45-40e3-9a7c-3ed172504543.png)

Or change the bigger search icon to another icon or text, such as:

b) **Arrow up**
[...]

PR #41, opened 2022-05-31
Search UX improvements

dd32:
See #40 

 - Click/enter auto-complete item to go directly to it's page
 - Search/Enter on exact-match in text field still goes to search results
 - Archive and search count results in a more expected place (IMHO)

| Before | After |
| --- | --- |
| Instant load selected auto-complete item |
|  |  |
| Search result count between form |
| Screen Shot 2022-05-31 at 5 48 02 pm | Screen Shot 2022-05-31 at 5 44 11 pm |
| Pagination links not hiding single-page |
| Screen Shot 2022-05-31 at 5 50 07 pm |  Screen Shot 2022-05-31 at 5 51 06 pm |


ryelle (approved):
Works much more like I expect it to, now :)

PR #44, opened 2022-05-31
Switch code highlighting to use Code Syntax Block

### LINKED ISSUE #25: Code blocks should look like code blocks.
### Discussed in https://github.com/WordPress/wporg-developer/discussions/5

# Next Step

- Create PR with updates to code block styles 
  - Should probably have a background
  - Line numbers could probably be removed (especially for non source blocks)
 

<sup>Originally posted by **StevenDufresne** May 22, 2022</sup>
It's typical for code blocks to be given more contrast with the document to indicate that you are looking at code. It helps users identify code quickly when scanning a document.

| Ours| Others|
| --- | --- |
|   |   |
|  |  |

 </div>

See https://github.com/WordPress/wporg-developer/discussions/5#discussioncomment-2830291 for notes on style

ryelle:
This removes the use of SyntaxHighlighter Evolved in favor of rendering using the core code block + enhancements from the Code Syntax Block (using prismjs to do syntax highlighting). I've made a semi-custom theme for the developer site, based on a combination of the existing theme, the a11y-dark CSS, and the core WordPress color system.

These are known issues and I think it would be best to turn these into separate issues once this PR is merged, rather than trying to do all of it in one:
 
- [ ] Turn on block editor for Explanations
- [ ] Convert shortcodes in Explanations to code blocks
- [ ] Add back in the collapsible toggle when code is longer than 12 lines
- [ ] PR upstream to Code Syntax Block to allow a custom line number start (added in code for the `source`, but not possible to add in the editor)
- [ ] Add `css` button to the user notes comment form, since that shortcode exists now

Fixes #25

## Screenshots

The function source:

| Before | After |
| --- | --- |
| source-before | source-after |

Code in the "User Contributed Notes" section - note that I added a `css` shortcode to mirror the PHP & JS ones.

| Before | After |
| --- | --- |
| user-notes-before | user-notes-after |

Code added in the Explanation section. Currently breaking due to `wpautop`, but if we convert this to the real block (instead of the shortcode workaround) it should fix that issue.

| Before | After |
| --- | --- |
| explanation-before | explanation-after |

### DIFF
--- a/source/wp-content/themes/wporg-developer/scss/_colors.scss
+++ b/source/wp-content/themes/wporg-developer/scss/_colors.scss
@@ -1,0 +1,86 @@
+//Colors
+
+$color-white: #fff;
+$color-black: #000;
+
+// Color map of all of the hex values.
+// Please keep this map in sync with the HTML
+$colors: (
+	white: #fff,
+	black: #000,
+	gray-0: #f6f7f7,
+	gray-2: #f0f0f1,
+	gray-5: #dcdcde,
+	gray-10: #c3c4c7,
+	gray-20: #a7aaad,
+	gray-30: #8c8f94,
+	gray-40: #787c82,
+	gray-50: #646970,
+	gray-60: #50575e,
+	gray-70: #3c434a,
+	gray-80: #2c3338,
+	gray-90: #1d2327,
+	gray-100: #101517,
+	blue-0: #f0f6fc,
+	blue-5: #c5d9ed,
+	blue-10: #9ec2e6,
+	blue-20: #72aee6,
+	blue-30: #4f94d4,
+	blue-40: #3582c4,
+	blue-50: #2271b1,
+	blue-60: #135e96,
+	blue-70: #0a4b78,
+	blue-80: #043959,
+	blue-90: #01263a,
+	blue-100: #00131c,
+	red-0: #fcf0f1,
+	red-5: #facfd2,
+	red-10: #ffabaf,
+	red-20: #ff8085,
+	red-30: #f86368,
+	red-40: #e65054,
+	red-50: #d63638,
+	red-60: #b32d2e,
+	red-70: #8a2424,
+	red-80: #691c1c,
+	red-90: #451313,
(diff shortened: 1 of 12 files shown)


source/wp-content/themes/wporg-developer/scss/_colors.scss
```diff
@@ -0,0 +1,86 @@
+//Colors
```
ryelle, line 1: I added this file, which is from [the Core WP Colors](https://codepen.io/ryelle/full/WNGVEjw), but could probably remove it since I'm not using many colors from it after all. Any opinions?

tellyworth:
We could also remove `syntaxhighlighter` from `composer.json` and `.wp-env.json` if it's definitely not needed elsewhere. But that could be a separate PR.

iandunn (approved):
Works for me 👍🏻 

I noticed that on `http://localhost:8888/reference/classes/_wp_editors/`, the first part isn't highlighted. I'm guessing we'll have quirks like that no matter which highlighter we use though.

[I like Weston's fork of the plugin](https://github.com/WordPress/wporg-developer/issues/25#issuecomment-1142385670) a bit better, but it's not a blocker

outdoor2kode (approved):

outdoor2kode:
> Works for me 👍🏻
> 
> I noticed that on `http://localhost:8888/reference/classes/_wp_editors/`, the first part isn't highlighted. I'm guessing we'll have quirks like that no matter which highlighter we use though.
> 
> [I like Weston's fork of the plugin](https://github.com/WordPress/wporg-developer/issues/25#issuecomment-1142385670) a bit better, but it's not a blocker

Just found that it is how prismjs works when there's a `<?php` inside a php file:

**With <?php**
Screen Shot 2022-06-02 at 1 40 39 AM

**Without <?php**
image

Ref: 
1. http://localhost:8888/reference/classes/_wp_editors/ L970
2. https://prismjs.com/test.html#language=php

Tried to find a workaround in its repo, but got nothing.

PR #46, opened 2022-05-31
Always show heading link indicator

### LINKED ISSUE #33: Add anchor indicator
### Discussed in https://github.com/WordPress/wporg-developer/discussions/26

<sup>Originally posted by **iandunn** May 28, 2022</sup>
Our pages [already have anchors for deep-linking to sections](https://developer.wordpress.org/reference/classes/wp_query/#properties-and-methods), but it's helpful to show a visual indicator of that.

Example: https://api.slack.com/reference/surfaces/formatting#escaping

Screen Shot 2022-05-27 at 1 20 16 PM
</div>

iandunn:
Only showing it on hover can results in folks not knowing it's there. Hovering isn't a UI concept on mobile devices either, which makes the problem worse. IMO, hovering is best used for design enhancements rather than functionality.

See https://ux.stackexchange.com/questions/20866/ui-hover-focus-to-reveal-controls
See https://ux.stackexchange.com/questions/82015/how-do-users-know-to-hover-over-elements

Fixes #33


dd32:
Personally I'm partial to not showing it, as it's visual clutter to always have it present IMHO.

It's worth noting that this was implemented via https://meta.trac.wordpress.org/ticket/6171 after the lengthy https://make.wordpress.org/docs/2022/02/20/the-hashtag-and-its-future-in-documentation-articles/ 

It's a hybrid of the [Github style (where it only shows the anchor on hover, but also only the anchor is the link)](https://docs.github.com/en/actions/learn-github-actions/understanding-github-actions) and improving accessibility by linking the entire heading.

outdoor2kode (approved):
👍

ryelle:
I added two follow-up commits to handle changes here:

1. https://github.com/WordPress/wporg-developer/commit/a517ab17cf3f0f2a15c8f77d73c79878571e2e5e to update the composer lockfile. When I ran  `composer update`, I had to let composer delete the handbook files, then restore them via git.
2. https://github.com/WordPress/wporg-developer/commit/d2060892add88a6be095d4a5dd3af525a2882793 to add the handbook plugin to the exceptions in the `.gitignore`. I'm not totally sure how the files were added without this.

Was the handbook plugin added here because this will be its new source of truth? or is that just temporary to iterate on the sprint faster?

iandunn:
> Was the handbook plugin added here because this will be its new source of truth? or is that just temporary to iterate on the sprint faster?

I wasn't thinking it'd live here canonically, it just wasn't possible to PR the change w/out a local copy

Issue #35, opened 2022-05-31
Embed handler is broken on class pages


tellyworth:
For example https://developer.wordpress.org/reference/classes/atomfeed/embed/

Expected: it should work like https://developer.wordpress.org/reference/functions/apply_filters/embed/

Are there other page types with broken embeds also?

- [ ] fix class embed
- [ ] fix method embed
- [ ] backport to production/canonical theme

ryelle:
Methods on a class are also not working correctly — this is on the PR #59.

<img width="620" alt="Screen Shot 2022-06-02 at 12 21 38 PM" src="https://user-images.githubusercontent.com/541093/171677543-0058eff4-6a7c-4ff9-ae6e-50405b1e750e.png">

Example on production shows a different error: https://developer.wordpress.org/reference/classes/custom_background/admin_load/embed/

PR #59, opened 2022-06-02
Update the embed format.

### LINKED ISSUE #34: Improve function reference embeds
### Discussed in https://github.com/WordPress/wporg-developer/discussions/18

<sup>Originally posted by **tellyworth** May 27, 2022</sup>
Here's what it looks like when I embed a function reference page on a p2:
Screen Shot 2022-05-27 at 4 09 22 pm

It would be much more useful to display some extra info, notably the function parameters and return type.</div>

StevenDufresne:
Closes: https://github.com/WordPress/wporg-developer/issues/34

This PR updates the embed layout by creating an `embed.php` file and serving a custom embed view. 

**Changes:**
- If `function` add `()`
- If `hook`, apply the correct `function` that wraps it. 
- If `hook` of `functions` has arguments, add `...` to denote parameters
  - I tried to show all the properties but it's overwhelming, at least how I had them styled.
- Show parameters in a list
- If parameters list is `> 4` truncate 

There are no examples of classes, and this wasn't tested on "classes" because of #35. This PR should probably wait for that bug fix.

If you think there is more pertinent information to add to the embed that is missing, please comment below!

### DIFF
--- a/source/wp-content/themes/wporg-developer/embed.php
+++ b/source/wp-content/themes/wporg-developer/embed.php
@@ -7,6 +7,8 @@
 
 use DevHub;
 
+remove_action( 'embed_content_meta', 'print_embed_comments_button' );
+
 if ( ! headers_sent() ) {
 	header( 'X-WP-embed: true' );
 }
@@ -64,7 +66,7 @@ $embed_post_id         = get_the_ID();
 $params                = get_params( $embed_post_id );
 $embed_title           = get_signature( $embed_post_id, count( $params ) > 0 );
 $param_count           = count( $params );
-$parameter_display_max = 3; // We truncate the display of params
+$parameter_display_max = 4; // We truncate the display of params
 
 ?>
 <!DOCTYPE html>
@@ -133,7 +135,7 @@ $parameter_display_max = 3; // We truncate the display of params
 
 			<?php if ( $params ) : ?>
 				<div class="wp-embed-parameters">
-					<h6 class="wp-embed-parameters-title"><?php echo __( 'Parameters', 'wporg' ); ?></h6>
+					<h6 class="wp-embed-parameters-title"><?php echo esc_html__( 'Parameters:', 'wporg' ); ?></h6>
 					<ul>
 						<?php
 						for ( $i = 0; $i < min( $param_count, $parameter_display_max ); $i++ ) {
@@ -179,6 +181,15 @@ $parameter_display_max = 3; // We truncate the display of params
 				do_action( 'embed_content_meta' );
 				?>
 			</div>
+
+			<?php
+				/**
+				 * Print scripts or data before the closing body tag in the embed template.
+				 *
+				 * @since 4.4.0
+				 */
+				do_action( 'embed_footer' );
+			?>
 		</div>
 	</div>
 </body>
--- a/source/wp-content/themes/wporg-developer/embed.php
+++ b/source/wp-content/themes/wporg-developer/embed.php
@@ -95,6 +95,7 @@ $parameter_display_max = 4; // We truncate the display of params
 
 		p.wp-embed-heading {
 			font-weight: normal;
+			font-size: 1.25rem;
 			font-family:
 				Hack, "Fira Code", Consolas, Menlo, Monaco, "Andale Mono",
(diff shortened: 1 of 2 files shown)


ryelle:
This is looking good! I started playing with it locally so I hope you don't mind that I made a few changes to the CSS.

- Increased the contrast of the text so it passes AA
- Switched the title to use a code font, since it's the function/hook name
- Moved the quotes and comma into the `wp-embed-hook` span, since those are also green on the main page
- Switched to single quotes for hooks, unless there is a variable

Screen Shot 2022-06-02 at 12 32 27 PM

StevenDufresne:
Those are all great additions. What about we reduce the heading size to 20px? I find it kind of big, especially with the "Fira Code" font. 

The difference appears negligible in the screenshots below, but it goes a long way for longer functions.

| Before | After |
| --- | --- |
|  | |

source/wp-content/themes/wporg-developer/embed.php
```diff
@@ -38,13 +36,13 @@ function get_params( $post_id ) {
...
- * @param int     $post_id
- * @param boolean $has_args
+ * @param int $post_id
  *
  * @return string
  */
-function get_signature( $post_id, $has_args ) {
+function get_signature( $post_id ) {
 	$title = get_the_title();
+	$has_args = count( get_params( $post_id ) ) > 0;
```
StevenDufresne, line 45: @ryelle Why this change? I didn't want to make `get_params` a dependency of `get_signature` seeing that its values are not used. Although there are no tests, it does make it _less_ testable.

tellyworth:
These look great!

The hook stuff has got me thinking about the way we display those in the headline, ie `apply_filters( 'hook', ... )` and `do_action( 'hook', ... )`. Not just in embeds but in general.

That is definitely clearer than displaying just the hook name. And it'll definitely make some people happy. But it feels kind of backwards to me, and I wonder if some other people might feel the same. What I mean is, in the function reference, the title/headline represents _the code I should write_ in order to use the function, more or less. Whereas with the hooks, the headline represents _the code core uses to call the hook_, not the code I should write.

Perhaps both of these would be solved by having a one-line example immediately below the headline for every function, hook, etc. Something like:

## apply_filters( 'pre_wp_filesize', ... )
`add_filter( 'pre_wp_filesize', 'example_callback', 10, 2 );`

..where the example line is syntax highlighted and easily copyable. We could also allow multiple examples, ref https://github.com/WordPress/wporg-developer/discussions/12#discussioncomment-2807259

dd32:
> Perhaps both of these would be solved by having a one-line example immediately below the headline for every function, hook
> ...
> `add_filter( 'pre_wp_filesize', 'example_callback', 10, 2 );`

with #12 in mind, this would work well for hooks IMHO, but I'd expand that to have the function in there too:
```
add_filter( 'pre_wp_filesize', 'example_callback', 10, 2 );
function example_callback( $param1, $param2 ) {
    return $param1;
}
```

source/wp-content/themes/wporg-developer/embed.php
ryelle: Oops, I left my reasoning in the commit, not the comment — I did that so it would match the main theme's template tag (https://github.com/WordPress/wporg-developer/pull/59/commits/e52a950e31b70407c19ffbbfd2892dc4b70a6feb). It also surprised me that you would need to tell `get_signature` that the function has args, it seems like something that should be detected in the function (since usually it would also output those params - though I know it doesn't here).

StevenDufresne: Got it. I think I would prefer to have it less similar than more. Maybe it makes more sense to create a truncated form of the function in `template-tags.php` since it does appear like we'll be keeping the signature around as it is with some modifications. 

What about something like:
```
function get_truncated_signature() {
 return get_signature( ..., true )
}

...

function get_signature( ..., truncate = false ) {

}
```

StevenDufresne:
I agree that for hooks `do_action` & `apply_filters` are useless. Can we change the headline for those and then also follow up with a fully qualified example afterward?

source/wp-content/themes/wporg-developer/embed.php
ryelle: Yeah, if we might use it elsewhere, I like your suggestion 👍🏻

PR #60, opened 2022-06-02
Check if source is empty before printing block.

### LINKED ISSUE #58: When the source block is empty, the code container is still visible.
Most likely related to #44.

There are functions that don't havecode in source code block:
[media_date_column_time](https://developer.wordpress.org/reference/hooks/media_date_column_time/)

Since we use a new block to render the source code, when there is no code, the code container is still visible.

StevenDufresne:
Fixes: #58.

Check to make sure the source code is not empty.

### Considerations
I considered updating `post_type_has_source_code` since it kind of implies that there is source code, but instead it checks if it _should_ have source code based on its type.

We could add the `get_source_code()` check inside of `post_type_has_source_code` as well, but it does database calls so it more performant to leave it out.

### DIFF
--- a/source/wp-content/themes/wporg-developer/reference/template-source.php
+++ b/source/wp-content/themes/wporg-developer/reference/template-source.php
@@ -10,7 +10,7 @@ namespace DevHub;
 
 $source_file = get_source_file();
 if ( ! empty( $source_file ) ) :
-	$source_code = get_source_code();
+	$source_code = post_type_has_source_code() ? get_source_code() : '';
 	?>
 	<hr />
 	<section class="source-content">
@@ -32,7 +32,7 @@ if ( ! empty( $source_file ) ) :
 			?>
 		</p>
 
-		<?php if ( post_type_has_source_code() && ! empty( $source_code ) ) : ?>
+		<?php if ( ! empty( $source_code ) ) : ?>
 			<?php
 				echo do_blocks(
 					sprintf(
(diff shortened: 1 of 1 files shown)


ryelle:
I mentioned in https://github.com/WordPress/wporg-developer/issues/58#issuecomment-1144954764 that I don't see an empty code block, so I can't directly test this.

Are there valid cases where `post_type_has_source_code` is true, but the source code is empty? If not, it sounds like an import error, and it would be good to have some kind of indicator that something went wrong here.

StevenDufresne:
> Are there valid cases where `post_type_has_source_code` is true, but the source code is empty? If not, it sounds like an import error, and it would be good to have some kind of indicator that something went wrong here.

That's a good question. @dd32 Any idea?

dd32:
> Are there valid cases where `post_type_has_source_code` is true, but the source code is empty?

There shouldn't be, ish.
 - Prior to #28 parsing wouldn't import the hook source. I suspect maybe @StevenDufresne hasn't re-parsed since that was merged. The developer code-base does have the ability to import the source on-demand, maybe that doesn't work for hooks.
 - `post_type_has_source_code()` should be false for hooks in production, but the screenshot from #58 is seen on every hook page in production.

StevenDufresne:
> If not, it sounds like an import error, and it would be good to have some kind of indicator that something went wrong here.

I still think this PR adds better protection than it currently has and a missing source block can probably go quietly seeing that it's supplementary information and the source is linked.

ryelle (approved):
I don't think I have a way to test this, but the code LGTM 👍🏻

PR #62, opened 2022-06-02
Source: File link should have better language.

StevenDufresne:
If you click on the file name in the Source section, it takes you to a page with other function/hooks related to that file. I found that useful but also surprising. I expected it to go to the source file.

| Before | After |
| --- | --- |
|  |   |

### DIFF
--- a/source/wp-content/themes/wporg-developer/reference/template-source.php
+++ b/source/wp-content/themes/wporg-developer/reference/template-source.php
@@ -15,9 +15,20 @@ if ( ! empty( $source_file ) ) :
 	<section class="source-content">
 		<h2><?php _e( 'Source', 'wporg' ); ?></h2>
 		<p>
-			<?php printf( __( 'File: %s', 'wporg' ),
-				'<a href="' . esc_url( get_source_file_archive_link( $source_file ) ) . '">' . esc_html( $source_file ) . '</a>'
-			); ?>
+			<?php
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-developer/reference/template-source.php
```diff
@@ -15,9 +15,20 @@
 	<section class="source-content">
 		<h2><?php _e( 'Source', 'wporg' ); ?></h2>
 		<p>
-			<?php printf( __( 'File: %s', 'wporg' ),
-				'<a href="' . esc_url( get_source_file_archive_link( $source_file ) ) . '">' . esc_html( $source_file ) . '</a>'
-			); ?>
+			<?php
+			printf(
+				__( 'File: %s.', 'wporg' ),
+				esc_html( $source_file )
```
ryelle: What do you think about wrapping this in `code`?



```suggestion
				'<code>' . esc_html( $source_file ) . '</code>'
```

ryelle (approved):
Small suggestion, otherwise good change (I also always think it'll take me to the file).

source/wp-content/themes/wporg-developer/reference/template-source.php
StevenDufresne: Yep, that looks good! Thanks.

PR #64, opened 2022-06-02
Sass: CSS cleanup and linter fixes

ryelle:
This PR is the result of running the `stylelint` fixer, then doing some manual review and cleanup. I've removed some of the unnecessary fallbacks (ex, doubling up on `rem` declarations with a `px` fallback — `rem`s are supported now so the `px` duplicates have been removed). I also removed some but not all things that were marked as "for S5" or "IE8/9", just in case they're still relevant — I wanted to keep this as automated as possible so I don't need to manually check everything.

I also added a folder `scss/settings`, which contains the colors and the font variables which had previously been loaded in `_global.scss`, this way I can use the same font variables in `prism.scss`.

**To test**

1. Build the CSS
2. View the site
3. There should be no visual changes

Try running the linter:

```bash
yarn lint:css source/wp-content/themes/wporg-developer/scss
```


ryelle:
I'm merging this since there was no objection here or on #65, and the longer it sits the more likely there will be merge conflicts.

PR #49, opened 2022-06-01
Try some parameter doc fixes

dd32:
See #45.

Notes:
 - This is not mergable as-is, and is mostly untested.
 - Commit [Process lists within parameter blocks when deliberate br's are reta…](https://github.com/WordPress/wporg-developer/commit/249038c2dc524d3130c1bd6963327009b0537f83) relies upon this upstream change: https://github.com/WordPress/phpdoc-parser/compare/master...try/dev-45-newlines-in-docs

| Before | After |
| --- | --- |
| With Commit 1: [Parse lists within hash parameters.](https://github.com/WordPress/wporg-developer/commit/b7b60c323ade89245ce767cb5bf322b59b712663) |
| Screen Shot 2022-06-01 at 4 47 26 pm | Screen Shot 2022-06-01 at 4 11 42 pm |
| With Commit 2: [Process lists within parameter blocks when deliberate br's are reta…](https://github.com/WordPress/wporg-developer/commit/249038c2dc524d3130c1bd6963327009b0537f83) |
| See above | Screen Shot 2022-06-01 at 4 39 40 pm |
| Better example of newline retention |
| Screen Shot 2022-06-01 at 4 50 13 pm | Screen Shot 2022-06-01 at 4 50 38 pm |

PR #65, opened 2022-06-02
Reference Page: Update style of parameters and signature

ryelle:
Iterate a little on the reference page styles.

- Switch colors to the [core color](https://codepen.io/ryelle/full/WNGVEjw) counterparts over 50 to ensure contrast against the white background
- Parameters are not using the grey background `code` style, but are now using the blue from the call signature
- Links in the call signature should inherit the colors (see the hooks section before & after)
- Links in the parameter types have dotted underlines by default, and solid underline on hover (so it's clear this is still a link)
- Style nested return types using the same styles as parameters (see [comment here](https://github.com/WordPress/wporg-developer/pull/49#issuecomment-1144229284))

See #20.

**Screenshots**

| Before | After |
|--------|-------|
| before-header | after-header |
| before-complex-params | after-complex-params |

Hooks section with linked call signature

| Before | After |
|--------|-------|
| before-hooks | after-hooks |

Nested return types highlighted correctly

| Before | After |
|--------|-------|
| Screen Shot 2022-06-02 at 17 17 15 | Screen Shot 2022-06-02 at 17 15 56 |

Note: this builds on the linter fixes in #64, but if we decide we don't want all that code turnover I can remake this on `trunk`.

### DIFF
--- a/source/wp-content/themes/wporg-developer/scss/main.scss
+++ b/source/wp-content/themes/wporg-developer/scss/main.scss
@@ -1108,6 +1108,7 @@
 		}
 
 		.return-type,
+		.return .type,
 		.parameters .type {
 			color: get-color(red-50);
 
@@ -1122,7 +1123,8 @@
 			}
 		}
 
-		.parameters {
+		.parameters,
+		.return {
 			p {
 				margin-bottom: 0;
 			}
--- a/source/wp-content/themes/wporg-developer/scss/main.scss
+++ b/source/wp-content/themes/wporg-developer/scss/main.scss
@@ -1130,7 +1130,6 @@
 			}
 
 			dt code {
-				color: get-color(blue-60);
 				font-weight: 400;
 			}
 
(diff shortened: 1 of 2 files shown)


StevenDufresne:
I'm not quite sold on the blue color since it's very similar to the blue that we use for links. Thoughts?

dd32:
> I'm not quite sold on the blue color since it's very similar to the blue that we use for links.

I was sold on the blue, but then you pointed out links and I paused. I think the underlines on the links differentiates them enough though?

ryelle:
I'm using the blue because it's what's used in the signature (I assume that's why the types are red), but if #71 merges, those colors will change. I've dropped the blue change in https://github.com/WordPress/wporg-developer/pull/65/commits/d086e73923b1bb8c2b0ab1e43fc2beac652f7331, so we can get the other changes in, and we can revisit the param color later.

PR #68, opened 2022-06-03
Update pagination styles.

### LINKED ISSUE #63: The Pagination controls are hard to use.
The pagination controls are hard to use. 

- Items are close together
  - Click target is small.  
- Active item doesn't have a lot of contrast.

Could we use the [pattern directory pagination controls](https://github.com/WordPress/pattern-directory/blob/825db0b4816b2c02d21c9df91e83acb9e360ad65/public_html/wp-content/themes/pattern-directory/src/components/pattern-grid/pagination.js)? They are the newest pagination control designs we have. Reusing the code would be ideal, but even if we just use the styles, I think it will be much better.

### Controls
| Docs | Pattern Directory |
| --- | --- |
|  | |

StevenDufresne:
Closes: #63 

Take some of the applicable styles from the pattern directory pagination controls and move them to `main.scss`. This also adds the `wp4.css` link color to `_colors.scss` to keep it consistent. 

This theme overrides that color for links on hover but it's probably better to keep it around until we make the decision to remove it from everywhere.

### Changes
| Before | After |
| --- | --- |
|   |  |

### DIFF
--- a/source/wp-content/themes/wporg-developer/inc/loop-pagination.php
+++ b/source/wp-content/themes/wporg-developer/inc/loop-pagination.php
@@ -54,8 +54,8 @@ function loop_pagination( $args = array() ) {
 		'total'        => $max_num_pages,
 		'current'      => $current,
 		'prev_next'    => true,
-		'prev_text'  => __( 'Previous' ),
-		'next_text'  => __( 'Next' ),
+		'prev_text'    => __( 'Previous', 'wporg' ),
+		'next_text'    => __( 'Next', 'wporg' ),
 		'show_all'     => false,
 		'end_size'     => 2,
 		'mid_size'     => 1,
@@ -110,4 +110,4 @@ function loop_pagination( $args = array() ) {
 		return $page_links;
 }
 
-?>
+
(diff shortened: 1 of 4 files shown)


dd32 (approved):

source/wp-content/themes/wporg-developer/scss/_colors.scss
```diff
@@ -8,6 +8,7 @@ $color-black: #000;
 $colors: (
 	white: #fff,
 	black: #000,
+	wp4-link-hover: #d54e21,
```
ryelle: Can we add this as a separate sass variable? I think this map should stay as just the core color palette, since [it's copied as-is from the reference](https://codepen.io/ryelle/pen/WNGVEjw?editors=0100).

ryelle (changes requested):

source/wp-content/themes/wporg-developer/scss/_colors.scss
ryelle: I made this change since I rebased the PR after #64.

PR #69, opened 2022-06-03
Add a CLI script for scrubbing email addresses etc from export files

tellyworth:
We could periodically use this on the server side to generate sample export files.

I used `DomDocument` rather than regexes to better deal with large files and unexpected whitespace etc.

### DIFF
--- a/env/scrub-export.php
+++ b/env/scrub-export.php
@@ -1,0 +1,67 @@
+<?php
+
+/**
+ * A CLI tool for scrubbing a WordPress WXR export file of likely PII.
+ */
+
+$opts = getopt( '', ['namespace:'], $last );
+
+$namespace = $opts['namespace'] ?? 'http://wordpress.org/export/1.2/';
+$infile = $argv[ $last ] ?? 'php://stdin';
+$outfile = $argv[ $last + 1 ] ?? 'php://stdout';
+
+if ( is_null( $infile ) ) {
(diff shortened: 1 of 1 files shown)


tellyworth:
@iandunn mentioned the idea of taking an allowlist approach to scrubbing, rather than scrubbing selected elements like I've done here. That's a good idea for future, especially if we were to reuse this for other sites. That was my other reason for using DomDocument - we could traverse every node and scrub everything that's not explicitly allowed.

If you think this is ok as a starting point let's merge it and add some handbook imports.

env/scrub-export.php
```diff
@@ -0,0 +1,67 @@
...
+			++ $count_replaced;
+		}
+	}
+
+	fwrite( STDERR, "Replaced $count_replaced instances of wp:$tag\n" );
+}
+
+fwrite( $fp_out, $doc->saveXML() );
+
+fclose( $fp_out );
```
iandunn, line 67:
```suggestion
fclose( $fp_out );

readline( "\nMake sure you manually review the diff before committing this to the repository! This script might have missed some PII. \n" );
```

iandunn:
Here's an example of a safelist approach, for future reference:

https://github.com/iandunn/regolith/blob/824b999c11fae10c4777b8f517df63c6e6fd0a0d/bin/sync-production-content.sh#L91-L164

It works on a SQL dump rather than WRX, but has safelists of various data that might be useful.

tellyworth:
This is probably not needed now but worth keeping for future use.

PR #71, opened 2022-06-03
Use code block for `get_signature`.

### LINKED ISSUE #23: The first code block should be a copyable example.
### Discussed in https://github.com/WordPress/wporg-developer/discussions/12

We should prefer copyable code blocks over function definitions. While users can already copy the function definitions, maybe because of the [datatypes](https://github.com/WordPress/wporg-developer/discussions/8) or because [it doesn't look like a code block](https://github.com/WordPress/wporg-developer/discussions/5), it isn't apparent that I can use that line of code.

By removing the argument names, and optional parameters and giving the block more visual preference, we have a block of code that is easy to copy and paste into your own environment.

| Before | After |
| --- | --- |
|   |   |

Note: I've switched the description and parameters section. Related: https://github.com/WordPress/wporg-developer/discussions/9 https://github.com/WordPress/wporg-developer/discussions/7

</div>

StevenDufresne:
Closes: #23.
Related to: #65.

This PR uses `do_blocks` and `get_signature` to display the first code block on the page. We didn't agree on this approach, so we can have a discussion as to whether this approach makes sense.

**Changes**
- `do_blocks`
- Adds CSS to `prism.scss` for consistency since `<pre>` is not a valid child of `<h1>` and there doesn't appear to be a worthy `<h1>` other than the signature.

**Notes**
- If we were to go with this approach, we would probably want to change the data type colors in the parameters sections to match the code block.
- Also, if we wanted to go this route, we should probably update all the uses of `get_signature` (hook section..)
- The `wp` class below looks pretty sad, but it already does. I think maybe we should add `class` to it. I'll open another ticket.

## Screenshots
| Before | After |
| --- | --- |
|  |  |
|  |  |
|  |    |

### DIFF
--- a/source/wp-content/themes/wporg-developer/content-reference.php
+++ b/source/wp-content/themes/wporg-developer/content-reference.php
@@ -6,7 +6,13 @@
 
 	<?php echo get_private_access_message(); ?>
 
-	<h1><?php echo get_signature(); ?></h1>
+	<h1><?php echo do_blocks(
+			sprintf(
+				'<!-- wp:code {"lineNumbers":false} --><div class="wp-block-code"><code lang="php" class="language-php">%s</code></div><!-- /wp:code -->',
+				get_signature()
+			)
+		);
+	?></h1>
 
 	<section class="summary">
 		<?php echo get_summary(); ?>
--- a/source/wp-content/themes/wporg-developer/scss/main.scss
+++ b/source/wp-content/themes/wporg-developer/scss/main.scss
@@ -1028,8 +1028,6 @@
 
 		h1 {
 			margin: 24px 0;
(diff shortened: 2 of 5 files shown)


dd32:
My personal opinion is that the styling of the code as shown in the screenshots here is not heading-enough, and too code-blocky.

Maybe it's just a case of needing to style that with larger font / larger line-height?

ryelle:
I think the suggestion of having a copyable example (separate from the header) in #59/#23 sounds like a good compromise — we could keep the existing signature as a heading, but immediately under it would be a copyable example of use.

StevenDufresne:
I'll close this since it appears we'll stick with the signature as-is.

PR #72, opened 2022-06-03
Try adding `class %s {}` string to classes

StevenDufresne:
When viewing a class, we only use the title in the `get_signature` which seems pretty lame.

**Update: I realized that C in class should be lower cased. I don't however want to redo the screenshots :)**

Related to: #72

| Before | After |
| --- | --- |
|   |   |
 |   |

### DIFF
--- a/source/wp-content/themes/wporg-developer/inc/template-tags.php
+++ b/source/wp-content/themes/wporg-developer/inc/template-tags.php
@@ -707,8 +707,7 @@ namespace DevHub {
 		$types        = array();
 
 		if ( 'wp-parser-class' === get_post_type( $post_id ) ) {
-			/* translators: %s: The name of the class. */
-			return sprintf( __( 'class %s {}', 'wporg'), $signature );
+			return '<span class="keyword">class</span> ' . $signature . ' {}';
 		}
 
 		if ( $tags ) {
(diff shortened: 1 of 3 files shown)


dd32 (approved):

ryelle (approved):

source/wp-content/themes/wporg-developer/inc/template-tags.php
```diff
@@ -707,7 +707,8 @@ function get_signature( $post_id = null ) {
 		$types        = array();
 
 		if ( 'wp-parser-class' === get_post_type( $post_id ) ) {
-			return $signature;
+			/* translators: %s: The name of the class. */
+			return sprintf( __( 'class %s {}', 'wporg'), $signature );
```
ryelle: I know I just approved this, but should this be translatable? `class Blah {}` should always be English because it's PHP, right?

ryelle: In https://github.com/WordPress/wporg-developer/pull/72/commits/d4d1b8cdd4da3d050af79d0f3b4a84cffdff4513, I removed the translation and also added a span wrapper around `class`, so we can style it like the `apply_filters`/etc — so it's visible, but a little less visually important.

Screen Shot 2022-06-03 at 12 14 57 PM

StevenDufresne:
🤦 trying to translate "class". Haha. What was I thinking.. Thanks for fixing! 😄

PR #73, opened 2022-06-03
WIP: display a kinda-sorta sidebar menu on function reference pages

tellyworth:
There's a lot wrong with this, I hammered junk into place to make something work so we could see if it's worth doing properly.

Current screenshot:
Screen Shot 2022-06-03 at 3 50 33 pm

The menu displays other functions within the same package. There are probably pages where it all goes wrong, I haven't tested widely.

The width of the menu and content is wrong, I haven't figured out why yet.

### DIFF
--- a/source/wp-content/themes/wporg-developer/functions.php
+++ b/source/wp-content/themes/wporg-developer/functions.php
@@ -137,6 +137,11 @@ if ( is_admin() ) {
 }
 
 /**
+ * Function reference sidebar menu.
+ */
+require __DIR__ . '/inc/sidebar-menu.php';
+
+/**
  * Set the content width based on the theme's design and stylesheet.
  */
 if ( ! isset( $content_width ) ) {
@@ -172,6 +177,8 @@ function init() {
 
 	add_filter( 'mkaz_code_syntax_force_loading', '__return_true' );
 	add_filter( 'mkaz_prism_css_path', __NAMESPACE__ . '\\update_prism_css_path' );
+
+	add_filter( 'body_class', __NAMESPACE__ . '\\body_class' );
 }
 
 /**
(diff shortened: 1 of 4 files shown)


StevenDufresne:
Worth mentioning, I have an experiment PR #75 that moves the search form into the site-title. If this sidebar PR works, we also have the option to move the search control to the top of the sidebar like the block editor.

ryelle:
I'm not really sure if the function list is useful information for the sidebar — what about moving just the table of contents over? That would also fix #76, since the code source would be able to span the full width of the area.

Also a note for anyone testing - you need to drop any block/widget into the widget area for the sidebar to appear.

tellyworth:
I think you're right that the menu as implemented here isn't the best use of space, but it's not really clear what would be.

I thought about putting the inset TOC in the sidebar, but that's inconsistent with the other handbook sections in devhub. And the TOC isn't super useful either.

The function list (either by package or by file) could be quite useful, even if it's not in the sidebar. There are some thinly documented functions where there is no good way to discover related functions other than by clicking through to the file view. Having them listed right there on the page would be helpful - like the PHP manual does in the right-hand sidebar.

The block-editor handbook does have a function/component reference in its handbook:
Screen Shot 2022-06-06 at 2 29 56 pm

However the function reference is 10x as large as that handbook, and it would probably be both unwieldy for users and slow.

I think all of the sidebar menus are a bit broken really - the fact that you can't navigate to (or even discover) the other handbooks is a major flaw. 

I think these are mostly separate problems:

* Displaying a list of functions in the same file/package would be useful (like php.net) - but where?
* Having a sidebar menu for the function reference would make sense for consistency
* Making all of the handbooks discoverable from the menu

StevenDufresne:
Since WordPress is pretty consistent with naming conventions, could we maybe group by prefix? 
- `_` or `_wp` 
- `get_` 
- `is_`
- `post_`

StevenDufresne:
@tellyworth I'll close this issue seeing that we are currently implementing this in the new theme.

PR #74, opened 2022-06-03
Import GitHub-hosted handbooks

dd32:
This PR adds a command to import the GitHub hosted handbooks, by triggering the pending crons during provisioning.

The wp-env cron doesn't appear to be running, for whatever reason, but is not a problem from this repo specifically.

Changes:
 - Jetpack is now included
 - Local environments have a filter to switch URLs from developer.wordpress.org to localhost in markdown imports
 - All cron events run during provisioning.
 - Block Editor handbook: A warning from `WP_CORE_STABLE_BRANCH` is avoided if not defined. This might result in some invalid URLs, but I haven't found any.


dd32:
> The wp-env cron doesn't appear to be running, for whatever reason, but is not a problem from this repo specifically.

Upstream bug:
https://github.com/WordPress/gutenberg/issues/20569

Issue #57, opened 2022-06-02
Search autocomplete covers filter options


iandunn:
1. open `http://localhost:8888/reference/`
2. type `embed`

The autocomplete box appears and covers all of the filters. So to check `hooks`, you have to hit escape to close the autocomplete and check the box before submitting the search.

<img src="https://user-images.githubusercontent.com/484068/171521111-57f4fc44-d970-4cab-8786-5734cd12d4a4.png">

One way to solve that would be to move the filters above the query box:

<img src="https://user-images.githubusercontent.com/484068/171521356-88239f8e-0c2e-45db-bf0e-6c9d658d276b.png">

dd32:
I personally like the idea of moving it to above the search box.

outdoor2kode:
Note: search bar and filter options style and layout have been changed in #75.

iandunn:
Ah, thanks, I already had it done locally, so went ahead and opened the PR. It can wait until #75 is merged though.

adamwoodnz:
This is still an issue on the search page with the new theme

![Image](https://github.com/WordPress/wporg-developer/assets/1017872/433f0fe7-f22b-485a-a269-11e7913f94e8)

PR #75, opened 2022-06-03
Move the search bar control into the site-title.

StevenDufresne:
This PR is an experiment related to #40. Using the pattern directory & learn as the model, move the search bar into the site title.

**Example**
- [https://wordpress.org/patterns/categories/gallery/](https://wordpress.org/patterns/categories/gallery/)
- [https://learn.wordpress.org/workshops/](https://learn.wordpress.org/workshops/)

### Function/Class/Hook
| Before |  After |
| --- | --- | 
|   |   |
|   |  |

### `is_page( 'reference' )`

I removed everything below the breadcrumb and above the 2 boxes since I didn't really know what to do with the layout and I'm not convinced they are used much. 

| Before |  After |
| --- | --- | 
|  |   |
|  |  |

### `is_search()`
| Before |  After |
| --- | --- | 
|   |   |

### `is_handbook && is_search()`
| Before |  After |
| --- | --- | 
|   |   |


outdoor2kode:
This looks great!

Done testing on Mac/Chrome & Firefox with different screen sizes. 
Some thoughts:
1. The filter is removed from the API detail page, I think it's ok to me as it's still kept on the reference home page and I personally didn't use that filter that much. But not sure if it's a convenient tool for a veteran developer?

2. I saw there's a tiny space under the search icon and there's no outline style when clicking on it, perhaps we can give it some styles just like the one in the pattern directory?
image
image

3. It would be great if we can change the cursor to pointer for these two icons
image

4. I seem not able to see the search bar on the searching page like your screenshot shows
Screen Capture on 2022-06-08 at 01-09-25
5. I noted that if I'd like to get rid of the search suggestion block, I need to use the mouse to click elsewhere on the page instead of just pressing the ESC button. If so, users need to select the filter options first before typing in the search or they must use a mouse. I think in the original UX, the suggestion block can be gotten rid of by pressing ESC.
image
    5a. Also tested on Firefox, and the suggestion block can be closed by pressing ESC, but it has a weird behavior that after closing it, it will pop up again.
Screen Capture on 2022-06-08 at 01-32-08

6. Curious about what "This needs some clean up" means in the Reference Home screenshot, does it mean that you expect some new style design for it?

7. Commit message typepo: Move `searhc` into site title

StevenDufresne:
Thanks for the detailed review! I should have been more clear. The code needs cleaning up and some of the UI interactions/layouts are still a bit rough. 

I am interested to get this approach validated since we don't have an agreed-on path forward on the search UX. Do you think this is a good idea (moving the search bar)? Is it an improvement?

outdoor2kode:
To me, it's an improvement and I'm probably more into this new layout for 3 reasons. 

1. It's more consistent with the design of other WP sites.
2. I like it I don't have to toggle on the search bar every time, it's an extra step as far as I'm concerned.
3. I like the new UI, it's more obvious and looks nicer by putting it in the title header.

StevenDufresne:
@renintw Ok, cycling back to your feedback. I think I addressed most of them except:

>I noted that if I'd like to get rid of the search suggestion block, I need to use the mouse to click elsewhere on the page instead of just pressing the ESC button. If so, users need to select the filter options first before typing in the search or they must use a mouse. I think in the original UX, the suggestion block can be gotten rid of by pressing ESC.

This is an external plugin and I suspect this functionality is already _broken_. We can address it in a later PR.

source/wp-content/themes/wporg-developer/header.php
```diff
@@ -24,33 +33,27 @@
...
+			<nav id="site-navigation" class="main-navigation" role="navigation">
+				<button class="menu-toggle dashicons dashicons-arrow-down-alt2" aria-controls="primary-menu" aria-expanded="false" aria-label="<?php esc_attr_e( 'Primary Menu', 'wporg' ); ?>"></button>
+				<?php
+				$active_menu = is_post_type_archive( 'command' ) || is_singular( 'command' ) ? 'devhub-cli-menu' : 'devhub-menu';
+				wp_nav_menu( array(
+					'theme_location'  => $active_menu,
+					'container_class' => 'menu-container',
+					'container_id'    => 'primary-menu',
+				) ); ?>
+			</nav>
```
ryelle, line 50: If I add items to this menu, they show up kind of unexpectedly:

Screen Shot 2022-06-09 at 3 24 04 PM

It looks like this menu is empty on production, so perhaps it's fine, but it might be good to at least make this presentable (or remove the menu). That can happen in a followup PR.

ryelle (commented):
Great update, this is much more discoverable than the little 🔍  tab.

> I noted that if I'd like to get rid of the search suggestion block, I need to use the mouse to click elsewhere on the page instead of just pressing the ESC button. If so, users need to select the filter options first before typing in the search or they must use a mouse. I think in the original UX, the suggestion block can be gotten rid of by pressing ESC.

For what it's worth, using <kbd>ESC</kbd> to close the dropdown is working for me, in both Chrome & Firefox.

I really like this UI for the checked element 💯 

Screen Shot 2022-06-09 at 3 14 27 PM

I did notice the the autocomplete dropdown is now applied to the other searches too, like here in the block editor handbook. Hitting enter (ignoring the suggestions) does direct me to the correct search results page (ex, `/block-editor/?s=block`), so I'd guess the autocomplete JS is a little too overzealous.

Screen Shot 2022-06-09 at 3 34 08 PM

StevenDufresne:
> I did notice the the autocomplete dropdown is now applied to the other searches too, like here in the block editor handbook. Hitting enter (ignoring the suggestions) does direct me to the correct search results page (ex, `/block-editor/?s=block`), so I'd guess the autocomplete JS is a little too overzealous.

This must be a configuration issue. On the live site, the search gets `class="searchform searchform-handbook"` and the js specifically [excludes `searchform-handbook` search inputs](https://github.com/WordPress/wporg-developer/blob/4fed039f61bf696c45e94692c923bb287ca9f3d8/source/wp-content/themes/wporg-developer/js/autocomplete.js#L14).

StevenDufresne:
I added code to hide the menu on the homepage/reference, although the menu is still stacked under the search bar when you're looking at a page with the searchbar beside the site-title. Since production doesn't have a menu right now, as @ryelle mentions we can probably just follow it up later.

@renintw I had to squash merge because of conflicts introduced in #40. Do you mind double-checking to make sure I didn't overwrite any changes? I had to manually merge them. 🙏

outdoor2kode:
>> I noted that if I'd like to get rid of the search suggestion block, I need to use the mouse to click elsewhere on the page instead of just pressing the ESC button. If so, users need to select the filter options first before typing in the search or they must use a mouse. I think in the original UX, the suggestion block can be gotten rid of by pressing ESC.

> For what it's worth, using ESC to close the dropdown is working for me, in both Chrome & Firefox.

I should have been more clear. I can use ESC to close dropdown in Chrome as well, but in Chrome it also clears the user input while Firefox doesn't.
It would affect one scenario that if users want to select filter types below, they need to use the mouse to close the dropdown or select the types before inputting.

Note: A possible solution is mentioned in #57

outdoor2kode (approved):
Looks good to me, It's nice!

I've added three commits to adjust styles and a11y and also have rebased to trunk.
Original small style defect (focus state):
Markup on 2022-06-10 at 22:45:38

ryelle:
> This must be a configuration issue. On the live site, the search gets class="searchform searchform-handbook" and the js specifically [excludes searchform-handbook search inputs](https://github.com/WordPress/wporg-developer/blob/4fed039f61bf696c45e94692c923bb287ca9f3d8/source/wp-content/themes/wporg-developer/js/autocomplete.js#L14).

It wasn't a config issue, this PR was removing the logic that added the class. I [brought it back](https://github.com/WordPress/wporg-developer/pull/75/commits/3a26838a50b940e4e815a6956bdd6e4760fa9e94) & styled the search a little more. I also removed some unused code now that we don't have the drawer-search UI. 

Screen Shot 2022-06-10 at 2 24 06 PM
Screen Shot 2022-06-10 at 2 23 55 PM

I'm going to merge this so it doesn't block other work.

iandunn:
Anyone else seeing this in Firefox?

Screen Shot 2022-06-10 at 1 19 10 PM

_( on `trunk` after running `build`, at `1900px` )_

Oddly, the horizontal bar goes away in responsive mode.

iandunn:
I included a fix for ^ in #114

StevenDufresne:
Thanks for jumping in on this all!

Issue #50, opened 2022-06-01
Some function references aren't auto-linked.


dd32:
Spurred from #45, See https://developer.wordpress.org/reference/classes/WP_Query/parse_query/  when a method reference doesn't include a `()` after it or it uses `->` instead of `::` it fails to be linked.

Examples:
| -> | no () |
| --- | --- |
| <img width="520" alt="Screen Shot 2022-06-01 at 4 55 29 pm" src="https://user-images.githubusercontent.com/767313/171345681-edd0b760-b44d-4ea7-ad93-9af1fcb06b9b.png"> | <img width="665" alt="Screen Shot 2022-06-01 at 4 55 43 pm" src="https://user-images.githubusercontent.com/767313/171345695-647ac173-8972-4224-84d2-2670decdc6d1.png"> |

dd32:
After looking at the regular expression here...

The items missing `->` could probably be handled here, but given the below, might be better to just fix it in core.

The items missing a `()` kind of need to be fixed in core, for simple regex life here.. it's a docs-bug after-all I think.

Example diff to fix the `->` case, and which might show why handling `()` here isn't a choice I want to make..
```diff
diff --git a/source/wp-content/themes/wporg-developer/inc/formatting.php b/source/wp-content/themes/wporg-developer/inc/formatting.php
index 68ec682..797eb4e 100644
--- a/source/wp-content/themes/wporg-developer/inc/formatting.php
+++ b/source/wp-content/themes/wporg-developer/inc/formatting.php
@@ -386,7 +386,7 @@ class DevHub_Formatting {
                                                '~
                                                        (?!<.*?)       # Non-capturing check to ensure not matching what looks like the inside of an HTML tag.
                                                        (              # 1: The full method or function name.
-                                                               ((\w+)::)? # 2: The class prefix, if a method reference.
+                                                               ((\w+)(::|->))? # 2: The class prefix, if a method reference.
                                                                (\w+)      # 3: The method or function name.
                                                        )
                                                        \(\)           # The () that signifies either a method or function.
@@ -396,6 +396,8 @@ class DevHub_Formatting {
                                                        // Reference to a class method.
                                                        if ( $matches[2] ) {
                                                                // Only link actually parsed methods.
+                                                               $matches[1] = str_replace( '->', '::', $matches[1] );
+
                                                                if ( $post = get_page_by_title( $matches[1], OBJECT, 'wp-parser-method' ) ) {
                                                                        return sprintf(
[...]

PR #77, opened 2022-06-03
Use a more forgiving regex to match more class-like references.

dd32:
See #50

Try to ignore the style and line changes, focus on the textual link changes.

| Before | After |
| --- | --- |
| Screen Shot 2022-06-01 at 4 55 29 pm | Screen Shot 2022-06-03 at 5 40 02 pm |
| Screen Shot 2022-06-01 at 4 55 43 pm | Screen Shot 2022-06-03 at 5 40 42 pm |

This new regex is slightly more complicated, slightly more likely to match unexpected pieces (which won't be replaced as they won't match a post), and slightly more expensive to run.

I think that's worth the tradeoffs.

### DIFF
--- a/source/wp-content/themes/wporg-developer/inc/formatting.php
+++ b/source/wp-content/themes/wporg-developer/inc/formatting.php
@@ -379,40 +379,50 @@ class DevHub_Formatting {
 				$content = " $piece "; // Pad with whitespace to simplify the regexes
 
 				// Only if the text contains something that might be a function.
-				if ( false !== strpos( $content, '()' ) ) {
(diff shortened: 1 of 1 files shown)


dd32:
Worth noting, that the resulting text is not matching 1:1 with what's in the source PHP, as `Class->Method` is replaced with `Class::Method()`. I don't think it'll change the meaning at all ever though.

dd32:
Upstream core docs change: https://core.trac.wordpress.org/ticket/55928

ryelle (approved):
I don't have the regex magic to be confident here, but the comments & named groups really help. It seems to be working in my testing 👍🏻

PR #79, opened 2022-06-03
Code blocks: Add "Copy" button to copy code snippets

### LINKED ISSUE #24: Code block should be easily copyable.
### Discussed in https://github.com/WordPress/wporg-developer/discussions/6

<sup>Originally posted by **StevenDufresne** May 22, 2022</sup>
Right now, if you copy the code block, you also copy the line numbers making it difficult to copy and paste code into your environment.

</div>

https://github.com/WordPress/wporg-developer/discussions/6#discussioncomment-2830159 has some implementation ideas

ryelle:
This adds a "Copy" button to each code block on a page, which will copy the code in the block to the user's clipboard. The button is hidden until you hover over or tab to the code block, so that it doesn't overlap any code.

Fixes #24 

## Screenshots

I didn't take any "before" shots because they'd all be the same as this first one.

By default, no button:

after-source

When you hover over, it appears.

after-source-hover

If you click copy, the button transforms to say "code copied", and if you're using a screen reader that's also announced.

after-source-copied

This also works when there are multiple code blocks on the page, each gets a copy button and they each copy only that block.

after-user-content-hover

### DIFF
--- a/source/wp-content/themes/wporg-developer/inc/user-content.php
+++ b/source/wp-content/themes/wporg-developer/inc/user-content.php
@@ -163,14 +163,45 @@ class DevHub_User_Submitted_Content {
 	 */
 	public static function scripts_and_styles() {
 		if ( is_singular() ) {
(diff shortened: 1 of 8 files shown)


dd32 (approved):
LGTM and works well from my testing.

PR #80, opened 2022-06-03
Fix alignment of filter buttons

ryelle:
This fixes the alignment and padding around the filter select & button on the by-file archive page.

| Before (production) | After |
|-------|-------|
| Screen Shot 2022-06-03 at 15 47 51 | Screen Shot 2022-06-03 at 15 47 47 |

**To test**

1. View any file archive page, ex: http://localhost:8888/reference/files/wp-includes/post-template.php/
2. The select & button should be lined up

Note: This is different on Safari, but I don't want to rewrite the whole select box styles, so I think this is fine.

Screen Shot 2022-06-03 at 3 54 13 PM

### DIFF
--- a/source/wp-content/themes/wporg-developer/scss/_global.scss
+++ b/source/wp-content/themes/wporg-developer/scss/_global.scss
@@ -159,23 +159,6 @@ img {
 	max-width: 100%; /* Adhere to container width. */
 }
 
-/* Form Elements */
-button,
-input,
-select,
-textarea {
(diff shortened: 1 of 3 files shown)


outdoor2kode (approved):
👍

StevenDufresne:
The padding on the right of the arrow is rather slim:



**Env:**
Chrome: Version 102.0.5005.61 (Official Build) (x86_64)
MacOs: Monterey, 12.2.1

PR #81, opened 2022-06-06
Make icon in developer resources page clickable

### LINKED ISSUE #55: Make Icon in Developer Resources page clickable
### Discussed in https://github.com/WordPress/wporg-developer/discussions/51

<sup>Originally posted by **renintw** June  1, 2022</sup>
I always try to click on the icon to access the page, but only the text link below is clickable.
Is it a good idea to also make the icon clickable? or even a bigger range as the img below shows?

image

</div>

outdoor2kode:
This PR makes the icon + text on the developer resources page clickable and fixes their alignment.

Fixes #55 

## Screenshot 
The file size is over the upload limit so links are provided directly here.

**Before (production)**
https://d.pr/i/l5ZFZN
**After**
https://d.pr/i/Jz8hja

## How to test
1. Go to http://localhost:8888/.
2. Check if the icon and its text are clickable.
3. Check if the redirect link is correct.

### DIFF
--- a/source/wp-content/themes/wporg-developer/scss/main.scss
+++ b/source/wp-content/themes/wporg-developer/scss/main.scss
@@ -539,6 +539,10 @@
 
 		.widget-title {
 			line-height: 2.4;
+			cursor: pointer;
+			text-decoration: none;
+			color: inherit;
+			font-weight: inherit;
 		}
 
 		&.search-guide {
--- a/source/wp-content/themes/wporg-developer/scss/main.scss
+++ b/source/wp-content/themes/wporg-developer/scss/main.scss
@@ -594,6 +594,9 @@
 				display: block;
 				margin: 0 auto;
 				opacity: 0.4;
+				display: flex;
+    			align-items: center;
+    			flex-direction: column;
 
 				&.dashicons-rest-api {
 					font-size: 9rem;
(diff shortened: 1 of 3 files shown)


source/wp-content/themes/wporg-developer/scss/main.scss
```diff
@@ -594,6 +598,9 @@
 				display: block;
 				margin: 0 auto;
 				opacity: 0.4;
+				display: flex;
+    			align-items: center;
+    			flex-direction: column;
```
dd32:
```suggestion
				display: flex;
				align-items: center;
				flex-direction: column;
```
Looks like some spaces slipped in there at the start.

dd32 (approved):
I'm not in love with having multiple links here, especially as It doubles the links a screen reader see's..

But this seems like the most straight forward way without completely redesigning how this page works, so the benefit seems worthwhile to me.

Minor whitespace issue in the CSS, but LGTM.

source/wp-content/themes/wporg-developer/scss/main.scss
outdoor2kode: Can't tell the difference in IDE without linting... Thanks for catching it!
Fixed.

ryelle (approved):
Looks good! I think the duplicate links are okay for now since the text is different.

PR #83, opened 2022-06-06
Enhance search bar UX

outdoor2kode:
This PR makes three simple changes to the search bar UX mentioned [here](https://github.com/WordPress/wporg-developer/issues/40#issuecomment-1144148805).
1. Make the Search Icon aligned with the global menu icon direction.
2. Make the focus state clear - the input ends before the icon.
3. Change the placeholder from `Search ...` to `Search in Code Reference ...`.

As to how to make the search button in the search bar more obvious, since we might be able to just use the search on top of the handbook sidebar ([comment](https://github.com/WordPress/wporg-developer/issues/40#issuecomment-1144451874)) or just move the search bar control into the site-title ([PR](https://github.com/WordPress/wporg-developer/pull/75)), I think it isn't worth spending time for the moment to figure out whether to implement any of the [proposals](https://github.com/WordPress/wporg-developer/issues/40#issuecomment-1144148805) or just remove the whole search bar.

## Screenshots
| Before | After |
| --- | --- |
| image | image |

## How to test
1. Go to any API detail page.
2. Click Search Icon
3. Check if the Search Icon Direction, Focus state, and Placeholder are correct.

### DIFF
--- a/source/wp-content/themes/wporg-developer/scss/main.scss
+++ b/source/wp-content/themes/wporg-developer/scss/main.scss
@@ -798,6 +798,10 @@
 		height: auto;
 		position: relative;
 
+		.search-bar {
+			display: flex;
+		}
+
 		input[type="text"] {
 			border-radius: 0;
 			margin: 0 auto;
@@ -806,18 +810,27 @@
 		}
 
 		.button-search {
-			background: transparent;
-			border: none;
+			background: #fff;
 			border-radius: 0;
 			box-shadow: none;
 			color: #32373c;
 			display: block;
-			height: 40px;
 			padding: 0.5rem 1rem;
-			position: absolute;
-			right: 0;
-			top: 0;
 			text-shadow: none;
+			border: 1px solid #ccc;
+			border-left: unset;
+			margin: 0;
+
+			&:focus {
+				outline-style: auto;
+				outline-color: get-color(blue-50);
+				z-index: 1;
+			}
+
+			&:active {
+				background: get-color(gray-5);
+				transform: translateY(0);
+			}
 		}
 
 		label {
--- a/source/wp-content/themes/wporg-developer/scss/main.scss
+++ b/source/wp-content/themes/wporg-developer/scss/main.scss
@@ -198,6 +198,7 @@
 		}
 
 		.dashicons {
+			transform: scale(-1, 1);
 			vertical-align: text-bottom;
 		}
 	}
@@ -767,6 +768,7 @@
 				.dashicons-search {
 					height: auto;
 					width: auto;
+					transform: scale(-1, 1);
 
 					&::before {
 						font-size: 36px;
--- a/source/wp-content/themes/wporg-developer/searchform.php
+++ b/source/wp-content/themes/wporg-developer/searchform.php
@@ -19,9 +19,11 @@
 	?>
 
 	<form role="search" method="get" class="searchform<?php echo esc_attr( $form_class ); ?>" action="<?php echo esc_url( $search_url ); ?>">
-		<label for="search-field" class="screen-reader-text"><?php _ex( 'Search for:', 'label', 'wporg' ); ?></label>
-		<input type="text" id="search-field" class="search-field" placeholder="<?php echo esc_attr_x( 'Search &hellip;', 'placeholder', 'wporg' ); ?>" value="<?php echo esc_attr( get_search_query() ); ?>" name="s">
-		<button class="button button-primary button-search"><i class="dashicons dashicons-search"></i><span class="screen-reader-text"><?php _e( 'Search plugins', 'wporg' ); ?></span></button>
+		<div class="search-bar">
+			<label for="search-field" class="screen-reader-text"><?php _ex( 'Search for:', 'label', 'wporg' ); ?></label>
+			<input type="text" id="search-field" class="search-field" placeholder="<?php echo esc_attr_x( 'Search in Code Reference &hellip;', 'placeholder', 'wporg' ); ?>" value="<?php echo esc_attr( get_search_query() ); ?>" name="s">
+			<button class="button button-primary button-search"><i class="dashicons dashicons-search"></i><span class="screen-reader-text"><?php _e( 'Search plugins', 'wporg' ); ?></span></button>
+		</div>
 	<?php if ( $filters ) : ?>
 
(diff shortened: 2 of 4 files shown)


tellyworth:
Looks good to me!

tellyworth (approved):

source/wp-content/themes/wporg-developer/scss/main.scss
```diff
@@ -201,6 +201,11 @@
 		}
 
 		.dashicons {
+			-moz-transform: scale(-1, 1);
+			-webkit-transform: scale(-1, 1);
+			-o-transform: scale(-1, 1);
+			-ms-transform: scale(-1, 1);
```
ryelle: Which browsers need these prefixes?

ryelle (changes requested):
I found a few issues:

1. There's an underline under the search button — this is Mac/Firefox
    underline
2. The input has a focus style, but the search button also needs one — in this screenshot I'm focused on the button, but there's no indicator.
    no-focus
3. The search bar overlaps the icon on smaller screens, for example at 500px:
    500-w

StevenDufresne:
@renintw Just as an FYI: I am working on an alternate approach where the search bar moves to match other wordpress.org themes: https://github.com/WordPress/wporg-developer/pull/75. It's an experiment so it may not take but I would love it if you could take a look and let me know what you think!

source/wp-content/themes/wporg-developer/scss/main.scss
outdoor2kode: https://caniuse.com/?search=transform
There are about 0.91% of users in total globally still using older browsers needing these prefixes, I used to just add these prefixes, but maybe it's not a good practice, should I just remove them?

(Just found that `-o-` is used for Opera 11.5 and no one is using that version anymore, I'm going to remove it first)

ryelle: Generally speaking we've followed [the `@wordpress/browserslist-config`](https://github.com/WordPress/gutenberg/tree/trunk/packages/browserslist-config) for which browsers to support.

<details>
<summary>Browser List</summary>

```
and_chr 101
android 101
chrome 101
chrome 100
chrome 99
edge 101
edge 100
firefox 100
firefox 99
ios_saf 15.4
ios_saf 15.2-15.3
ios_saf 14.5-14.8
op_mini all
opera 86
opera 85
safari 15.5
safari 15.4
samsung 16.0
```

</details>

The only one on this list that doesn't support unprefixed transform is Opera Mini, but it doesn't support transform at all, so we should be okay to drop the prefixes. In the past we've had people flag issues in older browsers (https://github.com/WordPress/wporg-mu-plugins/issues/159), so I'd also check the support for Chrome 87 and make sure it's not broken, at least. But we're fine with unprefixed transform here, too 👍🏻

outdoor2kode: Thanks for the insight sharing!
I've removed them and also tested it on Chrome 87, it looks good 👍 
Screen Capture on 2022-06-07 at 23-24-56

outdoor2kode:
> I found a few issues:
> 
> 1. There's an underline under the search button — this is Mac/Firefox
>    underline
> 2. The input has a focus style, but the search button also needs one — in this screenshot I'm focused on the button, but there's no indicator.
>    no-focus
> 3. The search bar overlaps the icon on smaller screens, for example at 500px:
>    500-w

Thanks for catching the issue on Firefox and on smaller screens, I've fixed them in the commit: [Fix incorrect style](https://github.com/WordPress/wporg-developer/pull/83/commits/a65809ca12604336deb1abd955338e11679483fc), could you help review them again, thanks! 
And after the review, I'll rebase it to the earlier [commit](https://github.com/WordPress/wporg-developer/pull/83/commits/26fda88dd7badf6f61aed5492338d654239e82f2).

Tested on Firefox and chrome as well as on different screen sizes: https://d.pr/i/rb0Nt2

source/wp-content/themes/wporg-developer/scss/main.scss
```diff
@@ -804,18 +810,21 @@
...
 			padding: 0.5rem 1rem;
-			position: absolute;
-			right: 0;
-			top: 0;
 			text-shadow: none;
+			border: 1px solid #ccc;
+			border-left: unset;
+			margin: 0;
+
+			&:active {
```
ryelle:
```suggestion
			&:focus {
```

The keyboard focus state is controlled with `:focus`. You can test this by using your keyboard, click into the search bar, then hit tab. It will move you to the button, but there's no visual indicator in the PR currently.

source/wp-content/themes/wporg-developer/scss/main.scss
```diff
@@ -804,18 +810,21 @@
...
+				outline-style: auto;
+				outline-color: rgb(0, 95, 204);
```
ryelle: Where does this color come from? Maybe instead, try one of [the colors in this list](https://github.com/WordPress/wporg-developer/blob/d2060892add88a6be095d4a5dd3af525a2882793/source/wp-content/themes/wporg-developer/scss/settings/_colors.scss#L12) to keep the palette consistent. Like this:

```suggestion
				outline-color: get-color(blue-60);
```

outdoor2kode: I got it from Chrome devTools directly.
image

> [the colors in this list](https://github.com/WordPress/wporg-developer/blob/d2060892add88a6be095d4a5dd3af525a2882793/source/wp-content/themes/wporg-developer/scss/settings/_colors.scss#L12) to keep the palette consistent

I like the idea, thanks for sharing it! 👍 However, after applying it, I found the color is a bit different in Chrome (it's hard to tell without zooming in or looking carefully into it though), and it's somehow the same in Firefox. I think I'll change it to use `get-color` here anyway as it seems to me it's just a slight difference.
image

Done Changing.

source/wp-content/themes/wporg-developer/scss/main.scss
outdoor2kode: Thanks for pointing it out!

Done Fixing and also changing the active style to the one like the search bar in the pattern-directory.
image

ryelle (approved):
Looks great now! Thanks for iterating on it :)

PR #85, opened 2022-06-06
Handbooks: Use code syntax block in handbook code examples

ryelle:
The handbooks use a `code` shortcode with a `lang` attribute to display the embedded code snippets. This PR updates that shortcode (which doesn't exist with SyntaxHighlighter Evolved disabled) to render the Code Syntax Block HTML instead. The new block also supports more languages when syntax highlighting (see the markdown example in the screenshot).

I also added a feature to the copy button that strips the leading `$` prompt from shell scripts, so they can be pasted directly into a terminal window.

Screenshots:

| Before | After |
|-------|------|
| before-php | after-php |
| before-js | after-js |
| before-jsx | after-jsx |
| before-markdown | after-markdown |

**To test**

- Manually run the cron tasks to import the handbook content, if you don't have it yet: `wp cron event run --all`
- View pages with code, the Block Editor handbook is good for this
- The code should be syntax-highlighted correctly, with line numbers and correct formatting
- The "copy" button should work
- Try looking at a block editor page with code tabs, ex: `/block-editor/how-to-guides/block-tutorial/block-controls-toolbar-and-sidebar/`
- The tabs should render correctly and work, and each code block should be copyable

View a page with shell commands, ex: http://localhost:8888/block-editor/reference-guides/packages/packages-env/

- Try copying a code snippet that's just one line
- It should strip the leading `$`
- Non-shell commands or multi-line text should not change

### DIFF
--- a/source/wp-content/themes/wporg-developer/inc/formatting.php
+++ b/source/wp-content/themes/wporg-developer/inc/formatting.php
@@ -49,6 +49,7 @@ class DevHub_Formatting {
 		add_shortcode( 'php', array( __CLASS__, 'do_shortcode_php' ) );
 		add_shortcode( 'js', array( __CLASS__, 'do_shortcode_js' ) );
 		add_shortcode( 'css', array( __CLASS__, 'do_shortcode_css' ) );
+		add_shortcode( 'code', array( __CLASS__, 'do_shortcode_code' ) );
 
 		add_filter(
 			'no_texturize_shortcodes',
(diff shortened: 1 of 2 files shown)


tellyworth (approved):
Looks good to me. I noticed there are some one- and two-liner code blocks in some handbooks that look a bit silly with line numbers. For example:

http://localhost:8888/coding-standards/wordpress-coding-standards/html/
Screen Shot 2022-06-07 at 5 22 25 pm

We could skip line numbers if <= 2, and/or make them lighter. That could be a separate PR though.

ryelle:
> We could skip line numbers if <= 2, and/or make them lighter. That could be a separate PR though.

Yeah, I was thinking something similar. I'll give that a shot today.

PR #87, opened 2022-06-07
WIP: Experiment: Advanced search qualifiers

StevenDufresne:
Mentioned in: [meta:3158](https://meta.trac.wordpress.org/ticket/3158).

Allow some basic search qualifiers to make searching a little bit more powerful.

**Filter by type**:
You can combine multiple types to filter.

- `type:hook`
- `type:function`
- `type:method`
- `type:class`

**Filter by file**:
You can filter within 1 or more files. Will match partial paths. Uses `name__like`.

- `file:wp-settings.php`

**Filter by version**:
You can filter within 1 version. Must be a full version.

- `version:4.8.0`

<hr>

**Examples**:
Search: `version:4.8.0 file:class-wp-editor.php get`

See version `4.8.0`, file: `class-wp-editor.php`, looking for functions that have `get` in them.

**Returns:**
- [_WP_Editors::get_baseurl()](http://developer.wordpress.org/reference/classes/_wp_editors/get_baseurl/)
- [_WP_Editors::get_mce_locale()](http://developer.wordpress.org/reference/classes/_wp_editors/get_mce_locale/)

### DIFF
--- a/source/wp-content/themes/wporg-developer/inc/advanced-search-filters.php
+++ b/source/wp-content/themes/wporg-developer/inc/advanced-search-filters.php
@@ -1,0 +1,102 @@
+<?php
+
+class Advanced_Search_Filters {
+
+	public static function init() {
+		add_filter( 'the_posts', array( __CLASS__, 'modify_search' ), 10, 2 );
+	}
+
+	/**
+	 * Modifies the query if user has used advanced search filters
+	 *
+	 * @param \WP_Query $query
+	 * @return void
+	 */
+	public static function modify_query( $query ) {
+		$post_types = [];
+		$tax_queries = [];
+
+		// Clean up keyword
+		$keyword = trim( $query->get( 's' ) );
+
+		// Split on spaces
+		$groups = explode( " ", $keyword );
+
+		// Loop through groups
+		foreach( $groups as $group ){
+			// Check if {qualifier}:{keyword} is used;
+			$split = explode( ':', $group );
+
+			// We have mo qualifier
+			if( count( $split ) < 2 ) {
+				continue;
+			}
+	
+			$qualifier = strtolower( $split[ 0 ] );
+		
+			switch ( $qualifier ) {
+				case 'type':
+					// TO DO Validate the type
+					$post_types[] = 'wp-parser-'. strtolower($split[1]);
+
+					// Remove "type"	
+					$keyword = str_replace( $qualifier . ':' , '', $keyword );
+					break;
+				case 'file':
+					// Get the tax ids first
+					$tax_ids = get_terms( array(
+							'taxonomy' => 'wp-parser-source-file',
+							'name__like' => $split[1],
+							'fields' => 'ids'
+						) );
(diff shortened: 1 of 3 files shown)


source/wp-content/themes/wporg-developer/inc/advanced-search-filters.php
```diff
@@ -0,0 +1,147 @@
...
+			if ( ! isset( $split[1] ) ||
+				empty( $split[0] ) || // Missing a qualifier. Ie: :init
+				$split[1][0] == ':'  // Searching for a class method. Ie: {Class}::init()
+			) {
+
+				// If user has '()' at end of a search string, assume they want a specific function/method.
+				$s = htmlentities( $split[0] );
+				if ( str_contains( $s, '()' ) ) {
+					// Modify the search query to omit the parentheses.
+					$keyword = str_replace( '()', '', $keyword );
```
dd32:
```suggestion
				if ( str_contains( $s, '(' ) ) {
					// Modify the search query to omit the parentheses.
					$keyword = rtrim( $keyword, '()' );
```
This should make it compat with the changes in https://github.com/WordPress/wporg-developer/commit/3c9d9d12a3c5539d7728d648c3fbaa934535a380

source/wp-content/themes/wporg-developer/inc/advanced-search-filters.php
```diff
@@ -0,0 +1,147 @@
...
+			if ( ! isset( $split[1] ) ||
+				empty( $split[0] ) || // Missing a qualifier. Ie: :init
+				$split[1][0] == ':'  // Searching for a class method. Ie: {Class}::init()
+			) {
+
+				// If user has '()' at end of a search string, assume they want a specific function/method.
+				$s = htmlentities( $split[0] );
+				if ( str_contains( $s, '(' ) ) {
+					// Modify the search query to omit the parentheses.
+					$keyword = rtrim( $keyword, '()' );
```
dd32: Oh.. Well this breaks some of this code :D 

I guess I should've done this instead, @StevenDufresne does my change look proper this time?
```suggestion
				if ( str_contains( $s, '(' ) ) {
					// Modify the search query to omit the parentheses.
					$keyword = str_replace( array( '()', '(' ), '', $keyword );
```

StevenDufresne: I've updated the code to use the `str_replace` again. It appears to work. 👍

StevenDufresne:
I think this works well, the only problem I don't have a solution for is that the search query and the filter boxes can fall out of sync seeing that we don't automatically append the advanced search qualifiers when using the filter boxes. 

For example:

1. Start with a query `type:hook get`
2. Search
3. Click `[] Methods`

It will apply the filter but the search keyword will still be `type:hook get` but in reality it will be searching like:`type:hook type:method get`.

I don't think we want to add/remove to the searchbox everytime a checkbox is clicked though... 🤔

PR #88, opened 2022-06-07
Restore the code collapse & expand buttons

### LINKED ISSUE #22: Code blocks should not be collapsed.
### Discussed in https://github.com/WordPress/wporg-developer/discussions/4

When we provide code for users, we should assume that all code is relevant and that it's better to see all the code than just a portion of it. I'm sure there are instances where functions are too long. In those cases, maybe we can collapse, or consider not showing the block.

### Screenshot
</div>

### LINKED ISSUE #86: Source block is very long on some pages, no expand/collapse
http://localhost:8888/reference/classes/wp_query/get_posts/

..currently shows a Source block with 3k+ lines fully expanded. There are no expand/collapse buttons.

There's an issue #22 for changes to the expand/collapse logic. I think that was referring more to examples the description rather than the Source block itself. Either way we should implement some kind of height limit to the source block, whether or not it expands.

tellyworth:
This just restores the old code, with some adjustments to work with the new markup. See #22

### DIFF
--- a/source/wp-content/themes/wporg-developer/js/function-reference.js
+++ b/source/wp-content/themes/wporg-developer/js/function-reference.js
@@ -7,6 +7,74 @@
 
 jQuery( function ( $ ) {
 	let $usesList, $usedByList, $showMoreUses, $hideMoreUses, $showMoreUsedBy, $hideMoreUsedBy;
+	let $sourceCollapsedHeight;
+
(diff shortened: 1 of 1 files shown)


outdoor2kode (approved):
Tested on Chrome and Firefox. 👍

PR #90, opened 2022-06-07
Iterate on anchor links on headings

ryelle:
I noticed some issues with the heading positioning when the link was clicked, so I decided to try to fix it. While I was here, I made a few other tweaks too.

<details>
<summary>The initial bug:</summary>

https://user-images.githubusercontent.com/541093/172435040-0e2f0dfe-44b8-4b68-9383-298a207f138b.mp4

I didn't take an "after" screencast but the jumping is fixed :)
</details>

**Screenshots**

Visual changes:

- The top link & heading are wrapped in a `header` to put them on a single line. I see this is only the style on the code reference, so I can change this part back if it's not a desired change.
- Made the default link icon a lighter grey (icons can have a contrast of 3:1, this is the lightest grey we'd want to use here)
- When hovering/focusing, the link icon goes blue
- When a section is targeted (the URL has the `#create-a-block-tutorial`), the icon is black — I thought this would be a subtle way to draw attention, but we could go bigger here somehow. It looks like at some point this had a blue line like [the handbooks on p2](https://make.wordpress.org/meta/handbook/#documentation), but with the persistent link icon, this doesn't work.
- Remove the overflow from the source section, which was hiding the link icon. the `pre` adds its own overflow, so the one on the section was not needed.

| | Before | After |
|---|--------|-------|
| Default code reference | before-code-ref | after-code-ref |
| Heading targeted | before-code-ref-target | after-code-ref-target |
| One heading targeted, one hovered | before-handbook | after-handbook |

And for smaller screens, I shrunk down the icon a little to fit.

| Before | After |
|--------|-------|
| before-code-ref-small | after-code-ref-small |
| before-handbook-small | after-handbook-small |

### DIFF
--- a/source/wp-content/plugins/handbook/inc/table-of-contents.php
+++ b/source/wp-content/plugins/handbook/inc/table-of-contents.php
@@ -8,7 +8,7 @@ class WPorg_Handbook_TOC {
 	protected $post_types = array();
 
 	protected $styles = '<style>
-		.toc-header { display: flex; justify-content: space-between; }
+		.toc-header { display: flex; justify-content: space-between; margin-top: 48px !important; }
 		.toc-jump { text-align: right; font-size: 0.75em; order: 2; }
 		.toc-heading a { color: inherit; font-weight: inherit; margin-left: -32px; text-decoration: none !important; }
 		/* icon is 20px wide in a 32px space, so add a 12px margin. */
--- a/source/wp-content/plugins/handbook/inc/table-of-contents.php
+++ b/source/wp-content/plugins/handbook/inc/table-of-contents.php
@@ -209,7 +209,7 @@ class WPorg_Handbook_TOC {
 				$extra_attrs,
 				$title
 			);
-			$replacements[] = $replacement;
+			$replacements[] = '<header class="toc-header">' . $replacement . '</header>';
 		}
 
 		if ( $replacements ) {
--- a/source/wp-content/plugins/handbook/inc/table-of-contents.php
+++ b/source/wp-content/plugins/handbook/inc/table-of-contents.php
@@ -8,10 +8,16 @@ class WPorg_Handbook_TOC {
 	protected $post_types = array();
 
 	protected $styles = '<style>
-		.toc-jump { text-align: right; font-size: 12px; }
-		.toc-heading a { color: inherit; font-weight: inherit; margin-left: -25px; text-decoration: none !important; }
-		.toc-heading a:before { vertical-align: middle; margin-top: -5px; margin-right: 5px; }
-		.toc-heading:target a:before { margin-left: -8px; margin-right: 13px; }
+		.toc-header { display: flex; justify-content: space-between; }
+		.toc-jump { text-align: right; font-size: 0.75em; order: 2; }
+		.toc-heading a { color: inherit; font-weight: inherit; margin-left: -32px; text-decoration: none !important; }
+		/* icon is 20px wide in a 32px space, so add a 12px margin. */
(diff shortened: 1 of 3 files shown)


tellyworth:
@adamwoodnz does this affect the ToC PR you've been working on?

adamwoodnz:
> @adamwoodnz does this affect the ToC PR you've been working on?

No conflicts that I can see, this is a really nice improvement IMO, 👍 LGTM

source/wp-content/plugins/handbook/inc/table-of-contents.php
```diff
@@ -8,10 +8,16 @@ class WPorg_Handbook_TOC {
...
+		.toc-header { display: flex; justify-content: space-between; margin-top: 48px !important; }
+		.toc-jump { text-align: right; font-size: 0.75em; order: 2; }
+		.toc-heading a { color: inherit; font-weight: inherit; margin-left: -32px; text-decoration: none !important; }
+		/* icon is 20px wide in a 32px space, so add a 12px margin. */
+		.toc-heading a:before { vertical-align: middle; margin: -4px 8px 0 4px; }
+		@media (max-width: 876px) {
+			.toc-heading a { margin-left: -20px; }
+			/* icon is 16px wide in a 20px space, so add a 4px margin. */
+			.toc-heading a:before { margin: -2px 4px 0 2px; width: 14px; height: 14px; font-size: 14px; }
+		}
```
adamwoodnz: Not a major, but it would be easier to see individual property changes in future if these styles weren't all written on one line

source/wp-content/plugins/handbook/inc/table-of-contents.php
```diff
@@ -8,10 +8,16 @@ class WPorg_Handbook_TOC {
...
 	protected $styles = '<style>
-		.toc-jump { text-align: right; font-size: 12px; }
-		.toc-heading a { color: inherit; font-weight: inherit; margin-left: -25px; text-decoration: none !important; }
-		.toc-heading a:before { vertical-align: middle; margin-top: -5px; margin-right: 5px; }
-		.toc-heading:target a:before { margin-left: -8px; margin-right: 13px; }
+		.toc-header { display: flex; justify-content: space-between; margin-top: 48px !important; }
+		.toc-jump { text-align: right; font-size: 0.75em; order: 2; }
+		.toc-heading a { color: inherit; font-weight: inherit; margin-left: -32px; text-decoration: none !important; }
+		/* icon is 20px wide in a 32px space, so add a 12px margin. */
+		.toc-heading a:before { vertical-align: middle; margin: -4px 8px 0 4px; }
```
adamwoodnz: Seems like this comment is mismatched with the margin below

adamwoodnz (commented):
Tried rebasing this on my ToC repositioning PR and ran it. Everything looks good 👍 

Couple of minor comments inline.

source/wp-content/plugins/handbook/inc/table-of-contents.php
ryelle: The `12px` refers to the horizontal margin, `8px + 4px`. I'll make that more clear.

source/wp-content/plugins/handbook/inc/table-of-contents.php
ryelle: I was trying to keep my changes minimal, but by the time I was done I guess I could have reformatted the whole thing too 😅

ryelle: Reformatted this to be expanded 👍🏻

PR #92, opened 2022-06-07
Skip Link: Remove custom skip link in favor of mu-plugin

ryelle:
This removes the custom skip link functionality for [the skip-to `mu-plugin`](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/skip-to/skip-to.php). It also repositions the link so it can be used to skip over the menu.

I recorded before/after screencasts, and used an app to show my keypresses on the screen, you'll see tab & enter.

Before, you need to tab through the admin bar, menu, and header before hitting the skip link, so it's not actually helping you skip anything. Also, the destination (`#main`) doesn't exist on the homepage, so it does nothing.

https://user-images.githubusercontent.com/541093/172479879-cba99bca-ead7-4a5b-9588-fb7f4d824b15.mp4

After, the skip link is the first item to focus on the page. Activating it jumps down to the white background section.

https://user-images.githubusercontent.com/541093/172480032-7a571e9b-5d94-42e9-81d8-784743464d78.mp4


StevenDufresne:
This is working as expected for me. Feel free to merge if you feel it's ready.

Issue #76, opened 2022-06-03
Source code preview for Classes is too narrow


dd32:
The source-code displayed for classes is too narrow, caused by the table of contents.

It might make more sense to display the Method list before the Source code too, as that's likely more important to the reader.

<img width="895" alt="Screen Shot 2022-06-03 at 5 00 00 pm" src="https://user-images.githubusercontent.com/767313/171803938-969c1e4b-f349-4094-8092-35322f8f2538.png">

adamwoodnz:
Some options:
- float the contents top right when the browser is wide enough, otherwise hide it visually, leaving it still accessible by screen readers
- move the contents alongside the class title (may work better if the search button moves into the header
- include it in a sidebar on the left

tellyworth:
I'd also add the possibility:

* Remove the TOC entirely

iandunn:
Here's how some other docs handle it:

* https://jestjs.io/docs/mock-function-api - collapses
* https://reactjs.org/docs/hooks-reference.html#gatsby-focus-wrapper - always shows
* https://nodejs.org/dist/latest-v18.x/docs/api/http.html - always shows
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter - moves into sidebar

I think the MDN design is an interesting idea. I don't like the idea of hiding it since it feels useful to me, and even more useful when I'm on a smaller device and scrolling is worse (have to use gesture, page is longer, no pageup/down).

PR #93, opened 2022-06-08
Extract Table of Contents from content to allow wider source view

adamwoodnz:
See #76 

This PR moves the Table of Contents from the content of the page and inserts it into the header.

- On wide screens it's fixed in the top right
- On medium screens it's within the header
- On small screens it's at the top of the page

A further improvement might be to add expand/collapse behaviour and collapse by default on small screens. Although that might mask the utility of it for jumping down a long page.

Wide
Screen Shot 2022-06-08 at 2 34 38 PM

Medium
Screen Shot 2022-06-10 at 11 10 45 AM

Small
Screen Shot 2022-06-10 at 11 09 18 AM

### DIFF
--- a/source/wp-content/plugins/handbook/inc/table-of-contents.php
+++ b/source/wp-content/plugins/handbook/inc/table-of-contents.php
@@ -101,7 +101,7 @@ class WPorg_Handbook_TOC {
 	 * @access public
 	 *
 	 * @param string $content Content.
-	 * @return string Modified content.
+	 * @return array toc => Table of Contents, content => Modified Content.
 	 */
 	public function add_toc( $content ) {
 		if ( ! in_the_loop() ) {
@@ -164,7 +164,7 @@ class WPorg_Handbook_TOC {
 
 		$toc .= "</ul>\n</div>\n";
 
-		return $toc . $content;
+		return array('toc' => $toc, 'content' => $content);
 	}
 
 	/**
--- a/source/wp-content/plugins/handbook/inc/table-of-contents.php
+++ b/source/wp-content/plugins/handbook/inc/table-of-contents.php
@@ -101,18 +101,31 @@ class WPorg_Handbook_TOC {
 	 * @access public
 	 *
 	 * @param string $content Content.
-	 * @return array toc => Table of Contents, content => Modified Content.
+	 * @return string toc => Modified Content.
 	 */
 	public function add_toc( $content ) {
+		$parts = $this->parse_content( $content );
+   		return $parts['toc'] . $parts['content'];
+	}
+
+	/**
+	 * Parses given content and returns modified content and the ToC.
+	 *
+	 * @access public
+	 *
+	 * @param string $content Content.
+	 * @return array toc => Table of Contents, content => Modified Content.
+	 */
+	public function parse_content( $content ) {
 		if ( ! in_the_loop() ) {
-			return $content;
+			return array('content' => $content, 'toc' => null);
 		}
 
 		$toc   = '';
 		$items = $this->get_tags( 'h(?P<level>[1-4])', $content );
 
 		if ( count( $items ) < 2 ) {
-			return $content;
+			return array('content' => $content, 'toc' => null);
 		}
 
 		// Remove any links we don't need.
@@ -138,7 +151,7 @@ class WPorg_Handbook_TOC {
 		$content = $this->add_ids_and_jumpto_links( $items, $content );
 
 		if ( ! apply_filters( 'handbook_display_toc', true ) ) {
-			return $content;
+			return array('content' => $content, 'toc' => null);
 		}
 
 		$contents_header = 'h' . reset( $items )['level']; // Duplicate the first <h#> tag in the document for the TOC header
--- a/source/wp-content/plugins/handbook/inc/table-of-contents.php
+++ b/source/wp-content/plugins/handbook/inc/table-of-contents.php
@@ -154,9 +154,8 @@ class WPorg_Handbook_TOC {
 			return array('content' => $content, 'toc' => null);
 		}
 
-		$contents_header = 'h' . reset( $items )['level']; // Duplicate the first <h#> tag in the document for the TOC header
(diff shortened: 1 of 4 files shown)


source/wp-content/plugins/handbook/inc/table-of-contents.php
```diff
@@ -164,7 +164,7 @@ public function add_toc( $content ) {
 
 		$toc .= "</ul>\n</div>\n";
 
-		return $toc . $content;
+		return array('toc' => $toc, 'content' => $content);
```
dd32, line 179: My suggestion would be to rename this from `add_toc` to something like `parse_content`, then add a new `add_toc` method which is similar to:
```
function add_toc( $content ) {
   $parts = $this->parse_content( $content );
   return $parts['toc'] . $parts['content'];
}
```

That way the function syntax remains unchanged for other uses of the plugin.. and for this filter in the plugin: `add_filter( 'the_content', array( $this, 'add_toc' ) );` which would expect a stringy return.

adamwoodnz: Done! Thanks for the suggestion. Tested that this works with the old `content-reference.php`

source/wp-content/plugins/handbook/inc/table-of-contents.php
```diff
@@ -138,12 +151,11 @@ public function add_toc( $content ) {
 		$content = $this->add_ids_and_jumpto_links( $items, $content );
 
 		if ( ! apply_filters( 'handbook_display_toc', true ) ) {
-			return $content;
+			return array('content' => $content, 'toc' => null);
 		}
 
-		$contents_header = 'h' . reset( $items )['level']; // Duplicate the first <h#> tag in the document for the TOC header
```
adamwoodnz, line 144: The header shouldn't be an `h` tag now that the ToC is before the H1

StevenDufresne:
I wonder if we should keep it closer to the content as opposed to the browser's edge:

```
// Something like
right: calc(((100vw - 60em ) / 2 ) - ( 250px - 15px )  )
```
[See Example](https://d.pr/v/KhFSme.mp4)

adamwoodnz:
> I wonder if we should keep it closer to the content as opposed to the browser's edge:
> 
> ```
> // Something like
> right: calc(((100vw - 60em ) / 2 ) - ( 250px - 15px )  )
> ```
> 
> [See Example](https://d.pr/v/KhFSme.mp4)

Yeah I think that's a good call

tellyworth:
Sounds like this is broken on other handbook pages currently. Is it easily fixable or should we punt this from the milestone?

iandunn:
> On medium screens it's within the header

Other handbooks have variable-length ToCs, and some are much longer:

* https://developer.wordpress.org/plugins/wordpress-org/detailed-plugin-guidelines/
* https://developer.wordpress.org/rest-api/frequently-asked-questions/
* https://developer.wordpress.org/themes/basics/template-hierarchy/
* https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/

Screen Shot 2022-06-10 at 9 12 36 AM

On those, it might look better to make it take 100% width of `main.site-main`, like it does for mobile.

adamwoodnz:
> Sounds like this is broken on other handbook pages currently. Is it easily fixable or should we punt this from the milestone?

Correct, the other pages are still calling `add_toc` directly via this [filter](https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/plugins/handbook/inc/table-of-contents.php#L85):
```
        public function load_filters() {
		if ( is_singular( $this->post_types ) && ! is_embed() ) {
			add_filter( 'the_content', array( $this, 'add_toc' ) );
		}
	}
```

The result is that on medium screens where the ToC is displayed inline the layout is affected.

Screen Shot 2022-06-15 at 2 18 08 PM

For accessibility reasons my change in the reference handbook involves moving the ToC before the H1 and adding some layout markup.

`content_handbook.php` simply calls these functions:
```
<h1><?php the_title(); ?></h1>

<?php
/*
 * Content
 */
?>

<?php the_content( __( '(More ...)' , 'wporg' ) ); ?>
```

So I've been trying to do something using filters for `the_content` and `the_title` to split the ToC parts and render them separately. The content filter works, but I've hit a wall trying to get the ToC part within `the_title`:

```
        public function load_filters() {
		if ( is_singular( $this->post_types ) && ! is_embed() ) {
			add_filter( 'the_content', array( $this, 'content_with_links' ) );
			add_filter( 'the_title', array( $this, 'title_with_toc' ) );
		}
	}

	function content_with_links( $content ) {
		$parts = $this->parse_content( $content );

		return $parts['content'];
	}

	function title_with_toc( $title ) {
		// How do I get the content?
		$parts = $this->parse_content( $content );

		return $title . $parts['toc'];
	}
```

I did try saving the $toc value on the class instance but it still wan't available in `the_title`, I guess due to lifecycle but my WordPress knowledge is limited here.

EDIT:
Potentially moving the ToC above the H1 (away from the content) is not necessary. @dd32 you originally pointed to this doc https://www.w3.org/TR/WCAG20-TECHS/G64.html for guidance but I don't see a requirement there. Did you read this somewhere else perhaps?

adamwoodnz:
> > On medium screens it's within the header
> 
> Other handbooks have variable-length ToCs, and some are much longer:
> 
> * https://developer.wordpress.org/plugins/wordpress-org/detailed-plugin-guidelines/
> * https://developer.wordpress.org/rest-api/frequently-asked-questions/
> * https://developer.wordpress.org/themes/basics/template-hierarchy/
> * https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/
> 
> Screen Shot 2022-06-10 at 9 12 36 AM
> 
> On those, it might look better to make it take 100% width of `main.site-main`, like it does for mobile.

👍 This is broken on this page but I have made it full width on the Content Reference page

adamwoodnz:
Closed in favor of https://github.com/WordPress/wporg-developer/pull/122

Issue #45, opened 2022-05-31
Parameters list HTML is a little broken on `WP_Query::parse_query`


ryelle:
See https://developer.wordpress.org/reference/classes/WP_Query/parse_query/ — when the list in `orderby` is output, the indentation afterwards is off. 

See how the parameters are indented twice, until after `orderby`, when they're no longer indented enough and now they have bullets:

<details>
<summary>Very long screenshot of the params list</summary>

![Screen Shot 2022-05-31 at 18 58 18-fullpage](https://user-images.githubusercontent.com/541093/171298145-f4328588-73dc-419a-81b1-050c5421cdc3.png)

</details>

dd32:
Responsible code for lists in params:
https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer/inc/formatting.php#L479-L496

Looks like we don't support lists within lists, `Does not handle nesting of lists.`.

dd32:
Interestingly, there's also some issues in the parsing of some of the fields.

For `$fields` the final line of the documentaion, `Default ''` is missing:
https://github.com/WordPress/wordpress-develop/blob/6.0/src/wp-includes/class-wp-query.php#L656-L661

I've done more digging, and the way the PHPDoc source is imported causes some things to be lost.

For example, the `orderby` section imported lacks the newlines ([They're stripped out](https://github.com/WordPress/phpdoc-parser/blob/master/lib/runner.php#L136-L175)):
```
@type string\|array    $orderby                 Sort retrieved posts by parameter. One or more options may be passed.                                                    To use 'meta_value', or 'meta_value_num', 'meta_key=keyname' must be                                                    also be defined. To sort by a specific `$meta_query` clause, use that                                                    clause's array key. Accepts:                                                    - 'none'                                                    - 'name'                                                    - 'author'                                                    - 'date'                                                    - 'title'                                                    - 'modified'                                                    - 'menu_order'                                                    - 'parent'                                                    - 'ID'                                                    - 'rand'                                                    - 'relevance'                                                    - 'RAND(x)' (where 'x' is an integer seed value)                                                    - 'comment_count'                                                    - 'meta_value'                                                    - 'meta_value_num'                                                    - 'post__in'                                                    - 'post_name__in'                                                    - 'post_parent__in'                                                    - The array keys of `$meta_query`.                                                    Default is 'date', except when a search is being performed, when                                                    the default is 'relevance'.
```

[...]

PR #94, opened 2022-06-08
Merge PHPDoc Parser into this repo

dd32:
See #45 & #49 for the need for this.

This merges the upstream repo into this repo so that we can track everything in one mono-repo rather than having it spread out through multiple repo's.

I've kept the history (I wish I did that when I created this repo originally..)


GaryJones:
Does that mean https://github.com/WordPress/phpdoc-parser is no longer maintained, and is no longer standalone for folks to parse their own plugins and themes to generate their own developer references?

Or will wporg-developer pull in fresh snapshots of phpdoc-parser once new releases are made there (in which case Composer may have been a better approach)?

Issue #95, opened 2022-06-08
Expand argument arrays in the function reference


tellyworth:
### Discussed in https://github.com/WordPress/wporg-developer/discussions/19

<div type='discussions-op-text'>

<sup>Originally posted by **tellyworth** May 27, 2022</sup>
Many WP functions use some form of `$args` array parameter, which leads to docs like this:

https://developer.wordpress.org/reference/functions/wp_update_post/
<img width="1049" alt="Screen Shot 2022-05-27 at 4 44 08 pm" src="https://user-images.githubusercontent.com/7200686/170645925-c22c8dbf-0142-4c1d-917b-236a81f3d87f.png">

It would be great to pull in and expand the details of the array arguments so it's not necessary to click through to another page.
</div>

iandunn:
I think this was fixed by #96

PR #96, opened 2022-06-08
Expand `$args` array parameters that are described in other functions

tellyworth:
Quick WIP version. See #95.

Before, on the get_terms() function reference:
Screen Shot 2022-06-08 at 4 13 40 pm

After, collapsed (detail/summary tag - note the teeny tiny triangle):
Screen Shot 2022-06-08 at 4 15 11 pm

After, expanded:
Screen Shot 2022-06-08 at 4 15 25 pm

Before, on get_posts(). Note that this one has a few arguments explicitly documented:
Screen Shot 2022-06-08 at 4 21 35 pm

After, collapsed:
Screen Shot 2022-06-08 at 4 21 51 pm

After, expanded (note that it still has the explicit docs first, I haven't hidden that):
Screen Shot 2022-06-08 at 4 22 28 pm

There are definitely some improvements needed especially on the front-end. But this seems to work pretty well at recursing through and finding the correct array. For example it works with multiple levels of indirection, like `wp_count_terms($args) -> get_terms($args) -> WP_Term_Query::__construct($query)`.

### DIFF
--- a/source/wp-content/themes/wporg-developer/inc/formatting.php
+++ b/source/wp-content/themes/wporg-developer/inc/formatting.php
@@ -524,32 +524,9 @@ class DevHub_Formatting {
 			return $text;
 		}
 
-		// Convert dashes to a list.
-		// Example: https://developer.wordpress.org/reference/classes/wp_term_query/__construct/
-		// Example: https://developer.wordpress.org/reference/hooks/password_change_email/
-		if ( false !== strpos( $text, ' - ' ) )  {
-			// Display as simple plaintext list.
-			$text = str_replace( ' - ', "\n" . $li, $text );
-			$inline_list = true;
-		}
-
-		// If list detected.
-		if ( $inline_list ) {
-			// Replace first item, ensuring the opening 'ul' tag is prepended.
-			$text = preg_replace( '~^' . preg_quote( $li ) . '(.+)$~mU', "<ul><li>\$1</li>\n", $text, 1 );
-			// Wrap subsequent list items in 'li' tags.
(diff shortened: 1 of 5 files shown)


ryelle:
I pushed up a commit to remove the `summary` wrapper from the description — that was stripping the semantics from everything inside it. It's best to just have the content in `summary` be a single text node (or heading).

I just changed it so it's "More Arguments". Feel free to change that to whatever works better :)

An example on `get_posts`:

Screen Shot 2022-06-08 at 4 26 53 PM

And on `wp_count_terms`:

Screen Shot 2022-06-08 at 4 27 03 PM

I'm also seeing `<br>` on all the params now, I think that's something new on this branch.

tellyworth:
Thanks!

I'm not seeing the `<br>` thing at all. Not sure what's up with that. Some kind of nl2br formatting, maybe on import? I re-ran an import and it didn't reproduce.

I pushed a couple of commits riffing on this a bit. The summary now looks more obviously clickable, and explains where the extra info came from:
Screen Shot 2022-06-09 at 5 12 44 pm

It's a bit better but still needs some work. Might be almost good enough to merge now for a designer to look at later - what do you think? Assuming I can fix the br thing.

ryelle (approved):
> I'm not seeing the <br> thing at all. Not sure what's up with that. Some kind of nl2br formatting, maybe on import? I re-ran an import and it didn't reproduce.

🤷🏻  I'm not seeing them on trunk, just this branch. Ah okay — I just rebased and they're gone now. I'm not sure why you weren't seeing them but when this merges it should be fine, at least.

> Might be almost good enough to merge now for a designer to look at later - what do you think?

That works for me. I don't think it should look like a link (since it's not taking you anywhere), but we can iterate on that later.

PR #97, opened 2022-06-08
Mark some code-like items as code.

dd32:
This PR marks up some code-like references as code.

| Before | After |
| --- | --- |
| Hash placeholders |
| Screen Shot 2022-06-08 at 6 15 04 pm | Screen Shot 2022-06-08 at 6 15 25 pm |
| Pipe delimited types |
|  Screen Shot 2022-06-08 at 6 16 25 pm | Screen Shot 2022-06-08 at 6 16 31 pm |
| 'Quoted' strings. | I admit this one can look loud in some areas. |
| Screen Shot 2022-06-08 at 6 17 17 pm | Screen Shot 2022-06-08 at 6 17 28 pm |

### DIFF
--- a/source/wp-content/themes/wporg-developer/inc/formatting.php
+++ b/source/wp-content/themes/wporg-developer/inc/formatting.php
@@ -40,6 +40,7 @@ class DevHub_Formatting {
 		add_filter( 'devhub-format-description', array( __CLASS__, 'fix_param_hash_formatting' ), 9 );
 		add_filter( 'devhub-format-description', array( __CLASS__, 'fix_param_description_html_as_code' ) );
 		add_filter( 'devhub-format-description', array( __CLASS__, 'convert_lists_to_markup' ) );
+		add_filter( 'devhub-format-hash-param-description', array( __CLASS__, 'fix_param_description_quotes_to_code' ) );
 
 		add_filter( 'devhub-format-hash-param-description', array( __CLASS__, 'autolink_references' ) );
 		add_filter( 'devhub-format-hash-param-description', array( __CLASS__, 'fix_param_description_parsedown_bug' ) );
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-developer/inc/formatting.php
```diff
@@ -668,6 +669,40 @@ public static function fix_param_description_html_as_code( $text ) {
...
+			if ( str_starts_with( $piece, '<' ) ) {
+				$text .= $piece;
+				continue;
+			}
+
+			// Pipe delimited types inline.
+			$piece = preg_replace( "/(([\w'\[\]]+\|)+[\w'\[\]]+)/", '<code>$1</code>', $piece, -1 );
+
+			// Quoted strings.
+			$piece = preg_replace( "/('[^']{0,20}')/", '<code>$1</code>', $piece, -1 );
```
ryelle: I think the 20 character limit might be too short:

Screen Shot 2022-06-08 at 11 32 33 AM

source/wp-content/themes/wporg-developer/inc/formatting.php
```diff
@@ -40,6 +40,7 @@ public static function do_init() {
 		add_filter( 'devhub-format-description', array( __CLASS__, 'fix_param_hash_formatting' ), 9 );
 		add_filter( 'devhub-format-description', array( __CLASS__, 'fix_param_description_html_as_code' ) );
 		add_filter( 'devhub-format-description', array( __CLASS__, 'convert_lists_to_markup' ) );
+		add_filter( 'devhub-format-hash-param-description', array( __CLASS__, 'fix_param_description_quotes_to_code' ) );
```
ryelle: Should this also be added to `devhub-format-description`?

source/wp-content/themes/wporg-developer/inc/formatting.php
dd32: I originally chose 20 as something low to avoid mis-matching and generating a string like `Let<code>'s figure out a good length for the '</code>' quoting`. It would probably be better to just look for `''` without any spaces between them though perhaps?

Iteration is definitely needed here!

source/wp-content/themes/wporg-developer/inc/formatting.php
dd32: Yes, I think it does. Or more specifically, I think it should probably be moved to there, and only there now.

Originally I was having trouble with the amount of HTML in `devhub-format-description` (as it runs over the combined outputs from `devhub-format-hash-param-description`) causing it to match too much.

Now that it's splitting by HTML tags, that'll probably now work with less fuss, but it'll likely also need to detect when it's within `<pre>` or `<code>` tags and not affect the internal string values of those functions..

Thanks for mentioning this!

source/wp-content/themes/wporg-developer/inc/formatting.php
dd32: [8028739](https://github.com/WordPress/wporg-developer/pull/97/commits/8028739bd2f085e69ca99ce72fd1a6820c38cb42)

source/wp-content/themes/wporg-developer/inc/formatting.php
dd32: [3ea0d11](https://github.com/WordPress/wporg-developer/pull/97/commits/3ea0d118f90072478b471816d40c9fdf51f70714)

Issue #99, opened 2022-06-08
Copy button not sticky, and overlapping


dd32:
The Copy button is not sticky, but it's also too-sticky.

Scrolling:
![ezgif com-gif-maker (2)](https://user-images.githubusercontent.com/767313/172580958-77b99154-18b0-440e-9a23-0608e28a4941.gif)

Scrolling and hiding:
![ezgif com-gif-maker (3)](https://user-images.githubusercontent.com/767313/172580978-4d47f715-77c7-474e-af9e-3087fb7b91dc.gif)

PR #100, opened 2022-06-08
Code Syntax Block: Iterate on expand/collapse button

### LINKED ISSUE #22: Code blocks should not be collapsed.

ryelle:
This PR refactors the collapse/expand functionality. It moves the button up into a group with the "copy" button, so it appears before the code, and clicking it doesn't move your focus out off the bottom of the screen. This also switches the two links to a single real `button`, for better keyboard behavior.

This also fixes the uses/used by toggles (https://github.com/WordPress/wporg-developer/pull/37#issuecomment-1149576182), something in trunk broke the JS connection there.

| Production | Trunk | PR |
|------------|-------|----|
| production | before | after<br /><br />Hover<br /> after-hover |

Any code example more than 10 lines has the expand button now:

Screen Shot 2022-06-08 at 3 36 49 PM

Screencast of it in use:

https://user-images.githubusercontent.com/541093/172700358-f9421168-50de-48ca-ba19-82d4bf9b9464.mp4

### DIFF
--- a/source/wp-content/themes/wporg-developer/inc/user-content.php
+++ b/source/wp-content/themes/wporg-developer/inc/user-content.php
@@ -176,8 +176,8 @@ class DevHub_User_Submitted_Content {
 				array(
 					'copy'   => __( 'Copy', 'wporg' ),
 					'copied' => __( 'Code copied', 'wporg' ),
-					'expand'   => __( 'Expand full source code', 'wporg' ),
-					'collapse' => __( 'Collapse source code', 'wporg' ),
+					'expand'   => __( 'Expand code', 'wporg' ),
+					'collapse' => __( 'Collapse code', 'wporg' ),
 				)
 			);
 
--- a/source/wp-content/themes/wporg-developer/inc/user-content.php
+++ b/source/wp-content/themes/wporg-developer/inc/user-content.php
@@ -176,6 +176,8 @@ class DevHub_User_Submitted_Content {
 				array(
 					'copy'   => __( 'Copy', 'wporg' ),
 					'copied' => __( 'Code copied', 'wporg' ),
+					'expand'   => __( 'Expand full source code', 'wporg' ),
+					'collapse' => __( 'Collapse source code', 'wporg' ),
 				)
 			);
 
--- a/source/wp-content/themes/wporg-developer/js/function-reference.js
+++ b/source/wp-content/themes/wporg-developer/js/function-reference.js
@@ -13,13 +13,13 @@ jQuery( function ( $ ) {
 
 	function collapseCodeBlock( $element, $button ) {
 		$button.text( wporgFunctionReferenceI18n.expand );
-		$element.data( 'is-expanded', false );
+		$button.attr( 'aria-expanded', 'false' );
 		$element.css( { height: MIN_HEIGHT + 'px' } );
 	}
 
 	function expandCodeBlock( $element, $button ) {
 		$button.text( wporgFunctionReferenceI18n.collapse );
-		$element.data( 'is-expanded', true );
+		$button.attr( 'aria-expanded', 'true' );
 		$element.css( { height: $element.data( 'height' ) + 'px' } );
 	}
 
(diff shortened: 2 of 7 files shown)


tellyworth:
I like this. The only issue I have is the way the copy and collapse buttons overlap the code if the first line or two is long. That's fine when they appear only on hover, but they stay visible when focus is in the code block after clicking.

https://user-images.githubusercontent.com/7200686/172737235-9f9b248a-1e61-49e9-bd2a-d5253451d9bd.mov

I can see why you've done it that way, otherwise the buttons are hard to discover. It's probably fine like this for now, since the workaround is to just click outside the code box if you need to see that full first line. If a better idea turns up later that could be done as a separate PR.

tellyworth:
Out of curiosity I tried semi-transparent buttons and that's definitely a bad idea :)

outdoor2kode:
> The only issue I have is the way the copy and collapse buttons overlap the code if the first line or two is long. 

I try some fixes in this [PR](https://github.com/WordPress/wporg-developer/pull/106) for this issue, it's also mentioned in #99.

> That's fine when they appear only on hover, but they stay visible when focus is in the code block after clicking.

I guess it's done on purpose for a11y (users using screen reader can then see the buttons when focusing on the code block) as I see the `display: revert` is put inside `&:focus-within`.

source/wp-content/themes/wporg-developer/inc/user-content.php
```diff
@@ -176,6 +176,8 @@ public static function scripts_and_styles() {
 				array(
 					'copy'   => __( 'Copy', 'wporg' ),
 					'copied' => __( 'Code copied', 'wporg' ),
+					'expand'   => __( 'Expand full source code', 'wporg' ),
+					'collapse' => __( 'Collapse source code', 'wporg' ),
```
outdoor2kode: Can we make the text just `Expand/Collapse` or `Expand All /Collapse All`, the current button seems big to me. Or there're other reasons we use these texts?

source/wp-content/themes/wporg-developer/js/function-reference.js
```diff
@@ -5,83 +5,81 @@
...
 jQuery( function ( $ ) {
-	let $usesList, $usedByList, $showMoreUses, $hideMoreUses, $showMoreUsedBy, $hideMoreUsedBy;
-	let $sourceCollapsedHeight;
-
-	function onLoad() {
-		sourceCodeHighlightInit();
-
-		toggleUsageListInit();
+	// 22.5px (line height) * 10 for 10 lines + 15px top padding + 10px extra.
+	// The extra 10px added to partially show next line so it's clear there is more.
```
outdoor2kode, line 11: > to partially show next line so it's clear there is more.

It's great! 👍

source/wp-content/themes/wporg-developer/js/function-reference.js
```diff
@@ -14,13 +14,16 @@ jQuery( function ( $ ) {
 	function collapseCodeBlock( $element, $button ) {
 		$button.text( wporgFunctionReferenceI18n.expand );
 		$button.attr( 'aria-expanded', 'false' );
+		// This uses `css()` instead of `height()` to prevent jQuery from adding
+		// in the padding. We want to add in just the top padding, since the
+		// bottom is intentionally cut off.
```
outdoor2kode, line 19: 💯

outdoor2kode (approved):
LGTM! Just a small question, but it's not a blocker.

source/wp-content/themes/wporg-developer/inc/user-content.php
ryelle: Sure, I just copied the existing text. I'll make that change before merging this.

ryelle:
> > That's fine when they appear only on hover, but they stay visible when focus is in the code block after clicking.
>
> I guess it's done on purpose for a11y (users using screen reader can then see the buttons when focusing on the code block)

Yep, that's exactly it. When you click the button, you've set the focus there, so the button stays visible. Same as if you were navigating through with a keyboard.

PR #102, opened 2022-06-09
Fix class/method embeds

### LINKED ISSUE #35: Embed handler is broken on class pages

iandunn:
Fixes https://github.com/WordPress/wporg-developer/issues/35

1. checkout branch
2. visit `wp-admin/options-permalink.php` or `wp rewrite flush`
3. open `/reference/classes/wp_query/get/embed/`
3. open `/reference/classes/wp_query/embed/`

### DIFF
--- a/source/wp-content/themes/wporg-developer/inc/registrations.php
+++ b/source/wp-content/themes/wporg-developer/inc/registrations.php
@@ -74,10 +74,11 @@ class DevHub_Registrations {
 			'show_in_rest' => true,
 		) );
 
-		// Methods
+		// Rewrite rules. The more specific rules like `/embed` must be first, to override the more generic rules.
 		add_rewrite_rule( 'reference/classes/page/([0-9]{1,})/?$', 'index.php?post_type=wp-parser-class&paged=$matches[1]', 'top' );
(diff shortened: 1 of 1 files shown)


tellyworth:
This seems to work at a quick glance, but I can't remember exactly what the edge cases were that we talked about earlier.

Does it help to swap the order of the regexes? If the first one catches the `/embed/` case then it might not be necessary to use negative lookaheads at all.

iandunn:
Ah, yeah, you're right! f76f5f9bde630777dc71eafbe3df996fa4a5baf6

TIL `add_rewrite_rule( $regex, $query, 'top' )` doesn't add it to the _top_ of the rule stack, it adds it to the _bottom_ of the _custom_ rule stack (which is on _top_ of the Core rule stack) 😩 

`SimplePie_Enclosure::embed()` is the only method name I found named `embed()`, but it doesn't get a docs page anyway because it's 3rd party.

tellyworth:
Note that it's necessary to run this to test:

```bash
wp-env run cli "wp rewrite flush"
```

Issue #67, opened 2022-06-02
Source code with mixed HTML/PHP aren't highlighted correctly


ryelle:
See https://github.com/WordPress/wporg-developer/pull/44#issuecomment-1143947766

When a function mixes HTML & PHP, the PHP is not syntax-highlighted until after the `<?php`.

![](https://user-images.githubusercontent.com/541093/171743138-5e09d0b8-3b86-4acb-a802-4067c0c0423c.png)

Seems to be an issue with `prismjs` itself — [try it out here](https://prismjs.com/test.html#language=php&text=%0Afunction%20blah()%20%7B%0A%3F%3E%3Cb%3EThis%20is%20a%20test%3C%2Fb%3E%3C%3Fphp%0A%20%20%20%20return%20'foo'%3B%0A%7D%0A%2F%2F%20Comment).

ryelle:
I tried the [Syntax-highlighting Code Block](https://wordpress.org/plugins/syntax-highlighting-code-block/), and while it does highlight correctly, the line numbers can't be customized and they're read aloud by a screen reader. We'd also need to write a new CSS theme for the syntax colors.

<img width="455" alt="Screen Shot 2022-06-02 at 6 24 18 PM" src="https://user-images.githubusercontent.com/541093/171748344-88d5c251-f8ba-4d4a-bb96-270167e2b9d6.png">

Those issues could probably be fixed with PRs into the plugin & some CSS on our side, but ideally there'd be a way to keep using the current (prismjs) block.

PR #105, opened 2022-06-09
Fix/67 partial highlighting

dd32:
See #67 

This takes a hackyish approach and prepends `<?` to all function/method/source previews, which forces it into PHP mode.

Unfortunately I don't see this as being a use-case that would be fixed upstream, as most uses would be highlighting `<?php` within a html highlighted block (which is what is happening here) rather than assuming there's a hidden `<?php` tag.

The major downside of this approach is that the `<?` is displayed prior to the syntax highlighting on load, causing the first line to jump slightly after a few milliseconds as the HTML elements are added and then it's hidden.

It might be possible to do this same thing, but dynamically prepend `<?` in JS when it's parsing.. 

| Before | After |
| --- | --- |
| Screen Shot 2022-06-09 at 2 06 33 pm | Screen Shot 2022-06-09 at 2 05 53 pm |
| non-affected code blocks still work |
| Screen Shot 2022-06-09 at 2 06 54 pm | Screen Shot 2022-06-09 at 2 06 54 pm  |

### DIFF
--- a/source/wp-content/themes/wporg-developer/js/function-reference.js
+++ b/source/wp-content/themes/wporg-developer/js/function-reference.js
@@ -143,6 +143,12 @@ jQuery( function ( $ ) {
 				code = code.slice( 2 );
 			}
 
+			// For PHP scripts, trim off the initial `<?`, if exists.
+			// See https://github.com/WordPress/wporg-developer/issues/67
+			if ( 'php' === $code.attr( 'lang' ) && code.startsWith( '<?' ) && ! code.startsWith( '<?php' ) ) {
+				code = code.slice( 2 );
+			}
+
 			// This returns a promise which will resolve if the copy suceeded,
 			// and we can set the button text to tell the user it worked.
 			// We don't do anything if it fails.
--- a/source/wp-content/themes/wporg-developer/scss/prism.scss
+++ b/source/wp-content/themes/wporg-developer/scss/prism.scss
@@ -226,6 +226,6 @@ pre.wp-block-code {
 }
 
 /* Workaround for partial highlighting. See https://github.com/WordPress/wporg-developer/issues/67 */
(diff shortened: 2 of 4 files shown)


source/wp-content/themes/wporg-developer/scss/prism.scss
```diff
@@ -224,3 +224,8 @@ pre.wp-block-code {
 		display: revert;
 	}
 }
+
+/* Workaround for partial highlighting. See https://github.com/WordPress/wporg-developer/issues/67 */
+section.source-content code > span.token.language-php:first-child > span.token.delimiter.important:first-child {
+	display: none;
+}
```
outdoor2kode, line 231: No new line.

source/wp-content/themes/wporg-developer/js/function-reference.js
```diff
@@ -143,6 +143,12 @@ jQuery( function ( $ ) {
 				code = code.slice( 2 );
 			}
 
+			// For PHP scripts, trim off the initial `<?`, if exists.
+			// See https://github.com/WordPress/wporg-developer/issues/67
+			if ( 'php' === $code.attr( 'lang' ) && code.startsWith( '<?' ) && ! code.startsWith( '<?php' ) ) {
+				code = code.slice( 2 );
+			}
```
outdoor2kode, line 150: 💯

outdoor2kode (approved):
Done testing and looks good on my side.

This is hacky but smart...😆
I think the first line jumping issue can be accepted for the moment as if the Source section doesn't show on top of the page, it's relatively hard to find it jumping.

But also curious about usually in this case how we find its parsing logic and add code to it. Do we leverage filter and hook to do so? (guessing based on my limited knowledge...)

tellyworth:
I think this is superseded by #111 - can we close this one?

dd32:
Closing in favour of #111.

PR #109, opened 2022-06-09
Fix the search display of functions/hooks

### LINKED ISSUE #108: Search auto-suggest doesn't distinguish duplicate hook/function names
Let's say I want to find the `wp_authenticate` function. Steps:

1. In the search box, type `wp_auth`. 
2. Click on `wp_authenticate`.

Screen Shot 2022-06-09 at 4 07 34 pm

Expected: I should land on http://localhost:8888/reference/functions/wp_authenticate/, or perhaps http://localhost:8888/?s=wp_authenticate

Actual: I'm taken directly to http://localhost:8888/reference/hooks/wp_authenticate/

Possible solutions:
* Use prefixes or other visual indicators to distinguish between functions, hooks, classes in the auto-suggest dropdown
* For searches that match both a function/method and a hook, go to the search listing rather than directly to one of the pages
* ..?

dd32:
Fixes #108 

 - Functions/Methods have () appended.
 - Hooks + Functions with the same base-name are both shown
 - Search tweaks are applied, so that () works in the search box

I'm not sure if there's a better way to specify the type of search it is, which is why I went with the `_autocomplete_search` extra param..

| Before | After |
| --- | --- |
| Screen Shot 2022-06-09 at 4 07 34 pm |  Screen Shot 2022-06-09 at 4 47 28 pm |

### DIFF
--- a/source/wp-content/themes/wporg-developer/inc/autocomplete.php
+++ b/source/wp-content/themes/wporg-developer/inc/autocomplete.php
@@ -114,7 +114,9 @@ class DevHub_Search_Form_Autocomplete {
 				if ( in_array( $post->post_type, $post_types_function_like ) ) {
 					$title .= '()';
 				}
-
+		if ( $post->post_type == 'wp-parser-class' ) {
+			$title =  'class ' . $title . ' {}';
+		}
 				$form_data['posts'][ $title ] = $permalink;
 			}
 		}
--- a/source/wp-content/themes/wporg-developer/inc/autocomplete.php
+++ b/source/wp-content/themes/wporg-developer/inc/autocomplete.php
@@ -104,11 +104,18 @@ class DevHub_Search_Form_Autocomplete {
 		$search = get_posts( $args );
 
 		if ( ! empty( $search ) ) {
-			$titles             = wp_list_pluck( $search, 'post_title', 'ID' );
-			$titles             = array_unique( $titles );
-			$titles             = array_flip( $titles );
-			$titles             = array_map( 'get_permalink', $titles );
-			$form_data['posts'] = $titles;
+			$post_types_function_like = array( 'wp-parser-function', 'wp-parser-method' );
+
+			foreach ( $search as $post ) {
+				$permalink = get_permalink( $post->ID );
+				$title     = $post->post_title;
+
+				if ( in_array( $post->post_type, $post_types_function_like ) ) {
+					$title .= '()';
+				}
+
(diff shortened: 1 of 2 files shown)


source/wp-content/themes/wporg-developer/inc/autocomplete.php
```diff
@@ -99,16 +99,24 @@ public function autocomplete_data_update() {
...
-			$form_data['posts'] = $titles;
+			$post_types_function_like = array( 'wp-parser-function', 'wp-parser-method' );
+
+			foreach ( $search as $post ) {
+				$permalink = get_permalink( $post->ID );
+				$title     = $post->post_title;
+
+				if ( in_array( $post->post_type, $post_types_function_like ) ) {
+					$title .= '()';
+				}
```
dd32, line 116: I feel like this should be done automatically through a filter somewhere.. I couldn't locate one at the time though.

source/wp-content/themes/wporg-developer/inc/autocomplete.php
```diff
@@ -99,16 +99,24 @@ public function autocomplete_data_update() {
...
+
```
StevenDufresne: What about following the class signature for classes?

```suggestion
		if ( $post->post_type == 'wp-parser-class' ) {
			$title =  'class ' . $title . ' {}';
		}
```
### Example

source/wp-content/themes/wporg-developer/inc/search.php
```diff
@@ -85,11 +94,11 @@ public static function pre_get_posts( $query ) {
 		} else {
 			// If user has '()' at end of a search string, assume they want a specific function/method.
 			$s = htmlentities( $query->get( 's' ) );
-			if ( '()' === substr( $s, -2 ) ) {
+			if ( '()' === substr( $s, -2 ) || '(' == substr( $s, -1 ) ) {
```
dd32, line 97: This was to ensure that a search like `wp_authenticat(` would return results, rather than that single char having no results.

`parse_query` is a good test phrase too.

source/wp-content/themes/wporg-developer/inc/autocomplete.php
dd32: Well, ~the leading `class` kind of tells you what it is..~  but including the `class` and `{}` makes sense to me!

See the other comment I just added.. I feel like these suffixes should be through a filter or function or something..

edit: Didn't notice that the diff adds the `class` prefix too. yeah. This change is needed! Thanks!

PR #110, opened 2022-06-09
Experiment: Change site-title to  "Developer Resources".

StevenDufresne:
Source: https://meta.trac.wordpress.org/ticket/5585#comment:1

This PR makes the site title "Developers Resources" for all the subfolders except for `resources` and `resource` since they do not have breadcrumbs.


outdoor2kode (approved):
💯 Tested and looks good to me.

I like the idea to change and link all the site titles to the DevHub homepage!
Except that I'm wondering does it make sense to add titles to subfolders just like how Block Editor Handbook and REST API Handbook do? IMO it's clearer than breadcrumbs to show where we are currently at. 
image
image

PR #111, opened 2022-06-09
Code Blocks: Inject opening PHP tag via Prism hooks before parsing the syntax.

### LINKED ISSUE #67: Source code with mixed HTML/PHP aren't highlighted correctly

ryelle:
Alternative to #105, this uses [Prism hooks](https://prismjs.com/extending.html#writing-plugins) to add and remove the initial `<? ` on PHP code (where one doesn't already exist). 

Uses `before-highlight` to inject a `<? ` before running the highlight function, and `before-insert` to remove it from the `highlightedCode` string before injecting it into the DOM.

Fixes #67 

Screen Shot 2022-06-09 at 5 42 04 PM

### DIFF
--- a/source/wp-content/themes/wporg-developer/js/function-reference.js
+++ b/source/wp-content/themes/wporg-developer/js/function-reference.js
@@ -130,7 +130,9 @@ jQuery( function ( $ ) {
 	// Runs before the highlight parsing is run.
 	// `env` is defined here: https://github.com/PrismJS/prism/blob/2815f699970eb8387d741e3ac886845ce5439afb/prism.js#L583-L588
 	Prism.hooks.add( 'before-highlight', function ( env ) {
-		if ( 'php' === env.language && ! env.code.startsWith( '<?' ) ) {
+		// If the code starts with `<`, it's either already got an opening tag,
+		// or it starts with HTML. Either way, we don't want to inject here.
+		if ( 'php' === env.language && ! env.code.startsWith( '<' ) ) {
 			env.code = '<? ' + env.code;
 			env.hasAddedTag = true;
 		}
--- a/source/wp-content/themes/wporg-developer/js/function-reference.js
+++ b/source/wp-content/themes/wporg-developer/js/function-reference.js
@@ -1,4 +1,4 @@
-/* global jQuery, wporgFunctionReferenceI18n */
+/* global jQuery, Prism, wporgFunctionReferenceI18n */
 /**
  * function-reference.js
  *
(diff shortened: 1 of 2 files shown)


tellyworth:
I know @dd32 was working on a similar idea. This looks nice and simple.

source/wp-content/themes/wporg-developer/js/function-reference.js
```diff
@@ -126,4 +126,23 @@ jQuery( function ( $ ) {
...
+			env.code = '<? ' + env.code;
+			env.hasAddedTag = true;
+		}
+	} );
+
+	// Runs before `highlightedCode` is set to the `innerHTML` of the container.
+	Prism.hooks.add( 'before-insert', function ( env ) {
+		if ( env.hasAddedTag ) {
+			env.highlightedCode = env.highlightedCode.replace(
+				'<span class="token delimiter important">&lt;?</span> ',
```
dd32, line 145: I think this needs to be anchored to only replace the initial instance of it perhaps? Just to avoid cases where `<?` exists within any code blocks.

dd32 (approved):
This looks good to me!

I had initially looked at Prism hooks, but wasn't able to figure it out at first glance, so this looks much better to me.

I wonder if it would be best to check the code also doesn't start with `<`? ie. to not invalidly highlight `<html>...<?php`?

source/wp-content/themes/wporg-developer/js/function-reference.js
dd32: Turns out this is already a singular replace, as JS `.replace()` uses a regexp, and you have the explicitly state you want a "global" replace.. TIL.

ryelle:
> I wonder if it would be best to check the code also doesn't start with `<`?

It took me a minute to realize you meant `<` not `<?` 😆  But that's a good point — I added a test case (a user comment with the code `<b><?php esc_html_e( 'Hello World' ); ?></b>`), and sure enough this broke that highlighting. So now the code checks for an initial `<`, which should catch both HTML and opening PHP tags.

PR #113, opened 2022-06-10
Add copyable example code at the top of reference pages

### LINKED ISSUE #23: The first code block should be a copyable example.

tellyworth:
Fixes #23.

This automatically adds some short example code that's easily copyable to the top of the page, right under the summary.

So far it only gives sample code for actions and filters. It probably needs a bit of styling and thought as to exactly where to put it.

Some examples:
Screen Shot 2022-06-10 at 2 37 48 pm
Screen Shot 2022-06-10 at 2 37 41 pm

Some that could use a bit of tweaking:
Screen Shot 2022-06-10 at 2 37 25 pm
Screen Shot 2022-06-10 at 2 37 33 pm

### DIFF
--- a/source/wp-content/themes/wporg-developer/content-reference.php
+++ b/source/wp-content/themes/wporg-developer/content-reference.php
@@ -10,6 +10,7 @@
 
 	<section class="summary">
 		<?php echo get_summary(); ?>
+		<?php echo get_short_example(); ?>
 	</section>
 
 <?php if ( is_single() ) : ?>
--- a/source/wp-content/themes/wporg-developer/inc/template-tags.php
+++ b/source/wp-content/themes/wporg-developer/inc/template-tags.php
@@ -1675,6 +1675,70 @@ namespace DevHub {
 		return $summary;
 	}
 
+	function get_short_example( $post_id = null ) {
+
+		if ( empty( $post_id ) ) {
+			$post_id = get_the_ID();
+		}
+
+		// Decorate and return hook arguments.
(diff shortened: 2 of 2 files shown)


StevenDufresne:
As a WP novice, when looking at hooks, I don't understand how the example relates to the signature which traces back to [your comment](https://github.com/WordPress/wporg-developer/pull/59#issuecomment-1145535710) about `apply_filters` being not what users would use. I think the design and layout work well, and I can see it positively impacting functions. I'm torn about the data types since I'll need to manually remove them in my code but it's like a portable reference.


Found a bug btw:

StevenDufresne:
@tellyworth I'm going to close this as it is a couple design versions behind. We can reopen if we decide to go this route.

Thanks for the PR!

PR #117, opened 2022-06-10
Remove line numbers from short snippets

### LINKED ISSUE #101: Handbooks: Hide line numbers on single-line code blocks
I noticed there are some one- and two-liner code blocks in some handbooks that look a bit silly with line numbers. For example:

http://localhost:8888/coding-standards/wordpress-coding-standards/html/
Screen Shot 2022-06-07 at 5 22 25 pm

_Originally posted by @tellyworth in https://github.com/WordPress/wporg-developer/pull/85#pullrequestreview-997636067_

iandunn:
Fixes #101 

This isn't applied to `reference/template-source.php`, because the line number is important and absolute in that context.

Good pages to test:
* `/coding-standards/wordpress-coding-standards/html/#self-closing-elements`
* `/coding-standards/wordpress-coding-standards/css/#media-queries`
* `/coding-standards/inline-documentation-standards/php/#1-1-parameters-that-are-arrays`
*  `/coding-standards/inline-documentation-standards/javascript/`
* `/reference/hooks/oembed_providers/` (no change)
* `/reference/functions/_register_remote_theme_patterns/` (no change)

### DIFF
--- a/source/wp-content/themes/wporg-developer/inc/formatting.php
+++ b/source/wp-content/themes/wporg-developer/inc/formatting.php
@@ -783,6 +783,7 @@ class DevHub_Formatting {
 		// This should account for all languages used in the handbooks.
 		$lang_list = [ 'js', 'json', 'sh', 'bash', 'html', 'css', 'scss', 'php', 'markdown', 'yaml' ];
 		$lang = in_array( $attr['lang'], $lang_list ) ? $attr['lang'] : 'php';
+		$show_line_numbers = substr_count( $content, '<br />' ) > 5; // The content usually has 1-2 extra breaks.
 
 		// Shell is flagged with `sh` or `bash` in the handbooks, but Prism uses `shell`.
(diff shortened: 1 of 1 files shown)


ryelle:
I rebased this after https://github.com/WordPress/wporg-developer/commit/a40a7ce1c9b3f425ad2cf9fdf54ef3d3e74b0b8c, and updated it so we trim the content first & count `\n`, this seems to be a little more accurate than `<br />`. In my testing that seemed to always return 0, disabling line numbers regardless of length.

Now it keeps the line numbers on longer content, but removes them from shorter content:

Screen Shot 2022-06-13 at 12 25 16 PM
 
`/block-editor/getting-started/create-block/wp-plugin/` is a good one to test on, with varying lengths.

Feel free to merge this - or drop my commits, if you'd rather a different approach.

iandunn:
> In my testing that seemed to always return 0, disabling line numbers regardless of length.

Huh, it worked fine in my tests. This version still works in my tests too though, so 👍🏻

PR #122, opened 2022-06-16
Stop floating Table of Contents within content

adamwoodnz:
See #76 

Allows wider content including source.

Table of Contents is fixed to the right of the content on wide screens. On medium and small screens it's inline below the title and summary.

Code reference wide
Screen Shot 2022-06-16 at 7 11 46 PM

Code reference narrow
Screen Shot 2022-06-16 at 7 15 39 PM

Rest API wide
Screen Shot 2022-06-16 at 7 16 59 PM

Rest API narrow
Screen Shot 2022-06-16 at 7 17 37 PM

### DIFF
--- a/source/wp-content/themes/wporg-developer/content-reference.php
+++ b/source/wp-content/themes/wporg-developer/content-reference.php
@@ -23,7 +23,7 @@
 			'header_text' => __( 'Contents', 'wporg' )
 		) );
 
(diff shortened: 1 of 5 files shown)


adamwoodnz:
Reopening to seek feedback from community after removal wasn't well supported

PR #123, opened 2022-06-16
Remove Table of Contents from Code Reference

adamwoodnz:
See #76 

This PR removes the Table of Contents from the Code Reference pages of the dev docs, but leaves the jump to links on content headings.

This allows more space in the upper section of the page for source and other important info.

Before:
Screen Shot 2022-06-17 at 12 03 57 PM

After:
Screen Shot 2022-06-17 at 12 06 00 PM

Also included is a change to [improve display of the ToC on small screens as suggested](https://github.com/WordPress/wporg-developer/pull/93#issuecomment-1152525768) by @iandunn :
- Stop hiding it on really small screens (< 480px wide) as this is where it is very useful for jumping down a long page
- Make it full width on small screens (< 43em wide) so that it better handles really long ToC titles, eg. https://developer.wordpress.org/rest-api/frequently-asked-questions/

Before:
Screen Shot 2022-06-21 at 4 30 52 PM

After:
Screen Shot 2022-06-21 at 4 31 33 PM

### DIFF
--- a/source/wp-content/plugins/handbook/inc/table-of-contents.php
+++ b/source/wp-content/plugins/handbook/inc/table-of-contents.php
@@ -130,7 +130,7 @@ class WPorg_Handbook_TOC {
 	 * @access public
 	 *
 	 * @param string $content Content.
-	 * @return string toc => Modified Content.
+	 * @return string Modified content.
 	 */
 	public function add_toc( $content ) {
 		$parts = $this->parse_content( $content );
--- a/source/wp-content/plugins/handbook/inc/table-of-contents.php
+++ b/source/wp-content/plugins/handbook/inc/table-of-contents.php
@@ -130,7 +130,7 @@ class WPorg_Handbook_TOC {
 	 * @access public
 	 *
 	 * @param string $content Content.
-	 * @return string Modified content.
+	 * @return array toc => Table of Contents, content => Modified Content.
 	 */
 	public function add_toc( $content ) {
 		if ( ! in_the_loop() ) {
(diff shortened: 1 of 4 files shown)


tellyworth:
This looks good to me, I'd like to hear @ryelle's opinion on the a11y impact though.

ryelle:
A11y impact is a good question — as far as I know, screen readers have a heading navigation mode, so it wouldn't take anything away there, as the heading structure is still the same. Removing it might even be an improvement for SR users, so it's not just another thing to skip over.

<details>
<summary>Heading structure</summary>

Screen Shot 2022-06-20 at 6 01 05 PM

</details>

On the other hand, having a content outline can help with cognitive accessibility. I think this is more an issue on the handbooks, where the content on each page is different and having that outline view can help orient readers. The code reference structure is consistent & predictable, so it's less useful there.

Overall, I think this needs community input before making that decision for handbooks, but I'm 👍🏻  for removing it from the code reference.

adamwoodnz:
Thanks @tellyworth and @ryelle,

In terms of getting input on the handbook removal, should we:
a. merge this and see what feedback we get about it being gone from all the handbooks
or 
b. revert removing it from places other than the code reference and instead ask the question somewhere

?

tellyworth:
That's a great insight @ryelle about the code reference structure being very predictable.

@adamwoodnz I think in this repo we should remove it from the code reference only and leave it intact in the handbooks. That way we can show people our best representation of how we think things should work.

adamwoodnz:
> That's a great insight @ryelle about the code reference structure being very predictable.
> 
> @adamwoodnz I think in this repo we should remove it from the code reference only and leave it intact in the handbooks. That way we can show people our best representation of how we think things should work.

Done!

PR #124, opened 2022-06-17
Highlight php code examples in function descriptions

### LINKED ISSUE #48: Code examples parsed from description aren't syntax highlighted
This applies both on the old site and also in the new highlighter branch #44.

See https://developer.wordpress.org/reference/functions/apply_filters/
Screen Shot 2022-06-01 at 3 16 50 pm

The `Example usage` block isn't highlighted. I'm not sure if that's a data problem in the content, or if highlighting is sometimes not applied to code in the description. Note that this text comes from the parsed description, not an Explanation CPT.

tellyworth:
This is obviously very simplistic and makes some assumptions.

Fixes #48.


ryelle:
I looked at the `apply_filters` page for a while trying to figure out why it wasn't highlighted correctly, before realizing the multi-line comment isn't closed. It's not closed in [the source either](https://core.trac.wordpress.org/browser/tags/6.0/src/wp-includes/plugin.php#L135) (probably so it doesn't close the doc block), but it makes the code look wrong:

Screen Shot 2022-06-23 at 1 16 55 PM
 
I don't know how common this is across the descriptions, but maybe there's a way to check & fix it on the fly?

Do you think there's a way to use `DevHub_Formatting::do_shortcode_code`, so that these could get line numbers only if needed?

tellyworth:
I switched it to use the shortcode. Even simpler :)

That's interesting about the dangling comment, I'd never noticed that before. I did some tinkering with regexes and couldn't detect it reliably, though it probably can be done. I'd need to think that one over some more. I think that probably needs a separate issue and PR to handle it - `DevHub_Formatting::do_shortcode_code` would be a good place for that to go.

ryelle (approved):
Looks good 👍🏻 

Agreed that the comment thing can be fixed in another PR.

PR #126, opened 2022-07-08
Revert removing Table of Contents from Code Reference and fix outside content instead

adamwoodnz:
See #76 

We initially tried removing the Table of Contents entirely on Code Reference pages, however [this proved controversial](https://make.wordpress.org/meta/2022/07/01/exploration-improving-devhub/#comment-9329), so I want to put this solution out for comment.

This PR makes the ToC fixed to the right of the content on wide screens for all handbooks. On medium and small screens it's displayed inline below the title.

Code Reference wide
Screen Shot 2022-07-14 at 3 43 13 PM

Code reference wide scrolled
Screen Shot 2022-07-14 at 3 44 14 PM

Code Reference iPad
Screen Shot 2022-07-14 at 3 44 58 PM

Rest API wide
Screen Shot 2022-07-14 at 3 46 05 PM

Rest API iPad
Screen Shot 2022-07-14 at 3 45 47 PM

### DIFF
--- a/source/wp-content/themes/wporg-developer/scss/settings/_dimensions.scss
+++ b/source/wp-content/themes/wporg-developer/scss/settings/_dimensions.scss
@@ -1,3 +1,6 @@
 $single-handbook-content-width: 960px;
 $single-handbook-content-primary-width: 0.72;
+$single-handbook-toc-position-top: 200px;
+
 $devhub-wrap-content-width: 60em;
+$devhub-wrap-toc-position-top: 210px;
--- a/source/wp-content/themes/wporg-developer/scss/settings/_dimensions.scss
+++ b/source/wp-content/themes/wporg-developer/scss/settings/_dimensions.scss
@@ -1,0 +1,3 @@
+$single-handbook-content-width: 960px;
+$single-handbook-content-primary-width: 0.72;
+$devhub-wrap-content-width: 60em;
--- a/source/wp-content/plugins/handbook/inc/table-of-contents.php
+++ b/source/wp-content/plugins/handbook/inc/table-of-contents.php
@@ -133,28 +133,15 @@ class WPorg_Handbook_TOC {
 	 * @return string Modified content.
 	 */
 	public function add_toc( $content ) {
-		$parts = $this->parse_content( $content );
-   		return $parts['toc'] . $parts['content'];
-	}
-
-	/**
-	 * Parses given content and returns modified content and the ToC.
-	 *
-	 * @access public
-	 *
-	 * @param string $content Content.
-	 * @return array toc => Table of Contents, content => Modified Content.
-	 */
-	public function parse_content( $content ) {
 		if ( ! in_the_loop() ) {
-			return array('content' => $content, 'toc' => null);
+			return $content;
 		}
 
 		$toc   = '';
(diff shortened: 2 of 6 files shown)


tellyworth:
The breakpoint for the narrow ipad-style view seems to be with the main content area at about 900px, around 1500px for the window. That seems pretty wide for the ToC to jump over there, is that intended?

Screen Shot 2022-07-27 at 2 57 57 pm

adamwoodnz:
> The breakpoint for the narrow ipad-style view seems to be with the main content area at about 900px, around 1500px for the window. That seems pretty wide for the ToC to jump over there, is that intended?
> 
> Screen Shot 2022-07-27 at 2 57 57 pm

It is intended yes; that's the width that the ToC no longer fits alongside the content. We could look at an in-between layout where both the content and ToC widths are reduced so they can sit side by side down to around 1280px maybe.

Based on these [stats](https://gs.statcounter.com/screen-resolution-stats/desktop/worldwide) around 38% of people use a desktop screen > 1500px wide. Dropping the breakpoint to 1440 would net another 6% or 1366 would have a really significant impact:

Screen Shot 2022-08-29 at 11 40 52 AM

StevenDufresne:
> It is intended yes; that's the width that the ToC no longer fits alongside the content. We could look at an in-between layout where both the content and ToC widths are reduced so they can sit side by side down to around 1280px maybe.

👍 I think you could even do it earlier, mimicking existing behavior at the "desktop" breakpoint which I think is `890px`.

adamwoodnz:
> I think you could even do it earlier, mimicking existing behavior at the "desktop" breakpoint which I think is 890px

Yeah this width ends up working quite well for Code Reference pages. I've updated so we now have:

Below 890px -> full width inline ToC
890px to 1440px -> floating ToC (like existing)
1440px and above -> sticky ToC with responsive width

Chrome - Code Reference
Kapture 2022-08-30 at 10 46 37

Chrome - Coding Standards
Kapture 2022-08-30 at 10 52 30

source/wp-content/themes/wporg-developer/scss/settings/_dimensions.scss
```diff
@@ -0,0 +1,6 @@
+$single-handbook-content-width: 960px;
+$single-handbook-content-primary-width: 0.72;
```
StevenDufresne, line 2: Without looking at the implementation, this property is hard to understand.

StevenDufresne:
~~Viewing `coding-standards/inline-documentation-standards/javascript/`, at around `880px` the browser is clipping the Table of Contents on the right.~~

Screen Shot 2022-08-30 at 3 39 05 PM

**This was related to not having a sidebar configured. Since all the sites have a sidebar, I think we can disregard this for now.**

StevenDufresne:
The alignment is a bit distracting between the search bar, content and, ToC. I think it would be a nice improvement if we can get the content to align with the search bar. I'll take a look.

Screen Shot 2022-08-31 at 10 18 17 AM

StevenDufresne:
I think this is good enough to merge. Since we are looking to release the codebase, I'll merge it so others can test it the project in its entirety. We can open up a PR for any additional issues.

PR #137, opened 2022-09-06
Close description tag later.

### LINKED ISSUE #136: Broken docs page for register post type function
There are a few pages broken:

https://developer.wordpress.org/reference/functions/wp_remote_request/
https://developer.wordpress.org/reference/functions/register_post_type/ 

Source: https://meta.trac.wordpress.org/ticket/6471

StevenDufresne:
Fixes: #136

The `<div class="desc">` container when we close the outer `<li>` because we may have nested list.

#### `reference/functions/wp_remote_request/`
| Before | After |
| --- | --- |
|Screen Shot 2022-09-06 at 1 37 25 PM | Screen Shot 2022-09-06 at 1 36 21 PM |

#### `reference/functions/register_post_type/`
| Before | After |
| --- | --- |
| Screen Shot 2022-09-06 at 1 47 29 PM  | Screen Shot 2022-09-06 at 1 48 20 PM  |

### DIFF
--- a/source/wp-content/themes/wporg-developer/inc/formatting.php
+++ b/source/wp-content/themes/wporg-developer/inc/formatting.php
@@ -599,9 +599,9 @@ class DevHub_Formatting {
 				if ( $name ) {
 					$new_text .= "<code>{$name}</code>";
 				}
-				$new_text .= "<span class='type'>{$type}</span><div class='desc'>{$description}</div>";
+				$new_text .= "<span class='type'>{$type}</span><div class='desc'>{$description}";
 				if ( ! $skip_closing_li ) {
-					$new_text .= '</li>';
+					$new_text .= '</div></li>';
 				}
 				$new_text .= "\n";
 			}


source/wp-content/themes/wporg-developer/inc/formatting.php
```diff
@@ -599,9 +599,9 @@ public static function fix_param_hash_formatting( $text ) {
 				if ( $name ) {
 					$new_text .= "<code>{$name}</code>";
 				}
-				$new_text .= "<span class='type'>{$type}</span><div class='desc'>{$description}</div>";
+				$new_text .= "<span class='type'>{$type}</span><div class='desc'>{$description}";
 				if ( ! $skip_closing_li ) {
-					$new_text .= '</li>';
+					$new_text .= '</div></li>';
```
iandunn, line 604: What happens when `$skip_closing_li` is true? It seems like there'd never be a `</div>`. Do we need to add that in one of the `$in_list` conditions?

I haven't found an example where anything looks broken, though.

iandunn (approved):
Left a question, but not a blocker.

If it looks good to you then feel free to merge, and if there is a problem then we can always iterate.

source/wp-content/themes/wporg-developer/inc/formatting.php
StevenDufresne: Yeah it could be broken if it ends up true, which _could_ be the case with bad data. We could get a nested object that has an opening `{` brace but the closing one never shows up. It will never close. I think it's a risk that has always existed and short of refactoring this whole section, I'm not certain there's an easy win. If `$skip_closing_li` doesn't return to false, I _think_ the `<li>` would never close as well.

iandunn: 👍🏻

PR #140, opened 2022-09-14
Fix/6475 Table of Contents overflowing side of window and global footer

adamwoodnz:
Fixes https://meta.trac.wordpress.org/ticket/6475 

2 issues are fixed by this PR:

1. If a user has scrollbars set to always be displayed the ToC can overflow the screen on the right at certain sizes (~1440px wide)
2. On short screens the ToC can cover the global footer when scrolled right to the bottom

For 2 a script has been added which detects if the ToC is overlapping the footer on `scroll` and `resize`, and if so shifts it up above. `lodash.throttle` has been used to throttle the events, as opposed to `requestAnimationFrame`, as it provides `trailing` event support. This proved necessary in many browsers where there is a delay in restoring the scroll position after refreshing, and rAF throttling blocked these delayed events, resulting in the ToC not being repositioned on refresh when scrolled to the bottom of the page.

As there is no build process for JS in this repo I've opted for es6 local `import`, but we could introduce a build instead. I think browser support is good enough for this, based on our [support policy](https://make.wordpress.org/core/handbook/best-practices/browser-support/) and https://caniuse.com/es6-module-dynamic-import 

#### Horizontal overflow
| Before | After |  
|-|-|
| Screen Shot 2022-10-05 at 11 51 06 AM | Screen Shot 2022-10-05 at 11 51 51 AM |

#### Footer overlap

Before
Screen Shot 2022-10-05 at 11 54 02 AM

After
Kapture 2022-10-05 at 11 55 59

### How to test

1. Set scrollbars to always be displayed (MacOS: System Prefs > General > Show scroll bars: 'Always')
2. Open one of the pages with a ToC (eg. http://localhost:8888/reference/classes/wp_error/get_error_code/) and resize your browser window to 1440px x 705px
3. Ensure the ToC does not overflow the side of the window
4. Scroll to the bottom of the page, both manually and by using the links in the ToC
5. Ensure the ToC does not cover the footer
6. Refresh the page and ensure the position is updated when scroll position is restored.
7. Try combinations of scrolling up and down and resizing the window. The ToC should reposition itself.
NOTE: at smaller sizes the ToC will not be fixed outside the content, and the layout should not be effected by the new changes.
8. Try other pages like Rest API or Coding Standards. 
NOTE: locally the lefthand sidebar may not be displayed on these pages (it's not for me). To mimic the actual behaviour in prod you can comment out the return on this line: https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer/sidebar-handbook.php#L9


source/wp-content/themes/wporg-developer/scss/main.scss
```diff
@@ -2463,35 +2463,20 @@ body.responsive-show {
...
-			left: calc( #{ $single-handbook-content-width } + ( ( 100vw - #{ $single-handbook-content-width } ) / 2 ) );
+			transform: translateX( calc( ( #{ $single-handbook-content-width } * #{ $single-handbook-content-width-primary-proportion } ) - #{ $single-handbook-content-padding-horizontal } ) );
 			top: $single-handbook-toc-position-top;
 			max-height: calc(90vh - #{ $single-handbook-toc-position-top });
 		}
 	}
 }
 
-@media (min-width: 877px) {
-	// Pages which are not code reference and without a sidebar
```
adamwoodnz, line 2481: This was required for local dev, but in production all pages have a sidebar on the left so this code is redundant

StevenDufresne:
In regards to the footer overlap, since the TOC shows up at the right level relative to the document, could we utilize `position: sticky` as opposed to `position: fixed`? That would remove the need for js.

**Example**: 
https://d.pr/i/HZveB8

In the example, I just changed to sticky positioning and relied on a negative `margin-right` to pull it right.

adamwoodnz:
> In regards to the footer overlap, since the TOC shows up at the right level relative to the document, could we utilize `position: sticky` as opposed to `position: fixed`? That would remove the need for js.
> 
> **Example**: https://d.pr/i/HZveB8
> 
> In the example, I just changed to sticky positioning and relied on a negative `margin-right` to pull it right.

Hmm nice idea, I'll try it out thanks

adamwoodnz:
> In regards to the footer overlap, since the TOC shows up at the right level relative to the document, could we utilize `position: sticky` as opposed to `position: fixed`? That would remove the need for js.
> 
> **Example**: https://d.pr/i/HZveB8
> 
> In the example, I just changed to sticky positioning and relied on a negative `margin-right` to pull it right.

`position: sticky` does indeed work nicely for handling the overlap at the bottom. There is a problem however, in that the ToC is also `float: right` within the content by default. So whereas `absolute` and `fixed` remove the element from the page flow, leaving more space for the content (the original reason to position it out wide), `sticky` does not, leaving a blank space:

Kapture 2022-10-10 at 16 33 09

I have committed this so you can try it, but I think I'll have to revert to the JS solution unless you have another suggestion?

source/wp-content/themes/wporg-developer/scss/main.scss
```diff
@@ -2243,7 +2243,6 @@ div[class*="-table-of-contents-container"] {
 		background: linear-gradient(to right, #fafafa 35%, #fff 35%);
 		max-width: 100%;
 		padding: 0;
-		overflow: auto;
```
adamwoodnz, line 2246: A parent with this property stops `position: sticky` working on children. I'm unsure why it's being set and haven't observed any changes as a result of removing it.

adamwoodnz:
👍 Seems to work well @StevenDufresne, please give it a try

StevenDufresne:
Yep, this is working as expected. Good job!

PR #141, opened 2022-09-15
wp-cli json: Switch from the 'master' branch to the 'main' branch 

hz-tyfoon:
setting the branch name properly as per `wp-cli handbook` repo: https://github.com/wp-cli/handbook
`commands-manifest.json` blob: https://github.com/wp-cli/handbook/blob/main/bin/commands-manifest.json

may be related:👇
svn commit: [11951](https://meta.trac.wordpress.org/changeset/11951) ,
meta ticket: [6394](https://meta.trac.wordpress.org/ticket/6394)

### DIFF
--- a/source/wp-content/themes/wporg-developer/inc/cli.php
+++ b/source/wp-content/themes/wporg-developer/inc/cli.php
@@ -2,7 +2,7 @@
 
 class DevHub_CLI {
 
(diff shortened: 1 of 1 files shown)


iandunn:
This doesn't fix the edit link, but it's still good to have the correct branch set. I'll post some notes on #meta6394 about fixing the edit link.

PR #145, opened 2022-10-19
Fix related section collapsing issue

### LINKED ISSUE #142: Related section is collapsed on page load even though there only 2 uses.
When viewing https://developer.wordpress.org/reference/functions/get_stylesheet/, the related section is collapsed for only 2 uses. 

Expected: It should not be collapsed.

## Screenshot
Screen Shot 2022-09-21 at 9 11 50 AM

outdoor2kode:
Fixes #142 

Originally, the Uses table in the Related section would be collapsed and show 0 records even though there are actually some records existing. After the fix, if the amount of records <= 5, it would never be collapsed.

**Reason for changing from min to max**

The return value of `split_uses_by_frequent_funcs` is `infrequent_count`, and the purpose of the function is to _Rearrange the results of get_uses() so that frequent functions are pushed to the bottom._ I'm not 100% sure why we want to show infrequently used functions to users first here (maybe I misunderstood the real purpose of the function), but according to the function's goal and its [doc comment](https://github.com/WordPress/wporg-developer/blob/f528ed07ab2a6d32945d4e52ffb3f38ed396407f/source/wp-content/themes/wporg-developer/inc/template-tags.php#L1451), I'm thinking that we can change from `min` to `max` here so that we can show as many infrequently used functions to users as possible.

Or if we still want it to show five records at most, we can specify the least records to show as well.
Perhaps:
```php
$uses_to_show = min( $uses_to_show, max( split_uses_by_frequent_funcs( $uses->posts ), $least_uses_to_show) );
```

### DIFF
--- a/source/wp-content/themes/wporg-developer/inc/template-tags.php
+++ b/source/wp-content/themes/wporg-developer/inc/template-tags.php
@@ -1448,7 +1448,7 @@ namespace DevHub {
 	 * Rearrange the results of get_uses() so that frequent functions are pushed to the bottom.
 	 * Sorts the array in-place.
 	 *
-	 * @return int The number of infrequent items in the list (ie the cutoff point for show/hide).
+	 * @return int The number of infrequent items in the list.
 	 */
 	function split_uses_by_frequent_funcs( &$posts ) {
 
--- a/source/wp-content/themes/wporg-developer/reference/template-related.php
+++ b/source/wp-content/themes/wporg-developer/reference/template-related.php
@@ -13,11 +13,12 @@ if ( show_usage_info() ) :
 	$has_uses    = ( post_type_has_uses_info()  && ( $uses    = get_uses()    ) && $uses->have_posts()    );
 	$has_used_by = ( post_type_has_usage_info() && ( $used_by = get_used_by() ) && $used_by->have_posts() );
 
-	$uses_to_show    = 5;
-	$used_by_to_show = 5;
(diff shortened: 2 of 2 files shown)


StevenDufresne:
I can't speak to the particulars of frequency but the main goal was to avoid a very long list of functions.

ryelle:
> I'm not 100% sure why we want to show infrequently used functions to users first here

You can see the original PR #37 — the purpose was to avoid always showing the same functions, ex `__()` or `apply_filters` for all functions. By showing rarely-used functions, the idea is that the functions shown are more relevant to the current page.

I think I prefer adding a `min` value to show (maybe 2?), rather than switching the `min` to `max`, that way we always hide after 5 items. `/reference/functions/_wp_dashboard_recent_comments_row/` is a good test case with many rarely-used functions. On this PR, it shows 11 items before hiding the remaining 7.

outdoor2kode:
> I think I prefer adding a min value to show (maybe 2?), rather than switching the min to max, that way we always hide after 5 items. /reference/functions/_wp_dashboard_recent_comments_row/ is a good test case with many rarely-used functions. On this PR, it shows 11 items before hiding the remaining 7.

I prefer this as well. I've pushed a new commit for it, and in it I've removed the comment `(ie the cutoff point for show/hide).` as it's not used as the cutoff point. Thanks for sharing the test case!

ryelle (approved):
Looks good 👍🏻

PR #148, opened 2022-10-24
Fix code header overlap the bottom border on handbook

### LINKED ISSUE #147: Sticky code block header style a bit off on handbook page
The sticky code block header on the handbook page would overlap the bottom border.
e.g., https://developer.wordpress.org/themes/theme-security/using-nonces/

It should be something like

outdoor2kode:
Fixes #147 

## Root Cause

Parent font-size on handbook pages: `13px`
Parent font-size on reference pabes: `16px`

#### Handbook https://developer.wordpress.org/themes/theme-security/using-nonces/

handbook

#### Reference https://developer.wordpress.org/reference/functions/apply_filters/
reference

### DIFF
--- a/source/wp-content/themes/wporg-developer/scss/prism.scss
+++ b/source/wp-content/themes/wporg-developer/scss/prism.scss
@@ -248,11 +248,18 @@ pre.line-numbers > code {
 	}
 }
 
-/* Make the sticky wp-code-block-button-container not overlap its parent's border */
+/**
+ * Make the sticky wp-code-block-button-container not overlap its parent's border.
+ * A trikcy way is used here (plus and minus margin-bottom),
+ * as we can't just adjust the pre in _global.scss,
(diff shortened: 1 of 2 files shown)


ryelle:
Shouldn't this fix go in [`prism.scss`](https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer/scss/prism.scss)?

outdoor2kode:
Ah yeah, I missed that. Thanks!

ryelle (approved):
This fixes the issue, and can merge as-is.

The `1.63em` value seems a bit like a [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)) though, maybe you can add a comment explaining where that comes from before merging?

outdoor2kode:
The value `1.63em` was set to make the shape of the rounded corners at the bottom of the code block complete.
#### 1.6em
image

#### 1.63em
image

PR #152, opened 2022-11-17
Expand a code block by default if it has the 'expanded' CSS class.

dd32:
Some pages and code blocks should be expanded by default, for readability, this might be because it's 1 line over the line limit, or perhaps because it's the last entry on the page.

Or, it could be a page where the majority of the content is the code block itself:
https://developer.wordpress.org/plugins/settings/custom-settings-page/

This PR simply auto-expands the block upon load if the CSS class is applied.

### DIFF
--- a/source/wp-content/themes/wporg-developer/js/function-reference.js
+++ b/source/wp-content/themes/wporg-developer/js/function-reference.js
@@ -77,7 +77,12 @@ jQuery( function ( $ ) {
 				}
 			} );
 
-			collapseCodeBlock( $element, $expandButton );
+			if ( $element.hasClass( 'expanded' ) ) {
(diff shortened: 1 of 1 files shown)


ryelle (approved):
Looks good to me. Is `expanded` a class that the docs writers are already using?

dd32:
> Is `expanded` a class that the docs writers are already using?

No it's not, it was a shortcut I took instead of attempting to add an option to the Code block (Is that possible without defining a new block?)

I added it to some existing blocks through the Additional CSS classes on some code blocks, such as the above page.

I've followed up with the Docs team in #docs: https://wordpress.slack.com/archives/C02RP4WU5/p1668992574358299

PR #154, opened 2023-01-11
Merge in Personal fork changing theme to block theme.

StevenDufresne:
I am getting this PR up a bit early to allow others to contribute more easily.

This PR does the following:
 - Removes _almost_ all `.php` templates. 
    - ! page-dashicons.php
    - ! embed.php 
 - Removes no longer needed JS
 - Removes no longer needed setup steps
 - Renames theme to `wporg-developer-2023`
 - Adds basic block theme templates
 - Implements the search page
    - I think we can finish iterating on it when this merges. 

**Considerations**
- I created a "trunk" version called [old](https://github.com/WordPress/wporg-developer/tree/old) that we can use for hotfixes while this is in active development. I think it makes sense to just update this repository with the new code and work here. We won't have to update any sandbox or deployment-related code.
- Since this code base is relatively large, I decided to keep all the PHP & CSS and delete it as we go. I'm not certain I understand all the functionality in the theme and think it would be harder to import functions from the current theme into the new block theme. If this were more than a redesign, then I would start with a blank theme, but because it isn't, just keep all the functionality and try to delete non-used interface-related code.
- Lastly, I implement the search page. I hadn't planned to do this all in one branch but I wasn't exactly sure what workflow would work best and building out some components seemed to be a step in the right direction to inform that decision.

## Search Page (`search.html`)

I'll probably need another PR to fix some of the design discrepancies.

| State | Screenshot |
| --- | --- |
| Desktop Results | localhost_8888__s=abs |
| Mobile Results (no body padding)  | localhost_8888__s=abs(iPhone XR) |
| Filters working | localhost_8888__s=block post_type%5B%5D=wp-parser-hook (1) | 
| No matches | localhost_8888__s=blocksadfasdf post_type%5B%5D=wp-parser-hook |
| Autocomplete still working | Screen Shot 2023-01-11 at 3 14 23 PM|


source/wp-content/themes/wporg-developer-2023/inc/extras.php
```diff
@@ -35,47 +35,6 @@ function wporg_developer_body_classes( $classes ) {
 }
 add_filter( 'body_class', 'wporg_developer_body_classes' );
 
-/**
- * Prefixes excerpts for archive view with content type label.
- *
- * @param string  $excerpt The excerpt.
- * @return string
- */
-function wporg_filter_archive_excerpt( $excerpt ) {
```
StevenDufresne, line 44: This was repurposed in the search-filters dynamic block.

dd32 (approved):

StevenDufresne:
Done in: e45ff8b58611ce8163f31779ba1f309a2d49873f

Issue #128, opened 2022-07-14
Current state of the try/block-theme branch


apeatling:
I've been experimenting with this repo and updating the design of developer.wordpress.org to be based on a more compact version of the designs in [this HH Figma](https://www.figma.com/file/iM8I9yAYbwjbc9q9b2WWui/HH-Pages?node-id=1235%3A2987).

At the same time I've also been working on converting the theme to be block based, and support full site editing.

Here's the status of each section:

## Landing Page

The landing page keeps the same general layout, but matches the landing pages from the HH design. It also includes some links that might be good starting points.

<a href="https://user-images.githubusercontent.com/1464705/179047005-62d3e97b-0e1b-463d-8b34-863ef2659bee.png"><img width="300" alt="Screen Shot 2022-07-14 at 10 35 29 AM" src="https://user-images.githubusercontent.com/1464705/179047005-62d3e97b-0e1b-463d-8b34-863ef2659bee.png" style="float: left"></a>

**Some thoughts:**

- Could we eliminate this page entirely, and show the top level sections in the left nav bar on handbook and code reference pages? This would massively help the cohesiveness of the entire dev docs site. Right now all handbooks and the code reference feel disconnected from each other.
- The links could be dynamic. I did build a block for this, but have reverted to links added/edited via the site editor because it wasn't smart enough with the ordering and selection.

## Handbook Pages

The handbook pages continue to use the content coming directly from Github markdown files. The changes largely consist of the improved styling, as well as converting the sidebar and ToC sections into sever side rendering blocks.

<a href="https://user-images.githubusercontent.com/1464705/179047819-fbc9bce8-3bcc-403f-890f-3adec0cbff2e.png"><img width="300" alt="Screen Shot 2022-07-14 at 10 40 07 AM" src="https://user-images.githubusercontent.com/1464705/179047819-fbc9bce8-3bcc-403f-890f-3adec0cbff2e.png"></a>

**Some thoughts:**

- The design works best when each handbook page has a short first sentence that can be highlighted and styled differently. Many pages have this, but many pages do not. Those pages would need to be updated.
- Some pages are not well constructed in terms of the guidelines and the content they contain. I think as part of this design change, the worst offenders would need to be improved.
- The design of the sidebar navigation could use some refinement, I don't think the current blue/black selected states is optimal.
- I think the order of the breadcrumb and title could be reversed, so the breadcrumb is on top. Currently it breaks the flow of title to content.
[...]

PR #156, opened 2023-01-12
Add Single.php blocks

StevenDufresne:
This PR takes some blocks from the [block-theme](https://github.com/WordPress/wporg-developer/issues/128) branch, formats them to match .org and makes the linter _happier_. This also adds the missing blocks for the single view although they are not fully implemented.

## Screenshot
localhost_8888_reference_functions_add_settings_section_ (1)


source/wp-content/themes/wporg-developer-2023/src/code-section-title/block.php
```diff
@@ -0,0 +1,35 @@
+<?php
```
ryelle: What's the purpose of this block that the heading block doesn't handle? I'm not opposed to adding things & then iterating on them, but I'd rather start lighter - removing stuff later often doesn't happen.

source/wp-content/themes/wporg-developer-2023/templates/single.html
```diff
@@ -1,3 +1,5 @@
 <!-- wp:template-part {"slug":"header","className":"has-display-contents"} /-->
-single
+
+<!-- wp:pattern {"slug":"wporg-developer-2023/single-content"} /-->
```
ryelle, line 3: This template is also used for all other "single" pages - ex, handbook pages, class reference, etc. For now this works, but we'll probably want to use a specific template for the code reference pages, and let this be the fallback for handbook-style pages (and that will probably look more like the [documentation single.html](https://github.com/WordPress/wporg-documentation-2022/blob/trunk/source/wp-content/themes/wporg-documentation-2022/templates/single.html).

ryelle: I guess we should also make sure that `single-{post-type}.html` works in block themes.

ryelle (commented):
I've been jumping between other things and this PR because it's very intimidating to review 😣

I would prefer an approach where we add each block in as the function reference is built, so the code can be reviewed as it's added, but since this is already done it's faster to go with it. We'll need to update each with the correct styles at least, so maybe proper review can happen then.

source/wp-content/themes/wporg-developer-2023/src/code-section-title/block.php
StevenDufresne: Since the titles will require an anchor link, I thought it made sense to stub it out.

Screen Shot 2023-01-13 at 8 41 24 AM

ryelle: That's built in to the Table of Contents block, it works on any heading block (I can move that block into mu-plugins so we can share it here).

StevenDufresne:
> I've been jumping between other things and this PR because it's very intimidating to review 😣
> 
> I would prefer an approach where we add each block in as the function reference is built, so the code can be reviewed as it's added.

Yeah, that's fair, other than stubbing the blocks out themselves, most of the code came from the block-theme branch, which looks to me like it's just a port of other existing code. I also don't think there's any urgency to merge.

source/wp-content/themes/wporg-developer-2023/templates/single.html
StevenDufresne: I played with this a bit for handbooks but with moderate success, I can get the template to load for the specific handbooks but for some, the content is not loaded.

ryelle:
> which looks to me like it's just a port of other existing code

That's true, it probably doesn't need an in-depth review - the main thing I noticed was that some of the functions are still using the `wporg_` prefix when they don't need to be since the blocks are namespaced, but that's pretty minor.

> I also don't think there's any urgency to merge.

I think it will be easier to work on the blocks with some stubs already here, so you might as well merge it.

source/wp-content/themes/wporg-developer-2023/src/code-changelog/edit.js
```diff
@@ -0,0 +1,10 @@
+import { useBlockProps } from '@wordpress/block-editor';
+import ServerSideRender from '@wordpress/server-side-render';
+
+export default function Edit( { name, attributes } ) {
+	return (
+		<p { ...useBlockProps() }>
```
ryelle: Technically this adds the block wrappers twice, since the ServerSideRender adds them too. You can just return the `ServerSideRender` component.

source/wp-content/themes/wporg-developer-2023/src/code-section-title/block.php
StevenDufresne: Great. I'll remove the custom block.

StevenDufresne: I removed the custom block and used just an `h2` tag. It does leave us with a lot of HTML to change if the tag ever changes, so I see a path back to a component in the future. We could also remove the headings from the blocks and conceive a different way to conditionally render sections. Nonetheless, I'll merge this and we can update as necessary.

ryelle: I think about it differently - using headings gives us more freedom to change the heading architecture in the future, if we wanted to break up the sections into subsections or something (like, post reskin-redesign).

> We could also remove the headings from the blocks and conceive a different way to conditionally render sections.

That's an interesting idea… since that's the reason you're adding them in the block instead of putting the heading in the pattern? That would be a good future iteration, maybe title + level as an attribute and a helper function all the blocks can use to render it? or some use of InnerBlocks with only a heading block… anyway, that's getting ahead of things 😁

PR #166, opened 2023-01-16
Fix breadcrumbs.

### LINKED ISSUE #163: Component: Breadcrumbs
The breadcrumb component that we are using in other themes does not work here.

Notes:
The theme already has code to generate it.

StevenDufresne:
Closes: #163

This PR creates a component and calls `breadcrumb_trail` to create the site breadcrumbs. I chose to not use `site-breadcrumbs` because of how complex the breadcrumb logic is for this theme. 

Since handbook paths can be up to 5 parts long of varying lengths, I had to come up with a system to truncate. I added some CSS to handle that truncation by hiding items left to right starting after the "home" link as the viewport shrinks. It seems to work well although I'll admit I'm not certain I've seen the longest path to test against yet. I have also only handled up to 5 parts. I think we should reconsider our approach if articles are nested 6 or more times.

### Use [site-breadcrumbs filter](https://github.com/WordPress/wporg-mu-plugins/blob/b90d3ded0391f70a6b8ed0f18fc439b10a4fbf26/mu-plugins/blocks/site-breadcrumbs/index.php#L70).

When the breadcrumb code is created in this theme, HTML is already added to the `$items` list. If we wanted to make use of the `wporg_block_site_breadcrumbs` filter we would have to strip that all out. Not the hardest thing to do, but slight risk.

I could be convinced down this path, in `50/50`.

## Screenshots
| Desktop | Tablet | Mobile |
| --- | --- | --- |
| localhost_8888_block-editor_how-to-guides_internationalization_ | localhost_8888_block-editor_how-to-guides_internationalization_ (1) | localhost_8888_block-editor_how-to-guides_internationalization_ (2)|


source/wp-content/themes/wporg-developer-2023/functions.php
```diff
@@ -216,6 +218,29 @@ function breadcrumb_trail_items_for_hooks( $items, $args ) {
...
+	if ( ! is_singular() && ! is_single() && ! is_post_type_archive() && ! is_archive() ) {
+		return $items;
+	}
+
+	return array_filter(
+		$items,
+		function( $item ) {
+			// Remove the 'reference' parent based on the presence of its URL.
+			// We can't use the label because of internationalization.
+			$result = (bool) preg_match( '/href=".*\/reference\/"/', $item, $matches );
```
StevenDufresne: @dd32 I'm not sure if you are familiar with this code, but any opinion on pulling out the "reference" link from the breadcrumbs like this? 

I looked through the code that adds it, but I'm fearful of making changes during this "site refresh".

StevenDufresne:
I just found one that is mindblowingly long:

[Developer](#) / [Coding Standards Handbook](#) / [Inline Documentation Standards](#) / JavaScript Documentation Standards

@jasmussen, Any idea on what we can do here?

jasmussen:
> [Developer](https://github.com/WordPress/wporg-developer/pull/166#) / [Coding Standards Handbook](https://github.com/WordPress/wporg-developer/pull/166#) / [Inline Documentation Standards](https://github.com/WordPress/wporg-developer/pull/166#) / JavaScript Documentation Standards

Keeping in mind I'm not aware of what's technically feasible here, my instinct would be this:

> [Developer](https://github.com/WordPress/wporg-developer/pull/166#) / ... / [Inline Documentation Standards](https://github.com/WordPress/wporg-developer/pull/166#) / JavaScript Documentation Standards

That is, show the first and last clickable breadcrumbs. 

Alternately we just allow wrapping. It would be good to see this in context before we settle on it, though.

ryelle:
> Use [site-breadcrumbs filter](https://github.com/WordPress/wporg-mu-plugins/blob/b90d3ded0391f70a6b8ed0f18fc439b10a4fbf26/mu-plugins/blocks/site-breadcrumbs/index.php#L70).

This method was what I intended we'd do for all these subsites when I made it, you can see [in Documentation](https://github.com/WordPress/wporg-documentation-2022/blob/13f348f8949c355c861b8e9354cf4a2d65dedb78/source/wp-content/themes/wporg-documentation-2022/functions.php#L40-L96) how I used it there.

Could you refactor whatever's giving you HTML to return the url/title format used by `wporg_block_site_breadcrumbs`?

> Since handbook paths can be up to 5 parts long of varying lengths, I had to come up with a system to truncate.

We've had a similar issue on Documentation (up to 4 parts, but article titles can be [very long](https://wordpress.org/documentation-test/article/difference-between-wordpress-org-and-wordpress-com/). I also wrote some CSS to hide those middle paths in the Docs theme. Whatever the solution for this, it would be best to port back up to the `site-breadcrumbs` block, and IMO, this is the main reason for everyone to use the same block, since we're running into all the same problems.

StevenDufresne:
Alright, I updated to use the `site-breadcrumbs` and reverse out CSS dealing with making it look good on mobile. Let's address that in the my plugin itself: https://github.com/WordPress/wporg-mu-plugins/issues/325.

source/wp-content/themes/wporg-developer-2023/functions.php
dd32: For readability I'd use this regex:
```suggestion
			$result = (bool) preg_match( '!href="[^"]+/reference/"!', $item );
```

I don't see any issues with removing the item this way though.

source/wp-content/themes/wporg-developer-2023/functions.php
```diff
@@ -352,3 +377,25 @@ function rename_comments_meta_box( $post_type, $post ) {
...
+
+	foreach ( get_breadcrumbs()->items as $crumb ) {
+		// Get the link and title from the breadcrumb.
+		// 4 capture groups to handle variation in the markup.
+		// This code assumes that an <a> tag or a string is returned from the breadcrumb.
+		preg_match( '/<a(.*?)href="([^"]+)"(.*?)>(.*?)<\/a>/', $crumb, $matches );
+		$breadcrumbs[] = array(
+			'url' => isset( $matches[2] ) ? $matches[2] : '',
+			'title' => isset( $matches[4] ) ? $matches[4] : $crumb,
+		);
```
dd32:
```suggestion
		preg_match( '!<a[^>]+href="(?P<href>[^"]+)"[^>]+>(?P<title>.+)</a>!', $crumb, $matches );
		$breadcrumbs[] = array(
			'url'   => $marches['href'] ?? '',
			'title' => $matches['title'] ?? $crumb,
		);
```

Using a non-/ delimiter makes it easier, I would normally use `/`, `!`, `#` or `~` depending upon what characters are in the regex.. if `/` then `!`, if using `!` within negative lookahead/behinds then `#` or `~`.

Using named groups helps readability, and since 1 & 3 aren't being used there's no point in capturing these groups either.

Using the PHP 7 Null-Coalesce makes it a bit nicer to read as well.

StevenDufresne: Nice, that's definitely cleaner. I had to tweak the regex a bit:

`!<a[^>]+href="(?P<href>[^"]+)"[^>]+>(?P<title>.+)</a>!`
to:
`!<a[^>]+href="(?P<href>[^"]+)"[^>]*>(?P<title>.+)</a>!`

Issue #162, opened 2023-01-13
Layout: Code


StevenDufresne:
**Blocks:**
- [x] code-reference-deprecated #167
- [x] code-reference-private-access #168 
- [x] code-reference-title  #171 
- [x] code-reference-summary [3732004](https://github.com/WordPress/wporg-developer/commit/37320043390ab206ec59cd7f6587f08433a53377)
- [x] code-reference-description (I don't think there's anything to do here.)
- [x] code-reference-parameters #173
- [x] code-reference-return-value #176
- [x] code-reference-explanation #177 
- [x] code-reference-source #172
- [x] code-reference-hooks #175
- [x] code-reference-related #170 
- [x] code-reference-changelog  https://github.com/WordPress/wporg-developer/commit/bdbaa48e713acf0de22e439e1e8e9db71f87aa70
- [x] code-reference-user-notes #178

**Blocks:**
- [x] Table of Contents

**Layout:**
- [x] Section heading styles. #190 
- [x] Fix layout from deprecated & private message https://github.com/WordPress/wporg-developer/commit/c3c2b1f732cf80a4624a2979f93e6850f38dc4fd
- [x] Add First Published/Last Update controls

## Screenshot

[Figma Link](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-Pages?node-id=1160%3A10750&t=pS0XoTexhVvCl9Ft-0)

![image](https://user-images.githubusercontent.com/1657336/212214836-ccb481e4-7837-4c0d-bc8d-bdcc154b53e5.png)

PR #167, opened 2023-01-16
Block: Implement deprecated block.

StevenDufresne:
Related to #162 

This PR implements the deprecated code block.

## Screenshots
| Desktop | Mobile | 
| --- | --- |
| localhost_8888_reference_functions_absint_ |  localhost_8888_reference_functions_absint_(iPhone 12 Pro) |

## Testing
It's a bit tricky seeing that our local environment doesn't have deprecated functions. Here is how I tested.

1. Set [this](https://github.com/WordPress/wporg-developer/blob/c186671beab3622f3c11b632bd3e281963616b82/source/wp-content/themes/wporg-developer-2023/inc/template-tags.php#L1096) to false.
2. You should now see the default banner.
3. To check that html rendering will work as expected, select HTML from the live site. [Example](https://developer.wordpress.org/reference/functions/get_blog_list/). Paste it in the DOM.

**Replace the content with this:**
```
<div class="callout callout-warning"><p><span class="screen-reader-text">Warning:</span> This function has been deprecated. Use <a href="https://developer.wordpress.org/reference/functions/wp_get_sites"></a><a href="https://developer.wordpress.org/reference/functions/wp_get_sites/" rel="function">wp_get_sites()</a>  instead.</p>
</div>
```

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/inc/template-tags.php
+++ b/source/wp-content/themes/wporg-developer-2023/inc/template-tags.php
@@ -1078,11 +1078,10 @@ namespace DevHub {
 	 * Retrieve deprecated notice.
 	 *
 	 * @param int  $post_id   Optional. Post ID. Default is the ID of the global `$post`.
-	 * @param bool $formatted Optional. Whether to format the deprecation message. Default true.
 	 * @return string Deprecated notice. If `$formatted` is true, will be output in markup
 	 *                for a callout box.
 	 */
-	function get_deprecated( $post_id = null, $formatted = true ) {
+	function get_deprecated( $post_id = null ) {
 		if ( ! $post_id ) {
 			$post_id = get_the_ID();
 		}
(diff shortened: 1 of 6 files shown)


outdoor2kode (approved):
👍 Tested and looks good.

ryelle (changes requested):
I don't see a design in the Developers mockup for this, but in the design library there are some callout boxes designed, so I imagine this should match those:

| Site | Mockup |
|----|------|
| Screenshot 2023-01-16 at 6 03 16 PM | Screenshot 2023-01-16 at 6 02 30 PM |

I see this HTML is coming from `get_deprecated` which in turn is built from `WPorg_Handbook_Callout_Boxes->warning_shortcode`, but I think this is a good place to refactor `get_deprecated` to return if not block syntax, something using the block class names for colors etc.

StevenDufresne:
> I don't see a design in the Developers mockup for this, but in the design library there are some callout boxes designed, so I imagine this should match those:

Good catch. Can I get a link to the design library?

StevenDufresne:
I'm going to return later to refactor. There are higher-priority issues to work on at the moment. I'll leave this PR hanging until then.

StevenDufresne:
I updated the code to use the notice component currently in https://github.com/WordPress/wporg-mu-plugins/pull/322.

Relevant commits:
- [42d2d4e](https://github.com/WordPress/wporg-developer/pull/167/commits/42d2d4ef07f8b1c379cbf21995a307369232b75b)
- [35562d7](https://github.com/WordPress/wporg-developer/pull/167/commits/35562d77a5ddcd1342b6a4d6cf014193c8e16c2c)

### Screenshot
Screen Shot 2023-01-19 at 3 40 50 PM

ryelle (approved):
Refactor looks good — `get_deprecated` is used by `get_changelog_data`, but that passes in `$formatted = false` so it should be straightforward to update when the changelog block is built.

I've removed the now-unused CSS, which made the CSS file unnecessary. I also updated the notice type to warning to match the red style of the old notice.

Screenshot 2023-01-20 at 12 36 08 PM

PR #168, opened 2023-01-16
Block: Connect private message block to function.

StevenDufresne:
Related to #162 

This PR implements the private access message.

## Screenshot
| Desktop | Mobile |
| --- | --- |
| localhost_8888_reference_functions_get_the_block_template_html_ | localhost_8888_reference_functions_get_the_block_template_html_(iPhone 12 Pro) |

## Testing
1. Visit [get_the_block_template_html](http://localhost:8888/reference/functions/get_the_block_template_html/)
2. Expect so see the message.

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/inc/template-tags.php
+++ b/source/wp-content/themes/wporg-developer-2023/inc/template-tags.php
@@ -1945,18 +1945,7 @@ namespace DevHub {
 			$alternative_string
 		);
 
-		// Use the 'alert' callout box if it's available. Otherwise, fall back to a theme-supported div class.
-		if ( class_exists( 'WPorg_Handbook_Callout_Boxes' ) ) {
-			$callout = new \WPorg_Handbook_Callout_Boxes();
-			$message = $callout->alert_shortcode( array(), $contents );
-		} else {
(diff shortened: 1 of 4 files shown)


outdoor2kode (approved):
👍 I can see that the layout is the same as the screenshots in the description in my local env.

This [import](https://github.com/WordPress/wporg-developer/blob/ad305f6ee90c16f3a2f5e8d1b1d233a73be932e1/source/wp-content/themes/wporg-developer-2023/src/code-private-access/index.js#L2) is no longer needed as the `style.scss` is removed.

ryelle (changes requested):
The `import './style.scss';` in index.js is breaking the build step, but removing that builds fine.

I have the same comment as #167 though — there's a design for these alert-style notices too, and this PR does not match it.
Screenshot 2023-01-16 at 6 15 11 PM

StevenDufresne:
This was updated to use the [notice block](https://github.com/WordPress/wporg-mu-plugins/pull/322).

## Screenshot
Screen Shot 2023-01-19 at 3 59 10 PM

ryelle (approved):
Looks good 👍🏻

Issue #161, opened 2023-01-13
Layout: Handbooks


StevenDufresne:
This is a big issue that should likely be broken up into multiple ones.

**Handbooks**
- [ ] Block editor
- [ ] Themes
- [ ] Plugins
- [ ] APIs
- [ ] Admin

## Screenshot

[Figma Link](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-Pages?node-id=1246%3A11634&t=pS0XoTexhVvCl9Ft-0)

![image](https://user-images.githubusercontent.com/1657336/212214601-8bf9bde1-8def-4c94-bfa9-941e159c9342.png)

ryelle:
@jasmussen How should the content here shift on mid-smaller screens? This mockup is 1440px wide, but what should happen on a 1300px or 1200px wide screen? For reference, at 1200px on [Documentation (with only 1 sidebar)](https://wordpress.org/documentation/article/introduction-to-blogging/) it shifts down to one column. A smaller viewport would mean the post content width would need to shrink, and it's already pretty narrow at 680px. The breakdown with one sidebar at 1200px:

```
|      |       main content      |      |      ToC      |      |
| 80px |          680px          | 25px |     335px     | 80px | = 1200px
```

(note, that gap between content & ToC shrinks from 145px to 25px with the viewport)

Sketching out this the same way, unless one of these can shrink more, we need to drop a column at 1440px.

```
|      |   chapters   |      |       main content      |      |      ToC      |      |
| 80px |     260px    | 20px |          680px          | 20px |     300px     | 80px | = 1440px
```

jasmussen:
I think we can probably treat mid-smaller screens same as mobile, and @javierarce had some good concepts there:

> a) hide the chapters on mobile (users will need to use the search function or go to the Developer homepage and then browse through the content) 
> b) display the chapter navigation inside the article (I would need help from you to know the possible locations for the list)
> c) hide the chapter + show an independent breadcrumb navigation above the title that would only appear on mobile

Of those B would appeal the most to me, show it above the main post. What do you think?

ryelle:
Just to be clear, this "mobile" view would be any screen size smaller than 1440px.

I just played around a bit with positioning, and this would be the easiest way to position things without hacky workarounds and to still keep the markup consistent for accessibility. Obviously this isn't totally styled yet, just for placement, but what do you think?

| size | image |
|------|-------|
[...]

PR #164, opened 2023-01-13
Implement front page

StevenDufresne:
Closes: #158 

This PR implements the front page.

**Questions:**
- The designs have links without `text-decoration: underline;`, but the new documentation site has them underlined. I assumed we want to make them the same. @javierarce @jasmussen
- I'm not certain where all the links should go to @javierarce @jasmussen

## Screenshot
| Mobile | Tablet  | Desktop |
| --- | --- | --- |
 localhost_8888_(iPhone 12 Pro) (2)| localhost_8888_ (1) |  localhost_8888_ (2) |

## Test
1. Add `Post Content` below to the front page content.
2. Replace broken URL in `front-page-content.php` with `https://user-images.githubusercontent.com/1657336/212520185-d9840a21-b149-49c6-b4f2-c043eaf969ea.svg`.
3. Visit homepage.

## Image For Testing

Link: https://user-images.githubusercontent.com/1657336/212520185-d9840a21-b149-49c6-b4f2-c043eaf969ea.svg

developers

PR #169, opened 2023-01-17
Add Block Editor Handbook page template

ryelle:
This is a first-pass at adding in the `single.html` from Documentation, for use on the handbook pages. Right now this only applies to the Block Editor Handbook, I don't have any page content in the other handbooks locally to test. This is a start for #161, but does not fix it because we still need the left column handbook nav. 

This also serves as a demo for https://github.com/WordPress/wporg-mu-plugins/pull/321.

**Screenshots**

| Desktop | Mobile |
|-------|------|
| fullpage | mobile |

**To test**

1. View a block editor handbook page, like `http://localhost:8888/block-editor/getting-started/create-block/block-code/`

To remove the old Table of Contents I added the same filter that [I added to the support-helphub plugin](https://meta.trac.wordpress.org/changeset/12327/) which uses this same code (not literally, unfortunately). Replace `load_filters` in `source/wp-content/plugins/handbook/inc/table-of-contents.php` with this:

```php
public function load_filters() {
	$page_supports_toc = is_singular( $this->post_types ) && ! is_embed() && ! is_front_page();
	/**
	 * Filter whether the table of contents should be injected into the page.
	 *
	 * @param bool $page_supports_toc True if the current page supports a table of contents.
	 */
	$should_add_toc = apply_filters( 'wporg_handbook_toc_should_add_toc', $page_supports_toc );
	if ( $should_add_toc ) {
		add_filter( 'the_content', array( $this, 'add_toc' ) );
	}
}
```


source/wp-content/themes/wporg-developer-2023/templates/single-blocks-handbook.html
```diff
@@ -0,0 +1,85 @@
...
+			<!-- wp:post-title {"level":1,"style":{"spacing":{"margin":{"top":"var:preset|spacing|edge-space","bottom":"var:preset|spacing|40"}}}} /-->
+
+			<!-- wp:group {"tagName":"aside","className":"sidebar-container"} -->
+			<aside class="wp-block-group sidebar-container">
+				<!-- wp:search {"label":"Search","showLabel":false,"widthUnit":"px","buttonText":"Search","buttonPosition":"button-inside","buttonUseIcon":true,"className":"is-style-secondary-search-control","borderColor":"light-grey-1","placeholder":"Search the documentation"} /-->
+
+				<!-- wp:wporg/table-of-contents {"style":{"elements":{"link":{"color":{"text":"var:preset|color|blueberry-1"}}},"spacing":{"padding":{"top":"var:preset|spacing|20","right":"var:preset|spacing|20","bottom":"var:preset|spacing|20","left":"var:preset|spacing|20"}}},"backgroundColor":"blueberry-4","textColor":"blueberry-1"} /-->
+
+				<!-- wp:paragraph {"fontSize":"small","className":"is-link-to-top"} -->
+				<p class="has-small-font-size is-link-to-top"><a href="#wp--skip-link--target">↑ Back to top</a></p>
```
StevenDufresne: We'll want this to be translated.

source/wp-content/themes/wporg-developer-2023/templates/single-blocks-handbook.html
```diff
@@ -0,0 +1,85 @@
...
+			<!-- /wp:group -->
+
+			<!-- wp:group {"style":{"spacing":{"blockGap":"0"}},"layout":{"type":"constrained"}} -->
+			<div class="wp-block-group">
+				<!-- wp:paragraph {"style":{"typography":{"fontStyle":"normal","fontWeight":"700"}}} -->
+				<p style="font-style:normal;font-weight:700">Changelog</p>
+				<!-- /wp:paragraph -->
+		
+				<!-- wp:paragraph {"className":"external-link"} -->
+				<p class="external-link"><a href="#">See list of changes</a></p>
```
StevenDufresne: Are we going to follow up with these links?

source/wp-content/themes/wporg-developer-2023/src/style/style.scss
```diff
@@ -14,3 +19,72 @@
...
+		}
+
+		&.is-fixed-sidebar .is-link-to-top,
+		&.is-bottom-sidebar .is-link-to-top {
+			display: block;
+		}
+	}
+}
+
+.wp-block-search.wp-block-search__button-inside .wp-block-search__inside-wrapper {
```
StevenDufresne, line 57: This is also being added in #164 
https://github.com/WordPress/wporg-developer/pull/164/files#diff-5a967de023463491bebbfaca0909be4be783855fcace73f96a9a0e9c8060224aR18

StevenDufresne:
A couple of minor comments. Works well 👍

source/wp-content/themes/wporg-developer-2023/templates/single-blocks-handbook.html
```diff
@@ -0,0 +1,85 @@
...
+	<!-- wp:group {"layout":{"type":"constrained","justifyContent":"left"},"align":"wide"} -->
+	<div class="wp-block-group alignwide">
+
+		<!-- wp:group {"tagName":"article"} -->
+		<article class="wp-block-group">
+			<!-- wp:post-title {"level":1,"style":{"spacing":{"margin":{"top":"var:preset|spacing|edge-space","bottom":"var:preset|spacing|40"}}}} /-->
+
+			<!-- wp:group {"tagName":"aside","className":"sidebar-container"} -->
+			<aside class="wp-block-group sidebar-container">
+				<!-- wp:search {"label":"Search","showLabel":false,"widthUnit":"px","buttonText":"Search","buttonPosition":"button-inside","buttonUseIcon":true,"className":"is-style-secondary-search-control","borderColor":"light-grey-1","placeholder":"Search the documentation"} /-->
```
StevenDufresne: This is dropping into the code search results. We'll need to update that at some point.

source/wp-content/themes/wporg-developer-2023/templates/single-blocks-handbook.html
ryelle: Yeah, there's still plenty to follow up with — this section also needs mobile styling, for example. I honestly meant to draft this PR but when I didn't I decided to let it be a first-pass & iterate in trunk 😅 

If you'd rather I keep iterating in the branch though, I can swap this into to draft mode.

StevenDufresne: Nope, merge it in when you think it's ready.

source/wp-content/themes/wporg-developer-2023/templates/single-blocks-handbook.html
ryelle: This is all done in a `searchform.php` template in the old theme… we can probably do something like [the global header search](https://github.com/WordPress/wporg-mu-plugins/blob/16ee1d69b809ac6d4e1b6963e0192f2dd4802af9/mu-plugins/blocks/global-header-footer/blocks.php#L999-L1016).

At least it looks like the search is mostly works as long as we set the form URL correctly:

Screen Shot 2023-01-24 at 16 13 26
http://localhost:8888/block-editor/?s=test

source/wp-content/themes/wporg-developer-2023/templates/single-blocks-handbook.html
ryelle: I've moved everything into a pattern and wrapped all these in translation functions. This content isn't synced from anywhere, so it's all been manually wrapped in i18n functions.

source/wp-content/themes/wporg-developer-2023/patterns/single-handbook.php
```diff
@@ -0,0 +1,90 @@
...
+			<!-- /wp:group -->
+
+			<!-- wp:group {"style":{"spacing":{"blockGap":"0"}},"layout":{"type":"constrained"}} -->
+			<div class="wp-block-group">
+				<!-- wp:paragraph {"style":{"typography":{"fontStyle":"normal","fontWeight":"700"}}} -->
+				<p style="font-style:normal;font-weight:700"><?php esc_html_e( 'Edit article', 'wporg' ); ?></p>
+				<!-- /wp:paragraph -->
+
+				<!-- wp:paragraph {"className":"external-link"} -->
+				<p class="external-link"><a href="#"><?php esc_html_e( 'Improve it on GitHub', 'wporg' ); ?></a></p>
```
StevenDufresne, line 65: How are these generated?

StevenDufresne (approved):
Good start!

source/wp-content/themes/wporg-developer-2023/patterns/single-handbook.php
ryelle: It looks like this _would_ be generated just by the `get_edit_post_link` function, but due to how patterns are rendered the current post isn't available yet, so we can't just use that function. We'll probably need to create some shortcodes for these links.

ryelle: And I don't think code to generate those other links exists at the moment, but I didn't look very deeply.

PR #170, opened 2023-01-18
Block: Related Code

StevenDufresne:
Related: #162 

This PR introduces two blocks `wporg/code-reference-related` and  `wporg/code-table` to render related code.

| Design | This PR | 
| --- | --- |
|  image| Screen Shot 2023-01-18 at 4 04 18 PM |

### Column Widths
In the designs, the "Description" column is larger than the one on the right. However, I think we should probably keep the widths of the columns `50%`. When there is both a `uses` and `used_by` table, the misalignment is visually unappealing we can't really maintain a `~30/70` split because there are some file names & functions that are pretty long. It currently is `50/50` on the live site.

| Fluid | 50% | 
| --- | --- |
| Screen Shot 2023-01-18 at 3 46 17 PM | Screen Shot 2023-01-18 at 3 48 28 PM |

### Toggle Naming
The designs maintain the same naming conventions for the "show more" toggles.
- "Show X more uses/Hide more uses"
- "Show X more used by/Hide more used by"

I think we can simplify and go with "Show X more/Show less". It makes the code simpler and I think it accomplishes the same thing.

### Mobile
I copied the same functionality that exists on the live site. The "description" column is hidden on mobile.

### Font Sizes
I followed the designs, `16px` but I personally think its noisey. We think we could consider using `14px` for these tables.

### `wporg/code-table` block
I created a block because the changelog block will need a table just like the ones here. 
 
| Changelog |
| --- |
| image |


source/wp-content/themes/wporg-developer-2023/src/code-related/block.php
```diff
@@ -25,16 +33,95 @@ function init() {
...
+	$uses_to_show     = 5;
+	$min_uses_to_show = 2;
+	$uses_to_show = min( $uses_to_show, max( split_uses_by_frequent_funcs( $uses->posts ), $min_uses_to_show ) );
+	$headings     = array( __( 'Uses', 'wporg' ), __( 'Description', 'wporg' ) );
+
+	return sprintf(
+		'<!-- wp:wporg/code-table {"id":"use","itemsToShow":%s,"headings": %s,"rows":%s } /--> ',
+		esc_attr( $uses_to_show ),
+		json_encode( $headings ),
+		json_encode( get_row_data( $uses ) )
```
adamwoodnz: PHPCS doesn't like this much

Screen Shot 2023-01-20 at 3 09 40 PM

`wp_json_encode` [does a few checks](https://developer.wordpress.org/reference/functions/wp_json_encode/)

source/wp-content/themes/wporg-developer-2023/src/code-related/block.php
```diff
@@ -25,16 +33,95 @@ function init() {
...
+ * Returns a table for uses.
+ */
+function get_uses_table( $uses ) {
+	$uses_to_show     = 5;
+	$min_uses_to_show = 2;
+	$uses_to_show = min( $uses_to_show, max( split_uses_by_frequent_funcs( $uses->posts ), $min_uses_to_show ) );
+	$headings     = array( __( 'Uses', 'wporg' ), __( 'Description', 'wporg' ) );
+
+	return sprintf(
+		'<!-- wp:wporg/code-table {"id":"use","itemsToShow":%s,"headings": %s,"rows":%s } /--> ',
```
adamwoodnz: Seems like these attributes don't match up with those in the table block.json:

```
"attributes": {
		"className": {
			"type": "string",
			"default": ""
		},
		"columns": {
			"type": "array",
			"default": []
		},
		"rows": {
			"type": "array",
			"default": []
		},
		"itemsToShow": {
			"type": "integer",
			"default": 5
		}
	},
```

source/wp-content/themes/wporg-developer-2023/src/code-table/view.js
```diff
@@ -0,0 +1,25 @@
+const init = () => {
```
adamwoodnz, line 1: This isn't running for me, I guess it's not enqueued yet

adamwoodnz (commented):
LGTM, a few minor comments inline

source/wp-content/themes/wporg-developer-2023/src/code-table/view.js
StevenDufresne: If you visit: http://localhost:8888/reference/functions/add_settings_section/, does the expanding/collapsing work?

source/wp-content/themes/wporg-developer-2023/src/code-related/block.php
StevenDufresne: Thanks for catching this.

StevenDufresne:
@ryelle I know you've done some [table generation in wporg-main-2022](https://github.com/WordPress/wporg-main-2022/blob/trunk/source/wp-content/themes/wporg-main-2022/src/release-tables/index.php), any opinion on this approach? 

The table doesn't have its editor controls and so it's set to `inserter: false`. Having a working Gutenberg table component  would require quite a bit more work in my estimation. I debated exposing some PHP functions to render a table as opposed to the block but a block felt more in line with where we are going.

source/wp-content/themes/wporg-developer-2023/src/code-table/view.js
adamwoodnz: Yep, expand/collapse works, but there isn't enough in related to display show less/more



On http://localhost:8888/reference/functions/do_blocks/ there is, but clicking on 'Show 2 more' navigates to `/#`

StevenDufresne: Hmm, suggests there's a `js` error. I'll take a look.

source/wp-content/themes/wporg-developer-2023/src/code-table/style.scss
```diff
@@ -0,0 +1,80 @@
...
+		th,
+		td {
+			border-right: none;
+			border-left: none;
+			padding: var(--wp--preset--spacing--10) var(--wp--preset--spacing--10) var(--wp--preset--spacing--10) 0;
+			font-size: var(--wp--preset--font-size--normal);
+			width: 50%;
+
+			&:first-child {
+				max-width: 340px; // 340 is half of the theme's content width.
```
ryelle: You could use `calc(var(--wp--style--global--content-size) / 2);` to be more self-explanatory & future-safe.

ryelle: Actually, is this necessary? It seems to wrap correctly in FF and Safari even without it.

source/wp-content/themes/wporg-developer-2023/src/code-related/block.php
```diff
@@ -25,16 +33,95 @@ function init() {
...
+function get_row_data( $posts ) {
+	$rows = array();
+
+	while ( $posts->have_posts() ) {
+		$posts->the_post();
+		$is_function = ! in_array( get_post_type(), array( 'wp-parser-class', 'wp-parser-hook' ) );
+
+		$rows[] = array(
+			sprintf(
+				'<a href="%s">%s%s</a><code>%s</code></a>',
```
ryelle: Could you use numbered placeholders here?

```suggestion
				'<a href="%1$s">%2$s%3$s</a><code>%4$s</code></a>',
```

source/wp-content/themes/wporg-developer-2023/src/code-related/block.php
```diff
@@ -25,16 +33,95 @@ function init() {
...
+
 	$title_block = sprintf(
 		'<h2 class="wp-block-heading">%s</h2>',
 		__( 'Related', 'wporg' )
 	);
 
 	$wrapper_attributes = get_block_wrapper_attributes();
 	return sprintf(
-		'<section %s>%s %s</section>',
+		'<section %s>%s %s %s</section>',
```
ryelle: Could you use numbered placeholders here?

```suggestion
		'<section %1$s>%2$s %3$s %4$s</section>',
```

source/wp-content/themes/wporg-developer-2023/src/code-table/block.php
```diff
@@ -0,0 +1,110 @@
...
+	);
+}
+
+/**
+ * Render the block content.
+ *
+ * @return string Returns the block markup.
+ */
+function render( $attributes ) {
+	$table  = '<!-- wp:table {"className":"' . $attributes['className'] . '"} --><figure class="wp-block-table ' . $attributes['className'] . '">';
```
ryelle, line 28: I moved the spacing value onto the `wporg/code-table`s in `wporg/code-reference-related`, in case you ever need to configure it. it also makes this output a little cleaner. You can drop the commit if you disagree fe70065ee828d9cd5a2b96b3d2b24dff74ec77e4

ryelle (commented):
A few small comments, but this is working for me generally.

It would be nice to do an iteration on the show/hide actions to improve the accessibility - focus management & improved aria - but that's not a blocker here. Once all the subtasks in #162 are done we can do an a11y review.

source/wp-content/themes/wporg-developer-2023/src/code-table/style.scss
StevenDufresne: You are right, making the anchor tag [break-all](https://github.com/WordPress/wporg-developer/pull/170/files#diff-30ff2f768464a6da288430c5189306d3ddf722ce68e75e10f36b2423a6b647ceR32) ended up sorting it.

source/wp-content/themes/wporg-developer-2023/src/code-table/view.js
StevenDufresne: I ran into this running wp 6.1-alpha. Updating WordPress solved it. The automatic adding of the `viewScript` was added in 6.1: https://core.trac.wordpress.org/ticket/56470

PR #171, opened 2023-01-19
Block: Code-title component.

StevenDufresne:
Related #162 

This PR moves and simplifies code related to the display line of code. 

| Type | |
| --- | --- |
| Method | Screen Shot 2023-01-19 at 11 24 19 AM |
| Function | Screen Shot 2023-01-19 at 11 23 45 AM |
| Hook | Screen Shot 2023-01-19 at 11 23 21 AM |
| Class | Screen Shot 2023-01-19 at 11 28 25 AM |

## Block Editor
I added the ability to control the container tag because this is also needed for rendering hooks.

Screen Shot 2023-01-19 at 11 38 46 AM

## Testing
1. Load some single pages:
   - [wp_posts_list_table/]( http://localhost:8888/reference/classes/wp_posts_list_table/)
   - [wp_widget_rss/update/]( http://localhost:8888/reference/classes/wp_widget_rss/update/)
   - [add_menu_page/](http://localhost:8888/reference/functions/add_menu_page/)


source/wp-content/themes/wporg-developer-2023/src/code-title/style.scss
```diff
@@ -1 +1,39 @@
 /* css styles */
+.wp-block-wporg-code-reference-title {
+	> * {
+		font-size: var(--wp--preset--font-size--extra-large);
+		line-height: 42px;
```
adamwoodnz: Best practice to use unitless, see https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#values
```suggestion
		line-height: 1.73;
```

(value based on Figma)

source/wp-content/themes/wporg-developer-2023/src/code-title/style.scss
```diff
@@ -1 +1,39 @@
...
+		color: #008a20;
```
adamwoodnz: I can see these colors are also used for code examples; is it worth adding presets for them, or sharing some other way so that we only have to change them in one place?

ryelle: I agree these should be defined somewhere specific.
We can't add presets to `theme.json` without duplicating all the existing values, but we could sorta fake it in the CSS by creating some new custom properties. The only drawback is that they won't be available in the editor itself.

source/wp-content/themes/wporg-developer-2023/src/code-title/style.scss
```diff
@@ -1 +1,39 @@
...
+	}
+
+	.hook-func,
+	.keyword {
+		color: #646970;
+	}
+
+	.arg-type {
+		color: #d63638;
+		font-style: italic;
```
ryelle: Looks like this is not italic in the design:

Screenshot 2023-01-20 at 1 38 25 PM

(though the italic style is fancy)

ryelle:
I added the `postId` context to the block.json so that the value would be passed in to the render function & we can use that -- it's just a little safer than relying on the global object for if/when we use this in a Query Loop or other nested context.

I also added an `isLink` property since otherwise the link was never being rendered. You can test it by manually setting a link on `<!-- wp:wporg/code-reference-title {"isLink":true} /-->`.

ryelle:
So I'm not sure the best approach in general, but it looks like the lack of `useBlockProps` in the edit function breaks the block in the editor… but including it _definitely_ applies the class names & styles twice.

Screenshot 2023-01-20 at 4 35 21 PM

After trying a janky workaround with checking `REST_REQUEST` in render, I looked at how the core blocks handle this and found that `ServerSideRender` has a prop for `skipBlockSupportAttributes` which solves the problem 😑 

So I've also enabled the dimension support for this block, and tidied up the output & CSS. By not nesting the element in a div wrapper, we can apply the font size directly to it, and when custom font sizes are used it just works.

Lastly, I tried a local palette for the syntax colors (see https://github.com/WordPress/wporg-developer/pull/171/commits/cf8bc8f70463c88b1b1417b6a43a7fa4cd763cac), which does generate these custom properties (on the `.wp-block-wporg-code-reference-title` class). In this way, they're not reusable across blocks, but it's a step in that direction. This `palette` value overrides the base theme one, so if this block had a color control you would only see these 5 colors available (no more blueberry, charcoal, etc). I don't really know if that's any better than defining them in CSS and not having them in the editor? So we can iterate there.

StevenDufresne:
Thanks for all the help!

source/wp-content/themes/wporg-developer-2023/src/code-title/block.php
```diff
@@ -24,14 +24,33 @@ function init() {
...
  * @return string Returns the block markup.
  */
-function render( $attributes ) {
+function render( $attributes, $content, $block ) {
+	if ( ! isset( $block->context['postId'] ) ) {
+		return '';
+	}
+
+	$post_ID = $block->context['postId'];
+	$content = get_signature( $post_ID );
```
StevenDufresne, line 39: Any idea on how we will be able to render this dynamic block inside another dynamic block with the right context? In the hooks component, we'll need to do that. We'll need to loop through the hooks and construct a couple of these blocks.

I can't seem to find a way to retrieve a string of a _rendered_ dynamic block without using `do_blocks` which outputs the block to the document. 

My only other thought is to pass the `postId` through.

StevenDufresne: Actually, disregard this, it appears like I had a typo that was causing the immediate output. 👍 `do_blocks` doesn't output itself :).

PR #172, opened 2023-01-20
Block: Source code

StevenDufresne:
Related to: #162 

This PR updates the code-source block to match designs.

| [Designs](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-Pages?node-id=1160%3A10750&t=t7ekzK2bBZF5Ku9l-0) | PR |
| --- | --- |
| Screen Shot 2023-01-20 at 11 11 52 AM |  Screen Shot 2023-01-20 at 11 30 27 AM |

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/js/function-reference.js
+++ b/source/wp-content/themes/wporg-developer-2023/js/function-reference.js
@@ -75,7 +75,9 @@ jQuery( function ( $ ) {
 			$element.data( 'height', originalHeight );
 
 			const $expandButtonContainer = $( '<div class="wp-block-button is-style-outline"></div>' );
-			const $expandButton = $( '<a href="#" class="wp-block-button__link wp-element-button"></a>' );
+			const $expandButton = $(
+				'<a href="#" class="wp-block-button__link wp-element-button" aria-controls="wporg-source-code"></a>'
+			);
 			$expandButton.on( 'click', function ( event ) {
 				event.preventDefault();
 				if ( 'true' === $expandButton.attr( 'aria-expanded' ) ) {
--- a/source/wp-content/themes/wporg-developer-2023/js/function-reference.js
+++ b/source/wp-content/themes/wporg-developer-2023/js/function-reference.js
@@ -34,9 +34,13 @@ jQuery( function ( $ ) {
 
 		$element.wrap( '<div class="wporg-developer-code-block"></div>' );
 
-		const $copyButton = $( document.createElement( 'button' ) );
+		const $copyButtonContainer = $( '<div class="wp-block-button is-style-outline"></div>' );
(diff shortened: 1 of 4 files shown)


source/wp-content/themes/wporg-developer-2023/js/function-reference.js
```diff
@@ -60,16 +64,20 @@ jQuery( function ( $ ) {
...
 		// Check code block height, and if it's larger, add in the collapse
 		// button, and set it to be collapsed differently.
 		const originalHeight = $element.height();
 		if ( originalHeight > MIN_HEIGHT ) {
 			$element.data( 'height', originalHeight );
 
-			const $expandButton = $( document.createElement( 'button' ) );
-			$expandButton.on( 'click', function () {
+			const $expandButtonContainer = $( '<div class="wp-block-button is-style-outline"></div>' );
+			const $expandButton = $( '<a href="#" class="wp-block-button__link wp-element-button"></a>' );
```
adamwoodnz: Seems like it would be a good a11y improvement to also add an `aria-controls` attr to this, linking it to the source code wrapper, see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded

source/wp-content/themes/wporg-developer-2023/src/code-source/style.scss
```diff
@@ -1 +1,72 @@
...
+
+	&::after {
+		content: "\b7\a0";
+		position: absolute;
+		right: -4px;
+	}
+
+	&:last-child::after {
+		display: none;
+	}
```
adamwoodnz, line 70: Slightly simpler
```suggestion
	&:not(:last-child)::after {
		content: "\b7\a0";
		position: absolute;
		right: -4px;
	}
```

adamwoodnz:
The spacing on this is a bit tighter in the design; 8px padding, with 4px between.

Maybe you're trying to stick to our theme vars, but worth calling out

Design
Screen Shot 2023-01-23 at 2 19 07 PM

Implementation
Screen Shot 2023-01-23 at 2 19 15 PM

adamwoodnz (commented):
Nice work, few minor comments inline

source/wp-content/themes/wporg-developer-2023/src/code-source/style.scss
StevenDufresne: Yeah, we do get some users that have browsers that don't support `:not` and for things that are visually rendered, in this case, a dot, I would lean towards support. So, I don't see a compelling reason to use `:not` here.

StevenDufresne:
> The spacing on this is a bit tighter in the design; 8px padding, with 4px between.
> 
> Maybe you're trying to stick to our theme vars, but worth calling out

Yeah exactly. I prefer using our current grid where possible.

PR #174, opened 2023-01-24
Blocks: Add spacing & line-height support to all blocks

ryelle:
This updates all the blocks to support top & bottom margin, and padding (all sides). It also updates the typography support to support changing the line height as well as font size, since usually when one changes, the other needs to change too.

Try adding spacing to any of the blocks, for example:

```
<!-- wp:wporg/code-reference-title {"style":{"spacing":{"margin":{"top":"var:preset|spacing|edge-space","bottom":"var:preset|spacing|40"}}}} /-->
```

On `trunk`, that won't render any margins. On this branch, the margins are added to the element.


StevenDufresne (approved):
👍 LGTM.

PR #177, opened 2023-01-26
Block: Fix explanation

StevenDufresne:
Related To: #162 

This PR styles the "More Information" section when viewing the code. This adds an style that styles `<code>` tags to be used by other components. This style seems to be common accross a few components.

## Screenshots
Screen Shot 2023-01-26 at 11 28 52 AM

## How to test
Explanations are rich content explanations. I've gone through a number of them in production and nothing jumps out as complicated. They do include some basic HTML tags, including `<code>`

1. Visit http://localhost:8888/wp-admin/edit.php?post_type=wp-parser-function
2. Hover over item in the table, click "Add Explanation"
3. Add content into the explanation textarea and save.
4. Visit that `function`
5. Verify the "More Information" section appears

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/code-explanation/block.php
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-explanation/block.php
@@ -38,7 +38,7 @@ function render() {
 		__( 'More Information', 'wporg' )
 	);
 
-	$wrapper_attributes = get_block_wrapper_attributes();
+	$wrapper_attributes = get_block_wrapper_attributes( array( 'class' => 'wporg-has-embedded-code' ) );
 	return sprintf(
 		'<section %s>%s %s</section>',
 		$wrapper_attributes,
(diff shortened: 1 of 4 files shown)


ryelle (commented):
Is the `wporg-has-embedded-code` class needed for anything else? because you could use `.wp-block-wporg-code-reference-explanation code` in the CSS instead.

There are some complex explanations - for example, [add_menu_page](https://developer.wordpress.org/reference/functions/add_menu_page/) has a `php` shortcode. The CSS here is conflicting with the code block style:

Screenshot 2023-01-26 at 12 04 01 PM

It looks like [a bunch of explanations have PHP shortcodes](https://developer.wordpress.org/wp-admin/edit.php?s=%5Bphp%5D&post_status=all&post_type=wporg_explanations&action=-1&m=0&paged=1&action2=-1), so this isn't totally an edge case.

ryelle:
Ah, I caught the class name also used in https://github.com/WordPress/wporg-developer/pull/176, so you can ignore that 👍🏻

StevenDufresne:
> It looks like [a bunch of explanations have PHP shortcodes](https://developer.wordpress.org/wp-admin/edit.php?s=%5Bphp%5D&post_status=all&post_type=wporg_explanations&action=-1&m=0&paged=1&action2=-1), so this isn't totally an edge case.

Had no idea you could search like that! 

Thanks: [60b05a1](https://github.com/WordPress/wporg-developer/pull/177/commits/60b05a16cc9d468ccc16f13bba4635266f08e76f).

PR #178, opened 2023-01-26
Block: User Notes

StevenDufresne:
Related: #162 

This PR grew in scope and hopefully isn't too unwieldy but since multiple views utilized a couple of functions to render the comment form, it was tricky to update it without taking on all the contexts. Here's a brief explanation of what was going on.

There are 3 different contexts for when a comment form was loaded:
  - To submit a note on the function/hook/etc...
  - To submit feedback on a note on the function/hook/etc...
  - To edit a note or feedback for non-admin users. 

These three contexts call the same functions to retrieve form arguments and manage layouts. Since there is a lot of functionality under the hood, I did my best to change as little but still separate out some of the logic.

# Leaving Note (Comment)
Component: `<!--wp:wporg/code-reference-comment-form /-->`

**Details**
- Must log in to add a note
- Can "preview" the note

Note: This is not styled properly in this PR because it will rely on the layout of the template which doesn't exist here yet. I'll fix that in a follow-up.

| Screenshot |
| --- |
| Screen Shot 2023-01-31 at 2 09 47 PM |
| Screen Shot 2023-01-31 at 2 09 16 PM |
| Screen Shot 2023-01-31 at 2 09 33 PM |

# View Notes/ Leaving Feedback
Component: `<!--wp:wporg/code-reference-comments/-->`

**Details**
- Must log in to add feedback
- Show's feedback count
- Feedback is collapsible
- Feedback form is available via "Add feedback/Hide feedback form" link
- Admins can see edit link for all entries
  - Url is different that normal users, they can edit via `wp-admin` 
- Users should be able to see when their feedback/note is awaiting moderation
- Users should have a link to edit "awaiting moderation" feedback/notes
- Feedback is slightly different than main
  - No preview
  - Different rules displayed

| Context | Screenshot |
| --- | --- |
| Must log in | Screen Shot 2023-01-31 at 2 07 43 PM |
| Feedback form toggle | Screen Shot 2023-01-31 at 2 10 19 PM |
| Feedback awaiting moderation | Screen Shot 2023-01-31 at 2 10 42 PM |
| Expanded feedback, multiple | Screen Shot 2023-01-31 at 2 09 00 PM |
| Expanded feedback & feedback form |  Screen Shot 2023-01-31 at 2 09 09 PM |

### Considerations:
- The current PHP templates appear to support [comment paging](https://github.com/WordPress/wporg-developer/blob/old/source/wp-content/themes/wporg-developer/comments.php#L26) however, I can't get the results to page after setting the "[wp-admin/discussion](http://localhost:8888/wp-admin/options-discussion.php)" appropriately. I'm not familiar with the underlying mechanism, but it appears like paging was never supported at the query level. My assumption is that there would be paging [logic in the query](https://github.com/WordPress/wporg-developer/blob/c50d4dff91e49ed3f0b572d788ba4a54537ec746/source/wp-content/themes/wporg-developer-2023/inc/template-tags.php#L96). Additionally, comment paging configurations are not active in production. So with all that being said, this PR does not support comment paging even though the existing template did. I think that's ok.

# Editing Note/Feedback

Component: `<!--wp:wporg/code-reference-comment-edit /-->`

**Details**
- Admin's can always edit (although they do so via `wp-admin`
- Users can edit their note/feedback when they are awaiting moderation
- Template handling needed to be done via redirect. ([See code](https://github.com/WordPress/wporg-developer/pull/178/files#diff-169de219308bc77da5b199c533af3e8f30eadd5d8055e6feb74d5aed496feca4R135)) 

| Context | Screenshot |
| --- | --- |
|  Editing Feedback | Screen Shot 2023-01-31 at 2 10 50 PM |
|Editing Top Level Note | Screen Shot 2023-01-31 at 2 11 01 PM |

### How to test
You'll want to have 2 users:
 - admin
 - non-admin

**Non-admin User**
- Navigation to function
- Leave a note
- On page reload, expect to see the note "awaiting moderation"
- Click Edit
- Edit the note
  - Verify breadcrumb makes sense 
- On page reload, expect to see your note with the new edits.

**Admin User**
- Navigate to http://localhost:8888/wp-admin/edit-comments.php, approved comments
- Visit said function
- Expect to see a non-admin note
- Add Feedback
- Verify the feedback was added
- Click to edit the non-admin note
- Verify that you are taken to the admin panel to edit,
- Edit note
- View the original function, and expect to see edits

**Not logged in**
- Navigate to function with no notes, expect to see none.
  - Verify the form at the bottom requires you to log in
- Navigate to the function with some notes, expect to see them
  - Verify that under the note it requires you to log in


adamwoodnz:
Nice work, epic PR! I've run this and everything worked smoothly for me. I'll add some code review tomorrow.

StevenDufresne:
**Todo**
- [x] Remove the file name in the code block for comments.

outdoor2kode:
It's awesome! 💯  Done testing all steps instructed in How to test, and it works as expected. Its behaviors are the same as the ones in production. Halfway through the code review, will continue doing it tmr morning.

source/wp-content/themes/wporg-developer-2023/src/code-comment-form/block.php
```diff
@@ -0,0 +1,90 @@
...
+		$form = ob_get_clean(); // End capture
+
+		$editor_rules = \DevHub_User_Submitted_Content::get_editor_rules();
+		$output = "<div class='wp-block-wporg-code-reference-comment-form-content'>{$form}<div>{$editor_rules}</div></div>";
+
+	} else {
+		$output = '<p>' . sprintf(
+			/* translators: %s: login URL */
+			__( 'You must <a href="%s">log in</a> before being able to contribute a note or feedback.', 'wporg' ),
+			esc_url('https://login.wordpress.org/?redirect_to=' . urlencode( get_permalink() ) )
```
ryelle, line 52: Could you use [wp_login_url](https://developer.wordpress.org/reference/functions/wp_login_url/) here? It's not _that_ important, but it would help in local testing.

ryelle:
Looking good! I've scanned the code & done some basic testing, and it looks like it's working well 🏆 

> So with all that being said, this PR does not support comment paging even though the existing template did. 

I just did a quick check of the current site, and the item with the most comments is [`register_post_type`](https://developer.wordpress.org/reference/functions/register_post_type/) at 30, with only 8 others in the 20+ range. And I don't actually see pagination on that page, so I think the decision to not include it is fine.

I think the arrows for voting need an iteration - they're very small, and the use of `title` for feedback is not very clear, but that's not a blocker for this PR.

source/wp-content/themes/wporg-developer-2023/src/code-comment-form/block.php
StevenDufresne: I copied this over from the old code because trying to use `wp_login_url` wasn't working. I can't remember why. 

I'll take another look.

StevenDufresne:
> I think the arrows for voting need an iteration - they're very small, and the use of `title` for feedback is not very clear, but that's not a blocker for this PR.

Very fair. I'll try to bump the size up a bit and merge. I'll leave the "title" stuff for a later day.

PR #181, opened 2023-01-27
Blocks: Add postId to all blocks, fix ServerSideRender display

ryelle:
There are updates here that touch every code-reference block:
- **Use a shared edit component to render dynamic blocks**
    Almost all the edit components were the same, and while I was going to just add the `skipBlockSupportAttributes` prop to all the `ServerSideRender`s (see https://github.com/WordPress/wporg-developer/pull/171#issuecomment-1399037104), I thought this might be better. Instead of each block with an edit file, there's now a single `shared/dynamic-edit.js` which each block can call (and it includes the `skipBlockSupportAttributes`).
- **Add postId context to blocks**
    Since blocks can be used in different contexts, this ensures the global `post` being used to show information is the correct post. This will let us use the blocks in, for example, a Query Loop, for archive views.
    For ServerSideRender to understand this value, we need to pass it through via `urlQueryArgs`, but since the dynamic edit component exists, this is a much simpler change.

**To test:**

- Check that there are no regressions on the code reference page
- Try these out in the editor with a Query Loop block, for example:

```html
<!-- wp:query {"queryId":1,"query":{"perPage":"10","pages":0,"offset":0,"postType":"wp-parser-hook","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":false,"taxQuery":{"wp-parser-since":[103]},"parents":[]},"displayLayout":{"type":"list","columns":3}} -->
<div class="wp-block-query"><!-- wp:post-template -->
<!-- wp:wporg/code-reference-title {"tagName":"p"} /-->

<!-- wp:wporg/code-reference-summary /-->
<!-- /wp:post-template --></div>
<!-- /wp:query -->
```

Screenshot 2023-01-27 at 4 34 09 PM

Note — if you display the functions in a Query Loop, an extra `()` is added [thanks to this filter](https://github.com/WordPress/wporg-developer/blob/6b10d20524048cfca6324b9509d1ad2e8bad8c1c/source/wp-content/themes/wporg-developer-2023/inc/extras.php#L39). This doesn't mess with anything real, yet, so I've left it alone.

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/code-changelog/block.json
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-changelog/block.json
@@ -7,6 +7,7 @@
 	"category": "widgets",
 	"icon": "smiley",
 	"description": "Show the changelog for a function, class, or method.",
+	"usesContext": [ "postId" ],
 	"supports": {
 		"html": false,
 		"spacing": {
--- a/source/wp-content/themes/wporg-developer-2023/src/code-changelog/block.php
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-changelog/block.php
@@ -22,9 +22,13 @@ function init() {
 /**
  * Render the block content.
  *
+ * @param array    $attributes Block attributes.
+ * @param string   $content    Block default content.
+ * @param WP_Block $block      Block instance.
+ *
  * @return string Returns the block markup.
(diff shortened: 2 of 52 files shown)


StevenDufresne:
I like it. Go for it!

PR #189, opened 2023-01-30
New block: Chapter List for handbooks

ryelle:
See #161 — This PR adds the Chapter List block, which is used on the handbook pages to display the navigation in a given handbook. It uses the current post type to list all items in that post type, using the page hierarchy to build up a menu with submenus. 

It also adds styles to the theme to create the three-column layout when the screen is wider than 1440px, and a two col layout from 1200px and up (see https://github.com/WordPress/wporg-developer/issues/161#issuecomment-1406782932).

**Screenshots**

| Small screen (mobile) | Medium (small-mid desktop) | Large (large desktop) |
|---------------|-----------|--------------|
| wpcs-small | wpcs-mid | wpcs-large |

Other handbooks - on production, the Block Editor handbook also uses the page hierarchy, but Plugins & REST API use a navigation menu… so the chapter list on these is a little wonky. For consistency, I'd like to use this same block on all the handbooks, but I'm not sure if there's a reason to use a nav menu instead on these other pages. For now, all are using the Chapter List block.

| Block Editor | Plugins | REST API |
|----------|-----------|--------|
| block-editor | plugins | rest-api |

**To test**

- Make sure you have handbook content, either by exporting from the production site or by running the parser (`yearn wp-env run cli "wp cron event run --all"` has worked for me).
- View some pages in a handbook
- The menu should appear, and clicking through it should work
- Expanding & collapsing sections should work
- If you land on a nested page, the menu should be open to that page - ex, http://localhost:8888/block-editor/how-to-guides/platform/custom-block-editor/tutorial/

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/chapter-list/view.js
+++ b/source/wp-content/themes/wporg-developer-2023/src/chapter-list/view.js
@@ -1,0 +1,44 @@
+const init = () => {
+	const container = document.querySelector( '.wp-block-wporg-chapter-list > ul' );
+	if ( container ) {
+		container.parentNode.classList.toggle( 'has-js-control' );
+
+		const parents = container.querySelectorAll( '.page_item_has_children' );
+		parents.forEach( ( item ) => {
+			// Get link, remove (will re-ad later).
+			const link = item.querySelector( ':scope > a' );
+			link.remove();
+
+			// Get submenu
+			const submenu = item.querySelector( ':scope > ul' );
+
+			// Create the toggle button.
+			const button = document.createElement( 'button' );
+			button.setAttribute( 'aria-expanded', false );
+			// button.setAttribute( 'aria-label', '' );
+			button.onclick = () => {
+				submenu.classList.toggle( 'is-open' );
+				// This attribute returns a string.
+				const isOpen = button.getAttribute( 'aria-expanded' );
+				button.setAttribute( 'aria-expanded', isOpen === 'false' );
+			};
+
+			const buttonGroup = document.createElement( 'span' );
+			buttonGroup.className = 'wporg-chapter-list--button-group';
+			buttonGroup.append( button, link );
+
+			item.insertBefore( buttonGroup, submenu );
+
+			// Automatically open the trail to the current page.
+			if (
+				item.classList.contains( 'current_page_item' ) ||
+				item.classList.contains( 'current_page_ancestor' )
+			) {
+				submenu.classList.toggle( 'is-open' );
+				button.setAttribute( 'aria-expanded', true );
+			}
+		} );
+	}
+};
+
+window.addEventListener( 'load', init );
--- a/source/wp-content/themes/wporg-developer-2023/src/chapter-list/view.js
+++ b/source/wp-content/themes/wporg-developer-2023/src/chapter-list/view.js
@@ -1,3 +1,8 @@
+/**
+ * WordPress dependencies
+ */
+import { __, sprintf } from '@wordpress/i18n';
+
 const init = () => {
 	const container = document.querySelector( '.wp-block-wporg-chapter-list > ul' );
 	if ( container ) {
@@ -15,12 +20,26 @@ const init = () => {
 			// Create the toggle button.
 			const button = document.createElement( 'button' );
 			button.setAttribute( 'aria-expanded', false );
-			// button.setAttribute( 'aria-label', '' );
+			// translators: %s link title.
+			button.setAttribute( 'aria-label', sprintf( __( 'Open %s submenu', 'wporg' ), link.innerText ) );
 			button.onclick = () => {
 				submenu.classList.toggle( 'is-open' );
 				// This attribute returns a string.
 				const isOpen = button.getAttribute( 'aria-expanded' );
 				button.setAttribute( 'aria-expanded', isOpen === 'false' );
+				if ( isOpen === 'false' ) {
+					button.setAttribute(
+						'aria-label',
+						// translators: %s link title.
+						sprintf( __( 'Close %s submenu', 'wporg' ), link.innerText )
+					);
+				} else {
+					button.setAttribute(
+						'aria-label',
+						// translators: %s link title.
+						sprintf( __( 'Open %s submenu', 'wporg' ), link.innerText )
+					);
+				}
 			};
 
 			const buttonGroup = document.createElement( 'span' );
@@ -36,6 +55,11 @@ const init = () => {
 			) {
 				submenu.classList.toggle( 'is-open' );
 				button.setAttribute( 'aria-expanded', true );
+				button.setAttribute(
+					'aria-label',
+					// translators: %s link title.
+					sprintf( __( 'Close %s submenu', 'wporg' ), link.innerText )
+				);
 			}
 		} );
 	}
--- a/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
+++ b/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
@@ -27,13 +27,13 @@ body[class] {
 // Slot the search & table of contents into a floating sidebar on large screens.
 @media (min-width: 1200px) {
 	.sidebar-container {
-		--local--sidebar-width: 340px;
+		--local--block-end-sidebar--width: 340px;
 
 		position: absolute;
 		top: calc(var(--wp-global-header-offset, 0px) + var(--wp-local-header-offset, 0px));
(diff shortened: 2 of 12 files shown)


source/wp-content/themes/wporg-developer-2023/src/chapter-list/view.js
```diff
@@ -0,0 +1,44 @@
...
+			const link = item.querySelector( ':scope > a' );
+			link.remove();
+
+			// Get submenu
+			const submenu = item.querySelector( ':scope > ul' );
+
+			// Create the toggle button.
+			const button = document.createElement( 'button' );
+			button.setAttribute( 'aria-expanded', false );
+			// button.setAttribute( 'aria-label', '' );
```
ryelle: I meant to follow up with this— this should be something like "open [page name] submenu", will update the PR tomorrow.

source/wp-content/themes/wporg-developer-2023/theme.json
```diff
@@ -3,37 +3,79 @@
...
+						"color": {
+							"text": "var:preset|color|blueberry-1"
+						}
+					}
+				}
+			},
+			"wporg/table-of-contents": {
+				"color": {
+					"background": "var:preset|color|blueberry-4",
+					"text": "var:preset|color|blueberry-1"
```
StevenDufresne: Why do this use the `var:` format and the other ones don't? Is there a limitation?

source/wp-content/themes/wporg-developer-2023/src/style/style.scss
```diff
@@ -54,6 +51,19 @@ body[class] {
...
+@media (min-width: 1440px) {
+	.has-three-columns .sidebar-container {
+		--local--sidebar-width: 300px;
+		--wp--style--global--wide-size: 1280px;
+	}
+
+	.has-three-columns {
+		display: grid;
+		grid-template-columns: 260px 1fr;
+		gap: 20px;
```
StevenDufresne: Should this use our spacing grid: `var(--wp--preset--spacing--20)`?

StevenDufresne:
This is looking sharp!

### Can we reduce the amount of jank on load?
- Left to Right movement
- Top-to-bottom movement.

| Gif |
| --- |
|  |

### Nested line alignment

This should be alignment with the middle of the arrow:

Screen Shot 2023-01-31 at 3 27 44 PM

###  Selected item
The designs have a solid line to identify the selected page
| Design | PR |
| --- | --- |
| Screen Shot 2023-01-31 at 3 40 33 PM | Screen Shot 2023-01-31 at 3 39 27 PM |

### Click target
Since the title of the section is a link and doesn't toggle the section, the only way to toggle the section is to click the arrow. The arrow itself is very small and probably requires expert-level motor skills. Can we make it larger?

Screen Shot 2023-01-31 at 3 31 23 PM

### Layout

At `1250px` I got this layout:

localhost_8888_block-editor_reference-guides_richtext_

jasmussen:
Looks like it can work as a good starting point. The margin looks right on desktop in the last screenshot Steve shared. The margin below the chapters in the first screenshots in this PR, however, looks a little small.

source/wp-content/themes/wporg-developer-2023/theme.json
ryelle: No, probably just the way I copied them - swapped in the custom properties 👍🏻

ryelle:
> Can we reduce the amount of jank on load?

Good catch, I styled the default list a bit more so it won't shift left/right when the content loads. The collapse shift happens because the menu needs to be expanded to work if the JS doesn't load.

No JS view:
Screenshot 2023-01-31 at 11 33 53 AM

> Nested line alignment

Fixed it, though it might appear slightly off on different browsers because it needs to be on a fractional pixel to be centered (even width).

> Selected item

This is a question for design, but the menu as mocked up only has two levels, so the child elements don't need the <kbd>-</kbd> or <kbd>v</kbd> button. Since menus in reality can be arbitrarily deep, all levels use the buttons, and it seemed like having both the button + the border was going to be busy. I'm also not sure how the current border should work on parent items. @WordPress/meta-design What do you think, do we need that border somewhere on here?

Screenshot 2023-01-31 at 12 22 26 PM

> Click target

I've made this a bit larger, so it matches the full line-height of the link. It's almost 24px, which is the [recommended minimum target size for WCAG 2.2 AA](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum). (WCAG 2.2 is still technically draft, but this is the clearest non-touch target advice I found)

Screenshot 2023-01-31 at 12 23 27 PM

> Layout
> At 1250px I got this layout:

Yeah, from 1200-1440 it's a two column layout. This does need some design iteration, especially on a page like the component or package refs where the menu is _very_ long



> The margin looks right on desktop in the last screenshot Steve shared. The margin below the chapters in the first screenshots in this PR, however, looks a little small.

The margin there is the `edge-spacing` value, which is what the table of contents box uses — on small screens (at 890px), that size shrinks to 20px. I've updated it to always be 80px, but that looks strange in context with the rest of the margins at 20px. [These are the other spacing values we could use](https://github.com/WordPress/wporg-parent-2021/#spacing), if one of those would be better.

Screen Shot 2023-01-31 at 12 37 16

StevenDufresne:
Thanks for the updates! 

A couple of minor things that aren't blockers:

### List style with no children
Since we use the blueberry color primarily for actionable elements within content, I wonder if we should consider changing the visual appearance of the "-" in the chapter list. It looks like you would be able to interact with it but it's not an interactive element since it has no children. I've mocked up some examples to illustrate the difference. I don't know. Thinkout out loud. @WordPress/meta-design, @jasmussen 

| Current Design | No dashes | Subdued Dashes | 
| --- | --- | --- |
| Screen Shot 2023-02-01 at 9 16 15 AM | Screen Shot 2023-02-01 at 9 16 03 AM |  Screen Shot 2023-02-01 at 9 14 07 AM |





### First item

The first item in the chapter list is the name of the handbook and no longer gets a different visual treatment. Should we consider removing it (or removing the dash)? It's currently presented as a sibling of "Get Started" when in fact it's the parent.

| Current | This PR |
| --- | --- |
| Screen Shot 2023-02-01 at 9 02 09 AM | Screen Shot 2023-02-01 at 9 02 04 AM |

jasmussen:
In principle I find myself most attracted to the no dashes option, with less noise. But if you feel strongly, the subdued dashes can work too.

ryelle:
> The first item in the chapter list is the name of the handbook and no longer gets a different visual treatment. Should we consider removing it (or removing the dash)? It's currently presented as a sibling of "Get Started" when in fact it's the parent.

That white background/bold text style is the current page indicator, when you're on a different page, there is no visual difference between "Block Editor Handbook" & "Getting Started":

Screenshot 2023-02-01 at 12 11 39 PM

> In principle I find myself most attracted to the no dashes option, with less noise

Should I make that change, then? It should be just as easy to do either. Personally, I like the subdued dashes better because the spacing looks a little too big with the empty spaces, but I'll leave the final decision to a designer :)

jasmussen:
I see your point, and we can go with the subdued dashes. 

It's a bit of a tricky interface, and it might benefit from a full actual tree view, which is that the dashes connect to the vertical line, which is harder for last-child situations on shortening the vertical line.

ryelle:
Latest block with the lighter dashes:

Screenshot 2023-02-01 at 5 17 59 PM

jasmussen:
Sorry for being nitpicky, are those dashes the same gray as the vertical dash? I see them as essentially an ersatz tree view so it'd be good if it was the same gray.

jasmussen:
Sorry for being nitpicky, are those dashes the same gray as the vertical dash? I see them as essentially an ersatz tree view so it'd be good if it was the same gray.

ryelle:
> Sorry for being nitpicky, are those dashes the same gray as the vertical dash? I see them as essentially an ersatz tree view so it'd be good if it was the same gray.

They technically are, but I think it looks lighter in the screenshot because the dash is 1.5px tall so there's some anti-aliasing going on. I've tweaked the svg a little to avoid the anti-alias, and TBH it still looks slightly lighter to me even though I can color-pick and see they're both #d9d9d9.

Screenshot 2023-02-02 at 11 57 17 AM

I think it's just an optical illusion since the line is shorter? But they are definitely the same now.

jasmussen:
We can try this, and revisit. I think the optimization is to do the full tree view look, but it's a less important aspect we can revisit and ticket if need be.

ryelle:
Sounds good, I'm going to merge this as-is now so we can move forward on this page layout.

I'll leave it up to design to decide if we need to change to a more-connected tree view, and you can open a new issue if so.

PR #190, opened 2023-02-01
Updating heading spacing to function single page.

StevenDufresne:
This PR adds margin and font size to the headings on a single page. The headings themselves exist in the respective blocks and are consistently applied across all of them. All the blocks are not complete but this could still probably merge ahead of that.

Lastly, the designs have a custom size of `26px` for the headings. I've bumped that up to `30px` so it stays consistent across .org. 

**Question**
Is this a good approach for updating the headings for a specific page?

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
+++ b/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
@@ -146,3 +146,8 @@ body[class] {
 	font-size: var(--wp--preset--font-size--small);
 	border-radius: 2px;
 }
+
+.single .entry-content section > h2 {
+	margin-top: var(--wp--preset--spacing--40);
+	font-size: var(--wp--preset--font-size--huge);
+}


outdoor2kode:
[Found](https://github.com/WordPress/wporg-parent-2021/blob/trunk/source/wp-content/themes/wporg-parent-2021/theme.json#L564) that it's `32px` for `--wp--preset--font-size--huge`, if you're looking for `30px`, `--wp--preset--font-size--heading-4` might be the right one? And I'm curious what `it stays consistent across .org.` means, is it that font-size of`h2` on other sites of .org is of size `30px`? I'm not so sure what it stays consistent with.

ryelle:
> Lastly, the designs have a custom size of 26px for the headings. I've bumped that up to 30px so it stays consistent across .org.

It looks like the design is using "heading 5 style" (though it switches to heading 3 on mobile, so that it stays 26px across both, which is not how the design system should work 😕). Could you just update the headers in the blocks to have the class `has-heading-5-font-size`?

For the margin, maybe updating the `blockGap` in the `<!-- wp:group {"tagName":"main",` wrapper? Does it need to be space on the header, or would space between the sections be enough?

StevenDufresne:
I have updated the headings in this trunk commit: https://github.com/WordPress/wporg-developer/commit/97d5f063833e3fc86ebf878c2d5d08137481e0ad.

ryelle:
@StevenDufresne These titles aren't run through `do_blocks` so those comments are also output in the page source:
Screenshot 2023-02-03 at 12 24 19 PM

PR #191, opened 2023-02-01
Remove display:block to fix wrong line breaks

### LINKED ISSUE #180: Related block: Incorrect line breaks in uses/used by descriptions
Screenshot 2023-01-27 at 3 03 42 PM

Source: http://localhost:8888/reference/functions/get_stylesheet/

It looks like this is because the `code` is `display: block`.

outdoor2kode:
Fixes #180 

Screenshots
| Src | Before | After |
| ----| -------| ----- |
| `reference/classes/theme_upgrader/current_after` | image |image |
| `reference/functions/get_raw_theme_root` | image | image |
| `reference/functions/get_site_transient` | image | image |
| `reference/functions/apply_filters` | image | image |

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -2744,5 +2744,5 @@
     "platform-overrides": {
         "php": "7.2"
     },
-    "plugin-api-version": "2.3.0"
+    "plugin-api-version": "2.2.0"
 }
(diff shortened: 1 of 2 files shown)


ryelle:
The left side of the table should still be in two lines - the function link on one line, and the file name below it on a separate line. For example, the mockup:



But in your screenshots they run into each other:

216103907-593d69af-840c-4f79-a13b-aaa78f65e189

outdoor2kode:
> The left side of the table should still be in two lines - the function link on one line, and the file name below it on a separate line. For example, the mockup:

Thanks for catching that, missed the left side!
A fix commit was pushed, and the screenshots were updated.

StevenDufresne:
LGTM. Merge when ready.

PR #192, opened 2023-02-01
Add autocomplete to homepage search bar

### LINKED ISSUE #184: Autocomplete on homepage search bar.
The search bar on the homepage should also have the autocomplete functionality that the search page has.

## Example
Screen Shot 2023-01-30 at 5 37 10 PM

outdoor2kode:
Fixes #184 

## Screenshot
autocomplete

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
+++ b/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
@@ -152,11 +152,17 @@ body[class] {
 	}
 }
 
-.home .entry-content {
-	--wp--preset--spacing--60: clamp(40px, 5.55556vw, 80px);
+.home {
+	.search-wrap div.awesomplete > ul {
+		top: 46px;
+	}
 
-	a {
-		text-decoration: underline;
+	.entry-content {
+		--wp--preset--spacing--60: clamp(40px, 5.55556vw, 80px);
+
+		a {
+			text-decoration: underline;
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


StevenDufresne:
This is working as expected, any reason why this was changed back to a draft?

outdoor2kode:
> This is working as expected, any reason why this was changed back to a draft?

I'm thinking to adjust the `top` of autocomplete box a bit so that it doesn't overlap the search bar. I saw that there's a gap like this for the search bar on the search page.

**Update**
The style was tweaked and the screenshot was updated.

source/wp-content/themes/wporg-developer-2023/src/style/style.scss
```diff
@@ -137,11 +137,17 @@ body[class] {
 	}
 }
 
-.home .entry-content {
-	--wp--preset--spacing--60: clamp(40px, 5.55556vw, 80px);
+.home {
+	.search-wrap div.awesomplete > ul {
+		top: 46px;
```
renintw, line 142: The search bar height on the search page is 38px, and it's 44px on the homepage.
So additional 6px is added to the top.
Or we can add `is-style-secondary-search-control` to the search bar on the homepage to make its height 38px.
But it would be a bit tricky to make its width 432px.

outdoor2kode, line 157: The search bar height on the search page is 38px, and it's 44px on the homepage.
So additional 6px is added to the top.
Alternatively, we can add `is-style-secondary-search-control` to the search bar on the homepage to make its height 38px, but it would be a bit tricky to make its width 432px.

PR #194, opened 2023-02-02
Add table of contents to single.html.

StevenDufresne:
This PR updates the `single.html` template to be similar to the handbooks, adding the sidebar and the table of contents block.

### Why does this filter `the content`?
Our table-of-contents control pulls the content to generate itself and associate anchors. However, since the code blocks are generated dynamically, as part of the template, we don't have any post content. Therefore the approach here is to `do_blocks` the content when we are viewing code.

It feels like there should be a better way of doing it, but I can't think of anything else. 🤔

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/functions.php
+++ b/source/wp-content/themes/wporg-developer-2023/functions.php
@@ -168,6 +168,10 @@ add_action( 'init', __NAMESPACE__ . '\\init' );
 add_filter( 'wporg_block_site_breadcrumbs', __NAMESPACE__ . '\set_site_breadcrumbs' );
 add_filter( 'single_template_hierarchy', __NAMESPACE__ . '\add_handbook_templates' );
 
+// Priority must be lower than 5 to precede table of contents filter.
+// See: https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/table-of-contents/index.php#L70
+add_filter( 'the_content', __NAMESPACE__ . '\filter_code_content', 4 );
+
 // Remove table of contents.
 add_filter( 'wporg_handbook_toc_should_add_toc', '__return_false' );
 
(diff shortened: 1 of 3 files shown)


source/wp-content/themes/wporg-developer-2023/functions.php
```diff
@@ -407,3 +408,32 @@ function set_site_breadcrumbs() {
...
+	$post = get_post();
+
+	if ( ! is_single() && ! is_parsed_post_type( $post->post_type ) ) {
+		return $content;
+	}
+
+	return do_blocks(
+		'
+		<!-- wp:wporg/code-reference-summary /-->
+		<!-- wp:wporg/code-reference-description /-->
```
ryelle, line 491: I noticed while testing https://github.com/WordPress/wporg-mu-plugins/pull/332 that this is causing a fatal error — since [`the_content` is used in this block](https://github.com/WordPress/wporg-developer/blob/aa3dc20cc962e50ced78cbca85462f4eb64987f9/source/wp-content/themes/wporg-developer-2023/inc/template-tags.php#L1728), you get stuck in an infinite loop trying to render it.

ryelle:
> It feels like there should be a better way of doing it, but I can't think of anything else.

I wonder if we could put that list of blocks somewhere more expected, so that in the future we're not confused about how this set of blocks is built?

Aside from the fatal error, this doesn't seem like a bad approach, and it has the bonus of making it consistent with the other pages that display using the post-content block.

StevenDufresne:
> > It feels like there should be a better way of doing it, but I can't think of anything else.
> 
> I wonder if we could put that list of blocks somewhere more expected, so that in the future we're not confused about how this set of blocks is built?

The only thing I can think of is a template part (or PHP pattern) that is loaded as a string and parsed. 

I guess a dynamic block of blocks would also work but the extra asset loading seems a bit wasteful.

source/wp-content/themes/wporg-developer-2023/patterns/single.php
```diff
@@ -7,27 +7,40 @@
...
-<!-- wp:wporg/code-reference-changelog /-->
-<!-- wp:wporg/code-reference-comments /--></main>
-
-<!-- wp:group {"tagName":"div","style":{"spacing":{"padding":{"top":"var:preset|spacing|40","bottom":"var:preset|spacing|40", "left":"var:preset|spacing|edge-space","right":"var:preset|spacing|edge-space"}}}} -->
-<div class="wp-block-group alignfull" style="padding-top:var(--wp--preset--spacing--40);padding-bottom:var(--wp--preset--spacing--40);padding-left:var(--wp--preset--spacing--edge-space);padding-right:var(--wp--preset--spacing--edge-space)">
-	<!-- wp:wporg/code-reference-comment-form /-->
-</div>
-<!-- /wp:group -->
-<!-- /wp:group -->
+<!-- wp:group {"tagName":"main","layout":{"type":"constrained"},"style":{"spacing":{"padding":{"left":"var:preset|spacing|edge-space","right":"var:preset|spacing|edge-space"}}}} -->
```
StevenDufresne: I'll probably reverse these changes and rebase once we agree on the approach: https://github.com/WordPress/wporg-developer/pull/194#issuecomment-1414521184

ryelle:
> The only thing I can think of is a template part (or PHP pattern) that is loaded as a string and parsed.

Hm, but then it's unexpected that it doesn't render directly. What about this approach, but leaving a PHP comment in `patterns/single.php` explaining that the content is built from `filter_code_content` in `functions.php`?

PR #193, opened 2023-02-01
Update edit & changelog links on handbook pages

ryelle:
See #161 — In #169 the page template was added with placeholder content for the edit & changelog links below the handbook content. This PR adds in two new shortcodes to generate those URLs for handbooks that live on github.

For the other handbooks, non-WP users can't edit them, and they don't have changelogs, so they don't need these links. I've split the pattern into `single-handbook.php` and `single-github-handbook.php`, and show the edit/changelog links on `single-github-handbook.php`.

The handbooks:

- Common APIs
- Plugin
- Theme
- Block Editor — github
- Coding Standards — github
- REST API — github
- Advanced Admin (WIP) — github

| github handbook | non-gh handbook |
|-----|-----|
| Screenshot 2023-02-01 at 5 43 08 PM | Screenshot 2023-02-01 at 5 43 37 PM |

**To test**

Has links:
- Load a handbook page from one of the GH handbooks, like Block Editor
- The end of the article should have two links
- Clicking Edit will either take you to a edit screen on github or a button to fork the repo, depending on your permissions
- Clicking Changelog will take you to the commit history for that markdown file

No links:
- Load a page from one of the other handbooks (I manually exported some content from prod using the WP importer)
- There should be no edit or changelog link

PR #196, opened 2023-02-02
Split patterns into "partials", apply layout to all single pages

ryelle:
See #161, #162, https://github.com/WordPress/wporg-developer/pull/193/files#r1093866881.

I've added a filter to the template hierarchy here so that we can just create two generic templates for the handbook pages, rather than duplicating one for each post type. Along with that, I've moved most of the blocks from the `single-*` patterns, and instead created some pattern "partials" — the parts of the template that need translation and are repeated between layouts. Finally, I updated the general `single.html` to use the layout & these new patterns.

| Handbook (non-github) | Handbook (github) | Reference page |
|-----|-----|-----|
| handbook | gh-handbook | reference |


ryelle:
Forgot to mention, I used the `article` term for the patterns from documentation, but we could also rename these to "post-sidebar" etc.

source/wp-content/themes/wporg-developer-2023/templates/single.html
```diff
@@ -1,5 +1,44 @@
 <!-- wp:template-part {"slug":"header","className":"has-display-contents"} /-->
 
-<!-- wp:pattern {"slug":"wporg-developer-2023/single-content"} /-->
+<!-- wp:group {"tagName":"main","layout":{"type":"constrained","wideSize":"1280px","contentSize":"680px"},"style":{"spacing":{"padding":{"left":"var:preset|spacing|edge-space","right":"var:preset|spacing|edge-space"}}}} -->
```
StevenDufresne, line 3: Can you add in your [suggestion here](https://github.com/WordPress/wporg-developer/pull/190#issuecomment-1414364527) for block spacing.

StevenDufresne (approved):

PR #197, opened 2023-02-02
Add post navigation links to handbooks

ryelle:
See #161 — This adds the post navigation links to the bottom of handbooks. It needs to use the handbook hierarchy for determining which is next/previous, so I had to filter those links to use the correct post. The links should now reflect the next or previous item in the Chapter List sidebar.

On the production site, this navigates through the items slightly differently — the previous item will take you to the previous section at the same level you're on, not the previous item in the list. For example, the previous link on [Design Contributions](https://developer.wordpress.org/block-editor/contributors/design/) is "Code Contributions", but on this PR it's "How To Get Your Pull Request Reviewed". I can't get that working right now, but I might just need fresh eyes.

| block-editor | coding-standards | plugins | rest-api | 
|----|----|----|---|
|  |  |  |  |

**To test**

- Navigate through a few different handbook pages
- The next/prev buttons should work
- The path they take you should match the Chapter List order
- The first item should have no "previous", and the last item should have no "Next"

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/functions.php
+++ b/source/wp-content/themes/wporg-developer-2023/functions.php
@@ -518,10 +518,12 @@ function filter_code_content( $content ) {
  * @return string Updated link tag.
  */
 function get_adjacent_handbook_post_link( $output, $format, $link, $post, $adjacent ) {
-	$post_id   = get_the_ID();
-	$parent    = get_post_parent( $post_id );
-	$parent_id = $parent ? $parent->ID : 0;
-	$pages     = get_pages(
+	if ( wporg_is_handbook() ) {
+		return $output;
+	}
+
+	$post_id = get_the_ID();
+	$pages   = get_pages(
 		array(
 			'sort_column' => 'menu_order, title',
 			'post_type'   => get_post_type( $post_id ),
--- a/source/wp-content/themes/wporg-developer-2023/functions.php
+++ b/source/wp-content/themes/wporg-developer-2023/functions.php
@@ -518,7 +518,7 @@ function filter_code_content( $content ) {
  * @return string Updated link tag.
  */
 function get_adjacent_handbook_post_link( $output, $format, $link, $post, $adjacent ) {
-	if ( wporg_is_handbook() ) {
+	if ( ! wporg_is_handbook() ) {
 		return $output;
 	}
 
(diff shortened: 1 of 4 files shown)


StevenDufresne:
On: http://localhost:8888/block-editor/getting-started/devenv/,

The next page is [ColorPalette](http://localhost:8888/block-editor/reference-guides/components/color-palette/). That is unexpected. Is there something wrong with my environment?

source/wp-content/themes/wporg-developer-2023/functions.php
```diff
@@ -297,8 +299,8 @@ function pre_get_posts( $query ) {
 function register_nav_menus() {
```
StevenDufresne, line 344: I think this whole function is dead now.

ryelle:
> The next page is [ColorPalette](http://localhost:8888/block-editor/reference-guides/components/color-palette/). That is unexpected. Is there something wrong with my environment?

No… 🤦🏻  I introduced a logic error in https://github.com/WordPress/wporg-developer/pull/197/commits/5858762f32476e649912c479b33ae5a074a48b33 — that should be `if ( ! wporg_is_handbook() ) …`. Fixed in https://github.com/WordPress/wporg-developer/pull/197/commits/3a43effe6cc4f2d39bddfec528ed29cb7d11f675.

source/wp-content/themes/wporg-developer-2023/functions.php
ryelle: Probably, but that's actually not related to this change:

- `devhub-cli-menu` is the nav in the blue banner on [the WP-CLI handbook](https://developer.wordpress.org/cli/commands/)
- `devhub-menu` is the nav in the blue banner everywhere else, but it's empty
- `reference-home-api` is the API column on [the reference page](https://developer.wordpress.org/reference/)

The 3 handbooks that don't use github do use nav menus on production for their ToC — I mentioned that in https://github.com/WordPress/wporg-developer/pull/189 but didn't implement that split behavior. 

On production, they're just unassigned menus used in a widget.
Screenshot 2023-02-03 at 1 17 26 PM

All that is to say, I'd rather not remove this as part of this PR, but we can remove it in general.

StevenDufresne (approved):
Yep, looks good.

Issue #185, opened 2023-01-30
Archives: Function/Class/Method/Hook list page.


StevenDufresne:
We need to add a template to handle the archives for:
 - https://developer.wordpress.org/reference/functions/
 - https://developer.wordpress.org/reference/classes/
 - https://developer.wordpress.org/reference/methods/
 - https://developer.wordpress.org/reference/hooks/

The archive also allows the filtering by file and some other dimensions:
- Since: https://developer.wordpress.org/reference/since/6.1.0/
- File: https://developer.wordpress.org/reference/files/wp-includes/user.php/

PR #201, opened 2023-02-06
Parsed Post Type Archive

StevenDufresne:
See  https://github.com/WordPress/wporg-developer/pull/201#issuecomment-1420117373 for implementation.
<hr>

Related #185 

There are some archive pages that show a specific post type (ie: all functions). Users get there when they click on the post type in the breadcrumb: `Developer \ Functions \ absint `.

We don't have a design for these pages specifically. 

This is what is looks in [production](https://developer.wordpress.org/reference/functions/):

We no longer have the subnav search bar or the breadcrumb in the content. As a result, it's looks pretty sparse.

**The other examples:**

 | method | class | hook |
| --- | --- | --- | 
| localhost_8888_reference_methods_ | localhost_8888_reference_hooks_ | localhost_8888_reference_classes_ | 

# Help

@WordPress/meta-design Any thoughts here? Should we make it look more like search but without the filter checkboxes?

### Links
- [figma](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-Resources?node-id=1160%3A10164&t=oU0nxF52BZkGOBc3-0)

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/functions.php
+++ b/source/wp-content/themes/wporg-developer-2023/functions.php
@@ -166,7 +166,7 @@ require_once __DIR__ . '/src/code-comment-edit/block.php';
 require_once __DIR__ . '/src/code-comment-form/block.php';
 require_once __DIR__ . '/src/search-filters/index.php';
 require_once __DIR__ . '/src/search-results-context/index.php';
-require_once __DIR__ . '/src/post-title/index.php';
+require_once __DIR__ . '/src/code-short-title/index.php';
 require_once __DIR__ . '/src/post-type-usage-info/index.php';
 require_once __DIR__ . '/src/version-select/index.php';
 
--- a/source/wp-content/themes/wporg-developer-2023/functions.php
+++ b/source/wp-content/themes/wporg-developer-2023/functions.php
@@ -166,8 +166,8 @@ require_once __DIR__ . '/src/code-comment-edit/block.php';
 require_once __DIR__ . '/src/code-comment-form/block.php';
 require_once __DIR__ . '/src/search-filters/index.php';
 require_once __DIR__ . '/src/search-results-context/index.php';
-require_once __DIR__ . '/src/search-title/index.php';
-require_once __DIR__ . '/src/search-usage-info/index.php';
+require_once __DIR__ . '/src/post-title/index.php';
+require_once __DIR__ . '/src/post-type-usage-info/index.php';
 require_once __DIR__ . '/src/version-select/index.php';
 
 add_action( 'init', __NAMESPACE__ . '\\init' );
--- a/source/wp-content/themes/wporg-developer-2023/functions.php
+++ b/source/wp-content/themes/wporg-developer-2023/functions.php
@@ -158,16 +158,16 @@ require_once __DIR__ . '/src/code-private-access/block.php';
 require_once __DIR__ . '/src/code-related/block.php';
 require_once __DIR__ . '/src/code-return-value/block.php';
 require_once __DIR__ . '/src/code-source/block.php';
+require_once __DIR__ . '/src/code-short-title/index.php';
 require_once __DIR__ . '/src/code-summary/block.php';
 require_once __DIR__ . '/src/code-table/block.php';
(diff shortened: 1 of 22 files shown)


jasmussen:
What would be your preference?

We can make it look more like search if that's useful, yes. What's the hierarchy as far as breadcrumbs go? Can the same breadcrumbs not be applied in the new design?

ryelle:
I wonder if all we really need is some kind of archive title?

| This PR | Plus archive title | 
|---|---|
| none-hooks | title-hooks |
| none-file | title-file |

All I did was drop this archive title code into the template:

```html
<!-- wp:group {"style":{"spacing":{"margin":{"bottom":"var:preset|spacing|40"}}},"align":"wide"} -->
<div class="wp-block-group alignwide" style="margin-bottom:var(--wp--preset--spacing--40)">
    <!-- wp:query-title {"type":"archive"} /-->
</div>
<!-- /wp:group -->
```

It would also give a page like http://localhost:8888/reference/files/wp-includes/functions.php/ (or [on production](https://developer.wordpress.org/reference/files/files/wp-includes/functions.php/)) a place to put that filter dropdown.

StevenDufresne:
> I wonder if all we really need is some kind of archive title?

Yep, that works for me. We can come back here later if necessary. This archive view is not critical.

StevenDufresne:
I have updated the template to include the archive type.

## Since


**No pagination**




## Parsed post types






## Reference

jasmussen:
👍 👍

source/wp-content/themes/wporg-developer-2023/functions.php
```diff
@@ -161,8 +161,8 @@
 require_once __DIR__ . '/src/code-comment-form/block.php';
 require_once __DIR__ . '/src/search-filters/index.php';
 require_once __DIR__ . '/src/search-results-context/index.php';
-require_once __DIR__ . '/src/search-title/index.php';
-require_once __DIR__ . '/src/search-usage-info/index.php';
+require_once __DIR__ . '/src/post-title/index.php';
+require_once __DIR__ . '/src/post-type-usage-info/index.php';
```
ryelle: Is there a single term for all the reference post types? `post` seems too generic, since these blocks wouldn't be useful on the handbook pages.

This doesn't have to be a blocker if there isn't a better term, but since you're already renaming it 🙂

source/wp-content/themes/wporg-developer-2023/templates/archive.html
```diff
@@ -1,3 +1,45 @@
...
+	<!-- wp:group {"style":{"spacing":{"margin":{"bottom":"var:preset|spacing|40"}}},"align":"wide"} -->
+	<div class="wp-block-group alignwide" style="margin-bottom:var(--wp--preset--spacing--40)">
+		<!-- wp:query-title {"type":"archive"} /-->
+	</div>
+	<!-- /wp:group -->
+
+	<!-- wp:group {"className":"align-left","layout":{"type":"constrained","contentSize":"","justifyContent":"left"}} -->
+	<div class="wp-block-group align-left">
+
+	<!-- wp:wporg/search-results-context {"style":{"spacing":{"padding":{"top":"var:preset|spacing|20"}}}} /-->
```
ryelle: I think you can remove this, it's not returning anything for these archive views.

ryelle (commented):
Looks good! I see you marked this as fixing 185, but it doesn't include the extra filters on the file & since archives.

source/wp-content/themes/wporg-developer-2023/functions.php
StevenDufresne: Yeah :) I struggled here. The other option which matches legacy code is `parsed-post-type-title`

ryelle: hmmmm… maybe `code` to match the other blocks? I know `code-title` is taken, but maybe `code-short-title`? Or refactoring `code-title` to show either with an attribute toggle? That's definitely more involved for a minor devex improvement.

StevenDufresne:
I have updated the block names and merged them. Any additional changes can be done in a follow-up pr!

PR #203, opened 2023-02-06
Block: Add a version dropdown.

StevenDufresne:
Related: #201

We need a dropdown for the page that lists changes. Once we have the archive template setup in #201, we can use this control for the following design:

## Considerations
I'm not sure if the select HTML should be generalized and moved to the parent or a mu-plugin but for now, I have it here.

## Screenshot
 | Default |  Font size: 12px | Font size: 24px |
| --- | --- |  --- | 
| Screen Shot 2023-02-07 at 1 10 17 PM | Screen Shot 2023-02-07 at 1 10 23 PM | Screen Shot 2023-02-07 at 1 10 30 PM |  

| Truncated |
| --- | 
| Screen Shot 2023-02-07 at 1 08 47 PM |  

## How To Test
1. Add `<!-- wp:wporg/version-select /-->` to any template.

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/version-select/index.php
+++ b/source/wp-content/themes/wporg-developer-2023/src/version-select/index.php
@@ -36,10 +36,6 @@ function init() {
  * @return string Returns the block markup.
  */
 function render( $attributes, $content, $block ) {
-	if ( ! isset( $block->context['postId'] ) ) {
-		return '';
-	}
-
 	$versions = get_terms( 'wp-parser-since', array( 'order' => 'DESC' ) );
 
 	if ( is_wp_error( $versions ) || empty( $versions ) ) {
@@ -68,11 +64,22 @@ function render( $attributes, $content, $block ) {
 	$wrapper_attributes = get_block_wrapper_attributes();
 	return sprintf(
 		'<div %1$s>
-			<label class="screen-reader-text" for="version-select">%2$s</label>
-			<select name="version-select" id="version-select">%3$s</select>
+			<label class="screen-reader-text" for="%2$s">%3$s</label>
+			<select name="%2$s" id="%2$s">%4$s</select>
 		</div>',
 		$wrapper_attributes,
+		esc_attr( generate_id( $block->parsed_block ) ),
 		esc_html( __( 'Select version', 'wporg' ) ),
 		$options,
 	);
+}
+
+/**
+ * Generates a unique identifier.
+ *
+ * @param array $block Block object.
+ * @return string      The unique identifier.
+ */
+function generate_id( $block ) {
+	return 'wporg-version-select-' . md5( serialize( $block ) );
 }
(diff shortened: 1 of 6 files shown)


ryelle:
Can you use a regular html `select`, or does it need different features?

StevenDufresne:
> Can you use a regular html `select`, or does it need different features?

Right sorry, do we have a block or have we styled a dropdown?

ryelle:
> Right sorry, do we have a block or have we styled a dropdown?

As far as I know we haven't had a dropdown in the redesigned sites yet.

source/wp-content/themes/wporg-developer-2023/src/version-select/index.php
```diff
@@ -0,0 +1,78 @@
...
+	foreach ( array_merge( $versions, $mu_versions ) as $version ) {
+		$options .= '<option value="' . $version->name . '">' . $version->name . '</option>';
+	}
+
+	$wrapper_attributes = get_block_wrapper_attributes();
+	return sprintf(
+		'<div %1$s>
+			<label class="screen-reader-text" for="version-select">%2$s</label>
+			<select name="version-select" id="version-select">%3$s</select>
+		</div>',
```
ryelle: Do you think it's possible this would get used twice on one page? If so, these should be unique IDs per block. GB creates a unique class name [by `md5`ing the block](https://github.com/WordPress/gutenberg/blob/88335dbf42f456b2c89e3fa12ade3d5170100daa/lib/block-supports/elements.php#L8-L16).

source/wp-content/themes/wporg-developer-2023/src/version-select/index.php
```diff
@@ -0,0 +1,78 @@
...
+ * @param array    $attributes Block attributes.
+ * @param string   $content    Block default content.
+ * @param WP_Block $block      Block instance.
+ *
+ * @return string Returns the block markup.
+ */
+function render( $attributes, $content, $block ) {
+	if ( ! isset( $block->context['postId'] ) ) {
+		return '';
+	}
```
ryelle: I don't think this block needs a post context, does it?

ryelle (commented):
This does what it says, adds a select to the page. It won't do anything without a form wrapper though, so we'll need to build that up too (plus the other blocks for the type filter & apply button). 

I was really hoping form blocks would come to GB faster 😆

source/wp-content/themes/wporg-developer-2023/src/version-select/index.php
StevenDufresne: Awesome! I'll generate a unique id as well. Thanks for the context.

source/wp-content/themes/wporg-developer-2023/src/version-select/index.php
StevenDufresne: Nope.

ryelle (approved):
Looks good, and the preview in the editor is nice.

I did notice that the block can get fully collapsed, what do you think about adding a min-width to the select? Since most of the values are all the same width, and pretty small.

Firefox
Screenshot 2023-02-08 at 4 19 38 PM

Safari (it's smaller because I'd reset the font size)
Screenshot 2023-02-08 at 4 16 36 PM

approving, though it would be nice to address ^ before merge.

PR #210, opened 2023-02-09
Code refs: Use the new custom filter in the Table of Contents block

ryelle:
Relies on https://github.com/WordPress/wporg-mu-plugins/pull/334. That PR introduces a new filter for the heading-parsing, so we can simply add `filter_code_content` to that filter too.

The `filter_code_content` on `the_content` is still necessary for the main content replacement, and it still needs to run earlier than 5 so that the content is in place before the heading replacement is run on priority 5.

**To test**

Using both PRs, check that the Table of Contents still works on handbook pages & code references.


StevenDufresne (approved):
Tested. Works!

Issue #183, opened 2023-01-30
Fix searches from handbooks.


StevenDufresne:
If you search using the search control on the handbook template, you are taken to functions search results. It should search relative to the handbooks. 

For example :`/?s=keyword` -> `/blocks-editor/?s=keyword`.

As it sits now, if you visit  `/blocks-editor/?s=keyword`, the search results template will pick it up and it will look like:

<img width="400" alt="Screen Shot 2023-01-30 at 5 34 51 PM" src="https://user-images.githubusercontent.com/1657336/215426998-71c57edd-4499-41f8-8ca2-5c4550903459.png">

There isn't currently a design for those results although we can probably just remove the grey tab and then request feedback from design.

PR #211, opened 2023-02-10
Add WP CLI Interior Page

### LINKED ISSUE #160: Layout: WP CLI Interior page
## Screenshot

[Figma Link](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-Pages?node-id=1160%3A10487&t=pS0XoTexhVvCl9Ft-0)

image

outdoor2kode:
Fixes #160 

#### Commands with different content layouts used for testing.
- cli/commands/cache/
- cli/commands/find/
- cli/commands/eval/
- cli/commands/export/
- cli/commands/admin/

## Screenshots
**Desktop I**

**Desktop II**
image

**Mobile**
image


source/wp-content/themes/wporg-developer-2023/templates/single-command.html
```diff
@@ -0,0 +1,34 @@
...
+<!-- wp:group {"tagName":"main","layout":{"type":"constrained","wideSize":"1280px","contentSize":"680px"},"style":{"spacing":{"padding":{"left":"var:preset|spacing|edge-space","right":"var:preset|spacing|edge-space"}}}} -->
+<main class="wp-block-group alignfull" style="padding-left:var(--wp--preset--spacing--edge-space);padding-right:var(--wp--preset--spacing--edge-space)">
+
+	<!-- wp:group {"layout":{"type":"constrained","justifyContent":"left"},"align":"wide","style":{"spacing":{"margin":{"top":"var:preset|spacing|edge-space"}}}} -->
+	<div class="wp-block-group alignwide" style="margin-top:var(--wp--preset--spacing--edge-space)">
+
+		<!-- wp:group {"tagName":"article","style":{"spacing":{"blockGap":"var:preset|spacing|40","margin":{"top":"0px"}}}} -->
+		<article class="wp-block-group" style="margin-top:0px">
+
+			<!-- wp:wporg/code-reference-title {"level":1} /-->
```
StevenDufresne: This doesn't match the designs. Should we be using the same block or a new one? Or styled default Gutenberg blocks? We also need to handle the `<command>` portion:

Screen Shot 2023-02-13 at 10 45 35 AM

StevenDufresne: What about the GitHub buttons?

StevenDufresne:
I created a branch that uses the wporg-code-table to illustrate what we discussed earlier. 
 
https://github.com/WordPress/wporg-developer/compare/add/cli-shared-tables?expand=1

source/wp-content/themes/wporg-developer-2023/functions.php
```diff
@@ -492,9 +494,9 @@ function add_handbook_templates( $templates ) {
  * @return string
  */
 function filter_code_content( $content ) {
-	$post = get_post();
+	$post_type = get_post_type();
 
-	if ( ! is_single() || ! is_parsed_post_type( $post->post_type ) ) {
```
StevenDufresne: These changes should probably not be done in this branch.

outdoor2kode: Originally I changed it because I used an extra `$post_type` below (not anymore for now) and I saw this PHP warning locally somehow:
image
After applying this change, the error was gone. Not sure if you saw the warning as well on your side.

source/wp-content/themes/wporg-developer-2023/templates/single-command.html
outdoor2kode: > This doesn't match the designs. Should we be using the same block or a new one

I added a new one - [Block: Command Title](https://github.com/WordPress/wporg-developer/pull/211/commits/f0ce3a8f0e49adaa7a09c51ebac30710b27cfc65).

> What about the GitHub buttons?

[Block: Command Github](https://github.com/WordPress/wporg-developer/pull/211/commits/5c0d65765220d83b6940907769059089681b30c1)

source/wp-content/themes/wporg-developer-2023/functions.php
StevenDufresne: I think it's the right change but we should make it in trunk and rebase.

source/wp-content/themes/wporg-developer-2023/src/command-github/block.php
```diff
@@ -0,0 +1,133 @@
...
+				</a>
+			',
+			esc_url( $repo_url )
+		);
+	}
+
+	$content .= sprintf(
+		'
+		<div class="btn-group">
+			<a href="%s" class="button">View Open Issues
```
StevenDufresne: This should be localized.

source/wp-content/themes/wporg-developer-2023/src/command-title/block.php
```diff
@@ -0,0 +1,72 @@
...
+		array(
+			'post_parent'    => $post_ID,
+			'post_type'      => 'command',
+			'posts_per_page' => 1,
+			'orderby'        => 'title',
+			'order'          => 'ASC',
+		)
+	);
+
+	$content = sprintf(
```
StevenDufresne, line 45: This should be an `<h1>`

source/wp-content/themes/wporg-developer-2023/src/command-content/block.php
```diff
@@ -0,0 +1,108 @@
...
+			$wrapper_attributes,
+			do_blocks( $content ),
+		);
+	}
+
+	$repo_slug = str_replace( 'https://github.com/', '', $repo_url );
+	$command = get_the_title();
+	$installing_instructions = sprintf(
+		'
+		<!-- wp:heading {"level":3} --><h3 class="wp-block-heading">%1$s</h3><!-- /wp:heading -->
```
StevenDufresne: Let's make all the headings match the single template. [Example](https://github.com/WordPress/wporg-developer/blob/2ce6c3f0fa837d86e41c60544511051748d9a9fa/source/wp-content/themes/wporg-developer-2023/src/code-comments/block.php#L52)

StevenDufresne:
##  The headings 

- For headings that are coming from the content, how do we make sure they are at the correct level? 
- We control the spacing between the blocks with a blockGap but since the embedded content isn't at the same level, the spacing between the sections seems inconsistent.

## Uppercase Titles

We need these to be camel case. What do you think is the best way to do so?

**Example:**
Screen Shot 2023-02-14 at 10 26 12 AM

## Responsive 

This table should be responsive. [See design.](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-Resources?node-id=2014%3A19479&t=mycfroYAJWmt99bc-0)

Screen Shot 2023-02-14 at 10 27 57 AM

source/wp-content/themes/wporg-developer-2023/src/code-table/style.scss
```diff
@@ -54,6 +54,41 @@
 	}
 }
 
+&.is-responsive {
```
ryelle: I wasn't able to build this branch, I got the following error:

```
SassError: Top-level selectors may not contain the parent selector "&".
   ╷
57 │ &.is-responsive {
   │ ^^^^^^^^^^^^^^^^
   ╵
```

source/wp-content/themes/wporg-developer-2023/src/command-subcommand/block.php
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Render the block content.
+ *
+ * @return string Returns the block markup.
+ */
+function render( $attributes, $content, $block ) {
+	$children = get_children(
+		array(
+			'post_parent'    => get_the_ID(),
```
ryelle: This should be using the ID from the block context, like this:

https://github.com/WordPress/wporg-developer/blob/0216fb6063904af07315538f5ae15826f865f4c8/source/wp-content/themes/wporg-developer-2023/src/code-deprecated/block.php#L24-L38

source/wp-content/themes/wporg-developer-2023/src/command-content/block.php
```diff
@@ -0,0 +1,108 @@
...
+ * Render the block content.
+ *
+ * @return string Returns the block markup.
+ */
+function render( $attributes, $content, $block ) {
+
+	$content = filter_the_content( get_the_content() );
+
+	$wrapper_attributes = get_block_wrapper_attributes();
+	$repo_url = get_post_meta( get_the_ID(), 'repo_url', true );
```
ryelle: This should also use the ID from block context.

source/wp-content/themes/wporg-developer-2023/src/command-content/block.php
```diff
@@ -0,0 +1,108 @@
...
+}
+
+/**
+ * Render the block content.
+ *
+ * @return string Returns the block markup.
+ */
+function render( $attributes, $content, $block ) {
+
+	$content = filter_the_content( get_the_content() );
```
ryelle: All these global calls (in `filter_the_content` too) should be using the post from block context. You can either pass an ID to [get_the_content](https://developer.wordpress.org/reference/functions/get_the_content/) or get the post object and use `$post->post_content`.

source/wp-content/themes/wporg-developer-2023/functions.php
```diff
@@ -181,6 +185,8 @@
 // See: https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/table-of-contents/index.php#L70
 add_filter( 'the_content', __NAMESPACE__ . '\filter_code_content', 4 );
 add_filter( 'wporg_table_of_contents_post_content', __NAMESPACE__ . '\filter_code_content' );
+add_filter( 'the_content', __NAMESPACE__ . '\filter_command_content', 4 );
+
```
ryelle, line 191: Add the filter to the table of contents hook too so that the content shows up in the table of contents.

```suggestion
add_filter( 'wporg_table_of_contents_post_content', __NAMESPACE__ . '\filter_command_content' );
```

source/wp-content/themes/wporg-developer-2023/src/command-content/block.php
ryelle: `get_the_content` doesn't run the `the_content` filter, so I think `filter_the_content` is unnecessary here— since it's meant to undo some of the special character magic WordPress does on `the_content`. I'm not 100% sure about that, though.

source/wp-content/themes/wporg-developer-2023/src/command-content/block.php
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	$repo_slug = str_replace( 'https://github.com/', '', $repo_url );
+	$command = get_the_title();
+	$installing_instructions = sprintf(
+		'
+		<!-- wp:heading {"level":3} --><h3 class="wp-block-heading">%1$s</h3><!-- /wp:heading -->
+		<p>Use the <code>%2$s</code> command by installing the command\'s package:</p>
+		<pre><code>wp package install %3$s</code></pre>
+		<p>Once the package is successfully installed, the <code>%4$s</code> command will appear in the list of available commands.</p>
```
ryelle: Should this also be translatable?

source/wp-content/themes/wporg-developer-2023/src/command-github/style.scss
```diff
@@ -0,0 +1,56 @@
...
+					background: #eee;
+				}
+
+				.green {
+					color: #46b450;
+				}
+
+				.red {
+					color: #dc3232;
+				}
```
ryelle: Could these be updated to use the syntax colors? Technically they still fail color contrast on `#f7f7f7`, but it's better than these colors.

```suggestion
				.green {
					color: var(--wp--preset--color--syntax-green);
				}

				.red {
					color: var(--wp--preset--color--syntax-red);
				}
```

In general, I'm surprised this old button style is still being used, I would have expected some kind of update to the new button style (no change requested for button styles, since this does match the mockup).

source/wp-content/themes/wporg-developer-2023/src/code-table/style.scss
outdoor2kode: Ah, thanks for catching this! Fixed.

source/wp-content/themes/wporg-developer-2023/src/command-content/block.php
outdoor2kode: Not sure if the changes [in this commit](https://github.com/WordPress/wporg-developer/pull/211/commits/9191e5d9f574ba126820b66d526eac91605acb66) match what you were suggesting. Correct me if I misunderstood.

source/wp-content/themes/wporg-developer-2023/src/command-github/style.scss
outdoor2kode: Updated.
I was surprised by that too.
It's a bit inconsistent across the sites, but it doesn't look that outdated or out of tune overall to me though.

source/wp-content/themes/wporg-developer-2023/src/command-content/block.php
outdoor2kode: > get_the_content doesn't run the the_content filter, so I think filter_the_content is unnecessary here— since it's meant to undo some of the special character magic WordPress does on the_content. I'm not 100% sure about that, though.

TIL. I'll remove it first, I don't see anything go weird after removing it.

outdoor2kode: Possibly a really basic question. For not calling global calls here, is it because blocks sometimes would be used in the env that has no global variables? Is there any example of it?

ryelle: Yeah, the blocks could be used in a Query Loop on another page — so the global functions would refer to the page itself, not the current item in the query loop. GB has some protections in place to avoid that, but I've run into issues related to that (like [this one](https://github.com/WordPress/pattern-directory/pull/553)).

We're not doing anything like that on this project at the moment, but I see it as future-proofing things.

source/wp-content/themes/wporg-developer-2023/src/command-content/block.php
outdoor2kode: Who would be the best person to ask about this?
I'm tempted to translate it, but given that there is a big part of content coming from post content which is in en at the moment, I'm not so sure if we still want to translate it.

source/wp-content/themes/wporg-developer-2023/src/command-content/block.php
outdoor2kode: Thanks for the clarification!

outdoor2kode:
> ## Uppercase Titles
>
> We need these to be camel case. What do you think is the best way to do so?

This CSS could work only for a single word but multiple words - `OPTIONS -> Options`. But `GLOBAL PARAMETERS -> Global parameters`.
```
a {
    text-transform: lowercase;

    &::first-letter {
        text-transform: uppercase;
    }
}
```

If this is not acceptable, we probably need to use some regex here.
I'm tempted to either leverage the CSS way and ask for feedback from the design team or just directly provide feedback to them that some content was added in the editor.

> ## Responsive 
>
> This table should be responsive. [See design.](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-> > Resources?node-id=2014%3A19479&t=mycfroYAJWmt99bc-0)

Fixed.

outdoor2kode:
> ##  The headings 
>
> - For headings that are coming from the content, how do we make sure they are at the correct level? 

Probably some regex needed as well? I haven't seen any incorrect at the moment, if it's a rare case, perhaps leaving it to users to adjust it once it's spotted is acceptable?

> - We control the spacing between the blocks with a blockGap but since the embedded content isn't at the same level, the spacing between the sections seems inconsistent.

I saw there's `20px` margin on the top for every section, so I tried adding the same margin-top to `<h3>`. The result seems a bit full. The initial margin-top of `<h3>` is `30px`. Which command page you were referring to?

#### Result Screenshot 
image

source/wp-content/themes/wporg-developer-2023/src/command-content/block.php
StevenDufresne: Which commit? That link isn't working.

StevenDufresne:
> > ## Uppercase Titles
> > We need these to be camel case. What do you think is the best way to do so?
> 
> This CSS could work only for a single word but multiple words - `OPTIONS -> Options`. But `GLOBAL PARAMETERS -> Global parameters`.
> 
> ```
> a {
>     text-transform: lowercase;
> 
>     &::first-letter {
>         text-transform: uppercase;
>     }
> }
> ```
> 
> If this is not acceptable, we probably need to use some regex here. I'm tempted to either leverage the CSS way and ask for feedback from the design team or just directly provide feedback to them that some content was added in the editor.

Using a CSS approach would mean that the table of contents would still have uppercase titles, right? I think we'll need to update the content. We can look at the importer or alternatively filter them in code.

source/wp-content/themes/wporg-developer-2023/src/command-github/block.php
```diff
@@ -0,0 +1,143 @@
...
+			<div class="github-tracker">
+				<a href="%s">
+					<img src="https://make.wordpress.org/cli/wp-content/plugins/wporg-cli/assets/images/github-mark.svg" class="icon-github" />
+				</a>
+			',
+			esc_url( $repo_url )
+		);
+	}
+
+	$content .= sprintf(
```
StevenDufresne, line 53: I find this section hard to read because some of the tags are closed in the `sprintf` and some aren't. It would be pretty easy to make a mistake and lose the closing `</a` line. 

Could we maybe separate all the `<a>` elements into their own `sprintf` and then just add the containers to the template that is returned:

```
$issue_count = "";
if ( false !== $issue_count['open'] ) {
		$issue_count = sprintf(
			' <span class="green">(%s)</span>',
			(int) $issue_count['open']
		);
}

$content .= sprintf(
		'<a href="%1$s"class="button">%2$s %3$s</a>',
		esc_url( $closed_issues ),
		__( 'View Closed Issues', 'wporg' ),
                 $open_issue,
);
```
...
```
return sprintf(
		'<section %1$s><div class="btn-group">%2$s</div></section>',
		$wrapper_attributes,
		$content,
	);

```

StevenDufresne: I see there are other conditionals, so maybe the example maybe not work exactly.

source/wp-content/themes/wporg-developer-2023/src/command-github/block.php
```diff
@@ -0,0 +1,143 @@
...
+		esc_url( $closed_issues ),
+		__( 'View Closed Issues', 'wporg' ),
+	);
+
+	if ( false !== $issue_count['closed'] ) {
+		$content .= sprintf(
+			'
+			<span class="red">(%s)</span>
+			',
+			(int) $issue_count['closed']
```
StevenDufresne: I think we are supposed to use [number_format_i18n](https://developer.wordpress.org/reference/functions/number_format_i18n/) for numbers.

source/wp-content/themes/wporg-developer-2023/src/command-github/block.php
outdoor2kode: I separated them and tidied up the code a bit in [Adjust code format for a better readability](https://github.com/WordPress/wporg-developer/pull/211/commits/b154744be82ac0c577c5f04447ec5ec602325293)

source/wp-content/themes/wporg-developer-2023/src/command-github/block.php
outdoor2kode: 👌 

[Add number_format_i18n for numbers](https://github.com/WordPress/wporg-developer/pull/211/commits/2ce67f2a008e7eb5a0f4a7c898939a2b9e155ec7)

source/wp-content/themes/wporg-developer-2023/src/command-content/block.php
outdoor2kode: Force pushing fails it. This is the [one](https://github.com/WordPress/wporg-developer/pull/211/commits/1bb49ddc8c2f686c607771c6ccb5d464a9a3fd15). 🙏

outdoor2kode:
> Using a CSS approach would mean that the table of contents would still have uppercase titles, right?

👍 

> I think we'll need to update the content. We can look at the importer or alternatively filter them in code.

A function was added to filter the post content heading in this [commit](https://github.com/WordPress/wporg-developer/pull/211/commits/aab0a478b5f431ac0a2470d04d1f0b79b5f7e053).

source/wp-content/themes/wporg-developer-2023/src/command-content/block.php
StevenDufresne: I think if we ever solve post-content language translations, we'll probably not want to come back and fix these (and it'll be hard to find). Maybe it makes sense to just do it now.

source/wp-content/themes/wporg-developer-2023/src/command-content/block.php
```diff
@@ -0,0 +1,115 @@
...
+	return true;
+}
+
+/**
+ * Convert the heading of post content to camel case.
+ *
+ * @param string $content
+ * @return string Converted content.
+ */
+function camalize_heading( $content ) {
```
StevenDufresne: It doesn't really camel case but does title casing. Maybe something like`format_headings` may be a better name.

StevenDufresne (approved):
This is looking good. Let's merge it. 

There seems to be one issue that will probably require a similar fix to #183. The search drops into the main code search. Can we open a ticket?

outdoor2kode:
> This is looking good. Let's merge it.
> 
> There seems to be one issue that will probably require a similar fix to #183. The search drops into the main code search. Can we open a ticket?

Ticket opened: https://github.com/WordPress/wporg-developer/issues/225

Issue #200, opened 2023-02-03
Code reference search: unclear how to apply filters


ryelle:
When on the search results page, if I want to apply a filter (for example, only show functions), I can click the checkbox, but it's not clear how to apply that filter. I ended up clicking back into the search box and hitting enter.

The little magnifying glass is technically the search button, so it submits the filters too, but it doesn't look like a button, and since it's contained in the input it looks like it's only for the text field.

<img width="745" alt="Screenshot 2023-02-03 at 1 38 32 PM" src="https://user-images.githubusercontent.com/541093/216681501-a4a05007-6ddb-4067-9f10-fa57c17f6b33.png">

The checkbox should not automatically submit the form, as that causes a page load and would be bad for a11y/UX.

cc @WordPress/meta-design

StevenDufresne:
The taxonomy archive has the same problem:

![image](https://user-images.githubusercontent.com/1657336/216885773-66b745a8-268f-41c1-b21c-fb2e4ffc2d44.png)

[Figma](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-Resources?node-id=1160%3A10164&t=oU0nxF52BZkGOBc3-0)

StevenDufresne:
Here's a not-so-nice take on it, a button!

<img width="749" alt="Screen Shot 2023-02-06 at 2 11 09 PM" src="https://user-images.githubusercontent.com/1657336/216888749-4161409b-07f3-4a3b-a0bc-822de7a64f5f.png">

Branch: [fix/button-to-search](https://github.com/WordPress/wporg-developer/tree/fix/button-to-search)

jasmussen:
This issue exists in production already, correct? 
<img width="1028" alt="Screenshot 2023-02-06 at 09 20 56" src="https://user-images.githubusercontent.com/1204802/216920406-a8b03bf7-04fc-4813-95ea-1235eeb6dd57.png">

Since the search icon is technically a button, is it a bit awkward, tabbing-wise, to have two submit buttons that do the same, but have different labels? 

Openverse reloads the page immediately on click of a filter, though that appears to be using some React smarts so focus isn't lost. But that seems like a better longer term concept. 

We can make the apply button work if need be, but it strikes me as extra UI so it'd be nice to look at best practices in this space before we go that route. What do other non-react based search engines do?

ryelle:
> This issue exists in production already, correct?

So it does. We should still fix it 🙂 

[...]

Issue #179, opened 2023-01-26
Accessibility Audit


StevenDufresne:
Use devtools to scan all pages.

Test with keyboard.

Test with VoiceOver and NVDA.

Seek community feedback in #accessibility

adamwoodnz:
We addressed a lot of page layout issues in [Fix accessibility issues with page layouts](https://github.com/WordPress/wporg-developer/pull/376) and [Fix WP-CLI Commands layouts accessibility](https://github.com/WordPress/wporg-developer/pull/402).

adamwoodnz:
There are some contrast ratio issues on code reference pages with WCAG 2.1 AA

![Screenshot 2023-11-23 at 4 05 23 PM](https://github.com/WordPress/wporg-developer/assets/1017872/93357eb4-9df3-4e21-ab56-d7fdac2ab970)

adamwoodnz:
PHP is not a valid `lang` value, used on source code blocks

![Screenshot 2023-11-23 at 4 08 10 PM](https://github.com/WordPress/wporg-developer/assets/1017872/052110e4-bcb6-4082-a6ad-c1ad86c985f7)

adamwoodnz:
GitHub links don't meet contrast ratios with WCAG 2.1 AA

![Screenshot 2023-11-23 at 4 10 20 PM](https://github.com/WordPress/wporg-developer/assets/1017872/39c768db-2f4a-4da2-bb59-c3e299f2a9ad)

adamwoodnz:
Duplicate link in related with no accessible text

![Screenshot 2023-11-23 at 4 14 02 PM](https://github.com/WordPress/wporg-developer/assets/1017872/4539f501-2fea-407c-bf21-9252cc5dd96c)

adamwoodnz:
User notes form needs labels

![Screenshot 2023-11-23 at 4 15 03 PM](https://github.com/WordPress/wporg-developer/assets/1017872/6612733d-82cb-4dd5-b5e1-db60e64859b6)

![Screenshot 2023-11-23 at 4 16 54 PM](https://github.com/WordPress/wporg-developer/assets/1017872/79750089-29f0-4fc2-ba78-a5a23bcc5a57)

adamwoodnz:
Shared the call for testing https://wordpress.slack.com/archives/C02RP4X03/p1700709677752379

adamwoodnz:
> There are some contrast ratio issues on code reference pages with WCAG 2.1 AA
[...]

PR #217, opened 2023-02-10
Implement the "since" archive view.

### LINKED ISSUE #216: Implement WP version archive view.
Users should be able to view the code changes between wp versions. This page also contains filters and a unique update to the breadcrumb.

## Screenshot
image

[Figma](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-Resources?node-id=1382%3A14258&t=w09GjUs65FuEPvwO-0)

StevenDufresne:
Fixes #216 

## Considerations
- I added a button to the search filters discussed in #200. because it's needed for this page.
  - Ideally, we could use a Gutenberg `<Button>` but there doesn't seem to be a way to make it render as a `<button>` element. 
  - I had to add in styles manually which isn't ideal.
- I created a `form-wrapper` component that utilizes `<InnerBlocks>` so we can add blocks to a `<form>`.
  - I don't know if this is a good or bad idea :) 
- The designs don't have the `@since: x.x.x` title but we added it in #201.

## Screenshots

| Since Archive |  Search | 
| --- | --- |
| localhost_8888_reference_since_5 3 0__post_type%5B0%5D=wp-parser-function post_type%5B1%5D=wp-parser-hook post_type%5B2%5D=wp-parser-class post_type%5B3%5D=wp-parser-method | localhost_8888__s=get (1) |

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/form-wrapper/block.json
+++ b/source/wp-content/themes/wporg-developer-2023/src/form-wrapper/block.json
@@ -7,7 +7,6 @@
 	"category": "widgets",
 	"icon": "smiley",
 	"description": "Show the changelog for a function, class, or method.",
-	"usesContext": [ "postId" ],
 	"supports": {
 		"html": false,
 		"spacing": {
--- a/source/wp-content/themes/wporg-developer-2023/src/form-wrapper/block.json
+++ b/source/wp-content/themes/wporg-developer-2023/src/form-wrapper/block.json
@@ -1,0 +1,28 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/block.json",
+	"apiVersion": 2,
+	"name": "wporg/form-wrapper",
+	"version": "0.1.0",
+	"title": "Form Wrapper",
+	"category": "widgets",
+	"icon": "smiley",
+	"description": "Show the changelog for a function, class, or method.",
+	"usesContext": [ "postId" ],
+	"supports": {
+		"html": false,
+		"spacing": {
+			"margin": [
+				"top",
+				"bottom"
+			],
+			"padding": true
+		},
+		"typography": {
+			"fontSize": true,
+			"lineHeight": true
+		}
+	},
+	"textdomain": "wporg",
+	"editorScript": "file:./index.js",
+	"style": "file:./style-index.css"
+}
--- a/source/wp-content/themes/wporg-developer-2023/src/form-wrapper/block.json
+++ b/source/wp-content/themes/wporg-developer-2023/src/form-wrapper/block.json
@@ -10,15 +10,8 @@
 	"supports": {
 		"html": false,
 		"spacing": {
-			"margin": [
-				"top",
-				"bottom"
-			],
+			"margin": true,
 			"padding": true
-		},
-		"typography": {
-			"fontSize": true,
-			"lineHeight": true
 		}
 	},
 	"textdomain": "wporg",
--- a/source/wp-content/themes/wporg-developer-2023/src/form-wrapper/block.json
+++ b/source/wp-content/themes/wporg-developer-2023/src/form-wrapper/block.json
@@ -6,7 +6,7 @@
 	"title": "Form Wrapper",
 	"category": "widgets",
 	"icon": "smiley",
-	"description": "Show the changelog for a function, class, or method.",
+	"description": "A form wrapper block that allows you to add a form to your post or page.",
 	"supports": {
 		"html": false,
 		"spacing": {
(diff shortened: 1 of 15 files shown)


StevenDufresne:
@ryelle What do you think about creating a [generic form wrapper](https://github.com/WordPress/wporg-developer/blob/2ac64372e44dd6521a2d9c5f6fe05c1fd5cf84a9/source/wp-content/themes/wporg-developer-2023/src/form-wrapper/block.php)? Would that be useful for anything or making more specific to this use case make sense?

Because everything is already wired to look for specific query vars (`wp-parser-since`), we just need a form that submits.

source/wp-content/themes/wporg-developer-2023/src/form-wrapper/block.json
```diff
@@ -0,0 +1,20 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/block.json",
+	"apiVersion": 2,
+	"name": "wporg/form-wrapper",
+	"version": "0.1.0",
+	"title": "Form Wrapper",
+	"category": "widgets",
+	"icon": "smiley",
+	"description": "Show the changelog for a function, class, or method.",
```
ryelle: Copy-paste mistake :)

source/wp-content/themes/wporg-developer-2023/src/search-filters/index.php
```diff
@@ -62,8 +62,9 @@ function render( $attributes, $content, $block ) {
 
 	$wrapper_attributes = get_block_wrapper_attributes();
 	return sprintf(
-		'<div %1$s>%2$s</div>',
+		'<div %1$s>%2$s <button>%3$s</button></div>',
```
ryelle: Not required, but it could be more clear to explicitly set the type:
```suggestion
		'<div %1$s>%2$s <button type="submit">%3$s</button></div>',
```

source/wp-content/themes/wporg-developer-2023/src/version-select/index.php
```diff
@@ -57,19 +57,27 @@ function( $version ) {
...
 
 	$options = '';
+	$current_version = get_query_var( 'wp-parser-since' );
 	foreach ( array_merge( $versions, $mu_versions ) as $version ) {
-		$options .= '<option value="' . $version->name . '">' . $version->name . '</option>';
+
+		$options .= sprintf(
+			'<option value="%1$s" %2$s>%3$s</option>',
+			esc_attr( $version->slug ),
+			esc_attr( isset( $current_version ) && $version->name === $current_version ? 'selected' : '' ),
```
ryelle: WP actually has a helper function for this, [`selected(…)`](https://developer.wordpress.org/reference/functions/selected/).

ryelle (commented):
> What do you think about creating a [generic form wrapper](https://github.com/WordPress/wporg-developer/blob/2ac64372e44dd6521a2d9c5f6fe05c1fd5cf84a9/source/wp-content/themes/wporg-developer-2023/src/form-wrapper/block.php)? Would that be useful for anything or making more specific to this use case make sense?

We'll need one for theme directory filters too, so I figured we'd end up adding one to wporg-mu-plugins or the parent theme soon. Is there anything you want to add here that wouldn't be generic enough to be a wrapper? I thought a generic form wrapper would also include an `action` (and maybe a `method`) attribute, but I don't think either are required at the start - even on themes, it'll be submitting to itself.

I also think we'll want a separate submit button, since as you mention, the core Button block is actually a link. What you've got here (with it in the search-filters block) works for now, though.

As for the search filters block, it would be best if it was wrapped in a `fieldset` so that "Filter by type" could be the `legend`… but fieldset brings in browser styles :/ Maybe that can be part of https://github.com/WordPress/wporg-developer/issues/179.

source/wp-content/themes/wporg-developer-2023/src/form-wrapper/block.json
StevenDufresne: 🤦

StevenDufresne: Secrets revealed.

StevenDufresne:
> We'll need one for theme directory filters too, so I figured we'd end up adding one to wporg-mu-plugins or the parent theme soon. Is there anything you want to add here that wouldn't be generic enough to be a wrapper? I thought a generic form wrapper would also include an `action` (and maybe a `method`) attribute, but I don't think either are required at the start - even on themes, it'll be submitting to itself.

There's nothing specific to this implementation. I didn't add action & method props because I want to do the least amount of work necessary, hoping to get this theme out soon as possible. :) Any issues merging this as is with the promise to move it? I would love to have all the pages complete before I go temporarily offline for a meetup.

> I also think we'll want a separate submit button, since as you mention, the core Button block is actually a link. What you've got here (with it in the search-filters block) works for now, though.

👌 


> As for the search filters block, it would be best if it was wrapped in a `fieldset` so that "Filter by type" could be the `legend`… but fieldset brings in browser styles :/ Maybe that can be part of #179.

Yeah I'll follow up in #179.

ryelle (approved):
> Any issues merging this as is with the promise to move it? I would love to have all the pages complete before I go temporarily offline for a meetup.

I don't think I got to this fast enough, but no issues merging this 👍🏻

PR #220, opened 2023-02-14
Fix extra top border when there are no comments.

### LINKED ISSUE #205: Single page border issue under changelog
When a post doesn't have any user comments, the top border of the document details makes it appear like the changelog has an empty line:

Screen Shot 2023-02-07 at 2 19 34 PM

StevenDufresne:
Fixes: #205 

When we don't have user comments, the changelog bottom border combines with the article-meta border and makes it look broken.

This PR moves the article-meta into the filter so we can use a simple CSS statement to hide the top border when there are no comments. Alternatively, we could create separate templates or something like this, but this seems simpler.

| Before | After |
| --- | --- |
| Screen Shot 2023-02-14 at 11 35 50 AM | Screen Shot 2023-02-14 at 9 31 12 AM |

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -2570,12 +2570,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/WordPress/wporg-mu-plugins.git",
-                "reference": "e07b6db810900dfd71c121b84d9a45d26da91767"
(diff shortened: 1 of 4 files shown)


ryelle (approved):
Looks good 👍🏻

PR #228, opened 2023-03-02
Fix searches from wp-cli interior page

### LINKED ISSUE #225: Fix searches from WP CLI Interior Page
If searching with the search control on the WP CLI Interior Page (cli/commands/{command}), you are taken to functions search results. It should search relative to the WP CLI commands.

For example: `/?s=keyword` -> `/cli/commands/?s=keyword`.

As it sits now, if you visit `/cli/commands/?s=keyword`, the search results template will pick it up and it will look like:

image

Similar issue that can be referenced: https://github.com/WordPress/wporg-developer/issues/183

outdoor2kode:
Fixes #225

## Screencast

Screen Capture on 2023-03-03 at 06-17-57

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/inc/block-hooks.php
+++ b/source/wp-content/themes/wporg-developer-2023/inc/block-hooks.php
@@ -20,17 +20,9 @@ function filter_search_block( $block_content, $block ) {
 	}
 
 	if ( 'command' === get_post_type() ) {
-		$block_content = str_replace(
-			'action="' . esc_url( home_url( '/' ) ) . '"',
-			'action="' . esc_url( home_url( 'cli/commands/' ) ) . '"',
-			$block_content
-		);
+		$block_content = get_block_content_by_home_url( $block_content, home_url( 'cli/commands/' ) );
 	} elseif ( wporg_is_handbook() ) {
-		$block_content = str_replace(
-			'action="' . esc_url( home_url( '/' ) ) . '"',
-			'action="' . esc_url( get_query_var( 'current_handbook_home_url' ) ) . '"',
-			$block_content
-		);
+		$block_content = get_block_content_by_home_url( $block_content, get_query_var( 'current_handbook_home_url' ) );
 	} else {
 		if ( isset( $block['attrs']['className'] ) && strpos( $block['attrs']['className'], 'wporg-filtered-search-form' ) ) {
 			$block_content = str_replace( '</form>', do_blocks( '<!-- wp:wporg/search-filters /-->' ) . '</form>', $block_content );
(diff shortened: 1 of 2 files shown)


StevenDufresne:
Can we change the placeholder?

https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer-2023/inc/block-hooks.php#L51.

Maybe "Search for commands"?

source/wp-content/themes/wporg-developer-2023/inc/breadcrumb-trail.php
```diff
@@ -855,6 +857,24 @@ public function do_handbook_search_items() {
...
 	}
 
+	/**
+	 * Adds the items to the trail items array for wp-cli commands search results.
+	 *
+	 * @since  0.6.0
+	 * @access public
+	 * @return void
+	 */
+	public function do_cli_commands_search_items() {
```
StevenDufresne: This seems to be very similar to ` do_handbook_search_items`. Looks like the only difference is the home URL. Can we dry that up? Maybe pass in the home URL?

source/wp-content/themes/wporg-developer-2023/inc/block-hooks.php
```diff
@@ -24,6 +24,13 @@ function filter_search_block( $block_content, $block ) {
 		if ( isset( $block['attrs']['className'] ) && strpos( $block['attrs']['className'], 'wporg-filtered-search-form' ) ) {
 			$block_content = str_replace( '</form>', do_blocks( '<!-- wp:wporg/search-filters /-->' ) . '</form>', $block_content );
 		}
+		if ( is_singular( 'command' ) ) {
```
StevenDufresne: What do you think about adding this as an else-if clause to flatten the code? 

Maybe something like:

```
if ( is_singular( 'command' ) )  {
	// ...
} elseif ( wporg_handbook() ) {  // flip this logic
	// ..
} else {
	// ...
}
```

StevenDufresne: I also wonder if `is_singular` is too specific. What if a search bar was added to an `archive` page? Maybe just do it based on the post type?

source/wp-content/themes/wporg-developer-2023/inc/block-hooks.php
```diff
@@ -24,6 +24,13 @@ function filter_search_block( $block_content, $block ) {
...
+			$block_content = str_replace(
+				'action="' . esc_url( home_url( '/' ) ) . '"',
```
StevenDufresne: We are doing the exact thing here and below. Can we dry that up? Maybe a simple helper function?

source/wp-content/themes/wporg-developer-2023/inc/block-hooks.php
outdoor2kode: > flatten code?

Good point. I'll work on refactoring the code to make it more streamlined.

> too specific.

yeah, you make a good point. I didn't consider that scenario, thanks for bringing it up.

source/wp-content/themes/wporg-developer-2023/inc/block-hooks.php
```diff
@@ -61,3 +53,22 @@ function render_block_data( $parsed_block ) {
 }
 
 add_filter( 'render_block_data', __NAMESPACE__ . '\render_block_data', 10, 2 );
+
+/**
+ * Replaces the action URL in a block content string with a given URL path.
+ *
+ * @param string $block_content The block content string to modify.
+ * @param string $replacement_home_url Replceament string for the action attribute. Defaults to an empty string.
```
StevenDufresne:
```suggestion
 * @param string $replacement_home_url Replacement string for the action attribute. Defaults to an empty string.
```

source/wp-content/themes/wporg-developer-2023/inc/block-hooks.php
```diff
@@ -61,3 +53,22 @@ function render_block_data( $parsed_block ) {
...
+ * @return string The modified block content string.
+ */
+function get_block_content_by_home_url( $block_content, $replacement_home_url = '' ) {
+	if ( ! $replacement_home_url ) {
```
StevenDufresne: Since we are optionally setting this to an empty string, we can probably just remove this check.

PR #229, opened 2023-03-02
Fix content in code block escape the container

### LINKED ISSUE #198: Handbook: Content escaping container.
The contain is extending left indefinitely.

Screen Shot 2023-02-03 at 1 43 55 PM
Found on: 
http://localhost:8888/block-editor/getting-started/devenv/

outdoor2kode:
Fixes #198 

Fix the issue based on how it works in production. Here also tries to avoid affecting the style of code blocks that already have specific classes assigned to them on the [code page](http://localhost:8888/reference/functions/get_rss/). An alternative idea for this issue is to apply the same class to the `<pre>` tags to the whole site as the code blocks on the code page to maintain consistency in styling, not sure if it makes sense as it's not the case in production, probably for some reasons.

## Screencast
Screen Capture on 2023-03-03 at 07-26-38

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
+++ b/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
@@ -15,7 +15,7 @@ body[class] {
 		max-width: 100%;
 	}
 
-	pre:not(.line-numbers) {
+	pre:not(.wp-block-code) {
 		padding: 20px;
 		background-color: #f7f7f7;
 		border: 1px solid var(--wp--preset--color--light-grey-1);
--- a/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
+++ b/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
@@ -8,10 +8,20 @@ body[class] {
 	--wp-local-header-offset: 57px;
 }
 
-// Parsed handbook content doesn't use image blocks, so large images are
-// missing the max-width and break out of the page container.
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-developer-2023/src/style/style.scss
```diff
@@ -8,10 +8,20 @@ body[class] {
...
-.entry-content p > img {
-	max-width: 100%;
+.entry-content {
+	// Parsed handbook content doesn't use image blocks, so large images are
+	// missing the max-width and break out of the page container.
+	p > img {
+		max-width: 100%;
+	}
+
+	pre:not(.line-numbers) {
```
StevenDufresne: I don't think it's safe to style based on `line-numbers` since that could be removed easily. 
What about using `wp-block-code`? Would that work?

outdoor2kode: I was thinking that it would be ok to be replaced by the styles here when there's only `wp-block-code` since I can't think of or find any case where we wouldn't want line numbers for a code block at the moment. But yeah, it makes more sense to maintain a consistent style in case we ever decide to remove `line-numbers` some day.

StevenDufresne: It's possible that short code blocks, maybe in comments wouldn't get line-numbers... stuff like that.

StevenDufresne:
Looks good. Go for it.

PR #237, opened 2023-04-03
Import the Advanced Admin handbook

dd32:
See https://meta.trac.wordpress.org/ticket/6411

This has been committed to production for the old theme, creating this to import it with the new theme.

I haven't tested this in this theme, but it appears that it should work.


StevenDufresne:
Thanks!

PR #246, opened 2023-05-08
Templates: Use the new global Sidebar Container block

### LINKED ISSUE #242: Sidebar search bar is truncated.
The search bar is truncated. 

Example page url: https://developer.wordpress.org/redesign-test/block-editor/

## Screenshot
Screenshot 2023-05-03 at 8 42 50 PM

ryelle:
In https://github.com/WordPress/wporg-mu-plugins/pull/397, the sticky sidebar behavior was moved to the Sidebar Container block, out of the Table of Contents block. This PR adds in the Sidebar Container, so that the sticky sidebar works again. This also removes the unnecessary CSS for two-column layouts, this is now provided by the global block. We still need the CSS for the 3 col layout, so that's left alone.

Additionally, this fixes #242 by adding the 100% width to the search bar.

**Screenshots**

| Trunk | Branch |
|---|---|
| trunk-2-col | pr-2-col |
| trunk-3-col | pr-3-col |

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/patterns/article-sidebar.php
+++ b/source/wp-content/themes/wporg-developer-2023/patterns/article-sidebar.php
@@ -7,14 +7,8 @@
 
 ?>
 
-<!-- wp:group {"tagName":"aside","className":"sidebar-container"} -->
-<aside class="wp-block-group sidebar-container">
 	<!-- wp:search {"label":"<?php esc_attr_e( 'Search', 'wporg' ); ?>","showLabel":false,"widthUnit":"px","buttonText":"<?php esc_attr_e( 'Search', 'wporg' ); ?>","buttonPosition":"button-inside","buttonUseIcon":true,"className":"is-style-secondary-search-control","borderColor":"light-grey-1","placeholder":"<?php esc_attr_e( 'Search in the documentation', 'wporg' ); ?>"} /-->
+<!-- wp:wporg/sidebar-container -->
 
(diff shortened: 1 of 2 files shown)


StevenDufresne (approved):
I didn't test this locally, but it looks right 👍.

adamwoodnz (commented):
Layout works for me, including the Search bar width, but it's not sticky, and scrolls with the page.

I have my local mu-plugins loaded via wp-env, built, and I have the `.wp-block-wporg-sidebar-container`, but it's `absolute`. Am I missing something?

ryelle:
> Layout works for me, including the Search bar width, but it's not sticky, and scrolls with the page.

hm… is the sidebar shorter than your viewport? (unless your viewport is <450px that's unlikely)

Are you sure wp-env is loading your local mu-plugins? Try `composer update` to update the vendor-provided one, it should already have the sidebar container code in place.

If neither of those help, can you share a screenshot or recording of it not working?

adamwoodnz (approved):
Found the cause of the issue. I was running 6.1 alpha and `viewScript` in `block.json` wasn't working. `view.js` wasn't loaded. Upgraded to 6.2 and it works for me.

I know there were bugs with that in the past, I guess that version was susceptible.

adamwoodnz:
Should it be sticky if the sidebar is longer than the screen height? I'm unable to access the bottom of the ToC in this case

Screen Shot 2023-05-10 at 5 17 08 PM

In the current site I ended up keeping it sticky, but adding a max height and internal scroll, see https://developer.wordpress.org/coding-standards/inline-documentation-standards/javascript/

ryelle:
> Should it be sticky if the sidebar is longer than the screen height? I'm unable to access the bottom of the ToC in this case

No, it should not be sticky — but it doesn't update on resize, so if the sidebar fit, then you shrank the screen, it would still be sticky. if you reload the page at that size, it should stay at the top of the screen.

PR #250, opened 2023-07-26
Add more width to the search input

dd32:
See https://meta.trac.wordpress.org/ticket/7156

https://developer.wordpress.org/plugins/intro/what-is-a-plugin/

per @pitamdey


pitamdey:
Thanks, @dd32 
we can just reduce the padding so that full placeholder text should be fully visible 
`.search-wrap {
	.searchform {
		input[type="search"] {
			padding: 1rem 1rem 1rem 1rem;
                 }
             }
	}`

dd32:
> we can just reduce the padding

That's what the PR does, `1rem` is not appropriate value there though due to the overlayed images.

PR #251, opened 2023-07-26
Update code reference layout

### LINKED ISSUE #243: Code Reference update
The spacing between the sections is not consistent and a lot needs to be updated to match the new [design](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-Resources?type=design&node-id=2934-6876&mode=design&t=q9aehCWavRMKQ6KF-0):

Example Url: https://developer.wordpress.org/redesign-test/reference/functions/absint/

## Screenshot

Screenshot 2023-05-03 at 8 49 23 PM

adamwoodnz:
Closes #243 

Makes the spacing between sections in the code reference single page a consistent 40px, except for the comments section (60px).

Also fixes text wrapping in the comment form tabs and adds edge space to the comment edit page.

#### Desktop

| Before | After |
|-|-|
| localhost_8888_reference_functions_absint_(Desktop) (1) | localhost_8888_reference_functions_absint_(Desktop) (1) |

#### Mobile

| Before | After |
|-|-|
| localhost_8888_reference_functions_absint_(Samsung Galaxy S20 Ultra) (1) | localhost_8888_reference_functions_absint_(Samsung Galaxy S20 Ultra) (1) |

#### Edit comment

| Desktop | Mobile |
|-|-|
| localhost_8888_reference_comment_edit_13_(Desktop) | localhost_8888_reference_comment_edit_13_(Samsung Galaxy S20 Ultra) |


source/wp-content/themes/wporg-developer-2023/src/code-comment-edit/style.scss
```diff
@@ -14,84 +14,5 @@
 			padding-top: var(--wp--preset--spacing--30);
 		}
 	}
-
```
adamwoodnz, line 17: All this CSS is in the [main css](https://github.com/WordPress/wporg-developer/pull/251/files#diff-5a967de023463491bebbfaca0909be4be783855fcace73f96a9a0e9c8060224aR194)

Perhaps it can be scoped to this block... not sure if there are more comment forms at this point

adamwoodnz:
I need to edit the margin at the top of the content, but it's attached to a generated classname `.wp-container-22`. I cannot work out where the style is set, it's in the `core-block-supports-inline-css` stylesheet.

Help please! @StevenDufresne @ryelle 

Screenshot 2023-07-27 at 2 40 19 PM

Staging page https://developer.wordpress.org/redesign-test/reference/functions/absint/

ryelle:
That spacing corresponds to the blockGap set on the container element, it looks like it's set on each `article` group, for example:

https://github.com/WordPress/wporg-developer/blob/57fa8cc148a1d20164ba2c03439e979cdaee1372/source/wp-content/themes/wporg-developer-2023/templates/single.html#L9-L10

The use of this is more clear on pages with the deprecated or private warnings, like [ms_deprecated_blogs_file](https://developer.wordpress.org/redesign-test/reference/functions/ms_deprecated_blogs_file/).

Could you try adding a `blockGap` to [the post-content block here](https://github.com/WordPress/wporg-developer/blob/57fa8cc148a1d20164ba2c03439e979cdaee1372/source/wp-content/themes/wporg-developer-2023/patterns/single.php#L12) to see if that will add the correct spacing between sections? Rather than setting it in the CSS.

adamwoodnz:
> Could you try adding a `blockGap` to [the post-content block here](https://github.com/WordPress/wporg-developer/blob/57fa8cc148a1d20164ba2c03439e979cdaee1372/source/wp-content/themes/wporg-developer-2023/patterns/single.php#L12) to see if that will add the correct spacing between sections? Rather than setting it in the CSS.

This works great, thanks.

source/wp-content/themes/wporg-developer-2023/src/style/style.scss
```diff
@@ -8,6 +8,12 @@ body[class] {
 	--wp-local-header-offset: 57px;
 }
 
+p,
```
adamwoodnz: Can't find a way to do this through theme.json

source/wp-content/themes/wporg-developer-2023/patterns/article-sidebar.php
```diff
@@ -8,7 +8,13 @@
 ?>
 
 <!-- wp:wporg/sidebar-container -->
-	<!-- wp:search {"label":"<?php esc_attr_e( 'Search', 'wporg' ); ?>","showLabel":false,"width":100,"widthUnit":"%","buttonText":"<?php esc_attr_e( 'Search', 'wporg' ); ?>","buttonPosition":"button-inside","buttonUseIcon":true,"className":"is-style-secondary-search-control","borderColor":"light-grey-1","placeholder":"<?php esc_attr_e( 'Search in the documentation', 'wporg' ); ?>"} /-->
 
-	<!-- wp:wporg/table-of-contents /-->
+	<!-- wp:group {"tagName":"article","style":{"spacing":{"padding":{"top":"var:preset|spacing|20"}}}} -->
+	<div class="wp-block-group" style="padding-top:var(--wp--preset--spacing--20)">
```
adamwoodnz, line 13: This padding adjustment should be temporary until we can move the search field to the top of the Documentation pages too, then the margin-top of the sidebar-container should be changed instead.

source/wp-content/themes/wporg-developer-2023/src/code-comment-form/block.php
```diff
@@ -53,10 +53,12 @@ function render( $attributes, $content, $block ) {
...
 
-	$title_block = sprintf(
-		'<!-- wp:heading {"fontSize":"heading-4"} --><h2 class="wp-block-heading has-heading-4-font-size">%s</h2><!-- /wp:heading -->',
+	$ordered_comments = wporg_developer_get_ordered_notes();
+
+	$title_block = empty( $ordered_comments ) ? sprintf(
+		'<!-- wp:heading --><h2 class="wp-block-heading">%s</h2><!-- /wp:heading -->',
 		__( 'User Contributed Notes', 'wporg' )
-	);
+	) : '';
```
adamwoodnz, line 61: If there are comments the heading will be displayed at the top of that block, and won't be required again here.

source/wp-content/themes/wporg-developer-2023/src/style/style.scss
ryelle: What is it you're trying to do? Would adjusting the block gap help?

The root block gap can be set in the `styles` property in theme.json:

```json
{
	"styles": {
		"spacing": {
			"blockGap": "var(--wp--preset--spacing--20)"
		}
	}
}
```

adamwoodnz: Oh yeah maybe that's it. These elements are just using default browser margin at present

adamwoodnz: No that doesn't seem to work

adamwoodnz: What I'm trying to do is something like

```
"elements": {
			"p": {
				"spacing": {
					"blockGap": "var(--wp--preset--spacing--20)"
				}
			}
		},
```

But that isn't valid, the list of available elements is very limited according to the schema

ryelle: Hm, are you trying to add this spacing inside another block like the description?

Screenshot 2023-08-30 at 6 27 25 PM

I bet it's not working because it expects the custom block to handle it… it looks like you can trigger the block gap CSS by adding layout support to the description block (and others that need this), but that doesn't seem right, since you don't actually want full layout support.

```diff
	"supports": {
		"html": false,
		"spacing": {
			"margin": [ "top", "bottom" ],
			"padding": true
		},
		"typography": {
			"fontSize": true,
			"lineHeight": true
-		}
+		},
+		"layout": {
+			"default": {
+				"type": "constrained"
+			}
+		}
	},
```

Maybe mimicking the CSS in the block is a better idea? Like dropping this into `src/code-description/style.scss`:

```css
:where(.wp-block-wporg-code-reference-description) > * {
	margin-block-start: var(--wp--style--block-gap);
	margin-block-end: 0;
}

:where(.wp-block-wporg-code-reference-description) > :first-child {
	margin-block-start: 0;
}

:where(.wp-block-wporg-code-reference-description) > :last-child {
	margin-block-end: 0;
}
```

The `--wp--style--block-gap` variable should match whatever the theme has set as the default block gap, not the current block gap in the hierarchy of the page (noteworthy since this is inside the `post-content` block which sets blockGap to `var(--wp--preset--spacing--40)`).

Hopefully that helps a bit? 🙂

adamwoodnz: Yeah thanks that is helpful 🙂 

I've added a [commit](https://github.com/WordPress/wporg-developer/pull/251/commits/5f5a6a620b167290d82fc42adb4c1b5ccd843032) which uses the `--wp--style--block-gap` everywhere I need it in this page. Seems like so much more code required than being able to set base element styles through the theme settings.

Let me know if you think that's the right approach

source/wp-content/themes/wporg-developer-2023/src/code-comment-edit/style.scss
```diff
@@ -1,97 +1 @@
-.wp-block-wporg-code-reference-comment-form-content {
```
adamwoodnz, line 1: This seems to be a copy paste of the CSS from this other block

adamwoodnz:
I'm going to merge this to keep things moving, and I'll follow up any new feedback asap.

PR #252, opened 2023-07-27
Fix `get_page_by_title` deprecation warnings

adamwoodnz:
Fixes #206 

Removes warnings about `Deprecated: Function get_page_by_title is deprecated since version 6.2.0! Use WP_Query instead`.

### Testing

- Ensure you have a Resource page and a Resource/Dashicons page. Navigate to `/resource` and you should be redirected to `resource/dashicons`

I'm not sure how to test the usages in `/inc` as I'm not sure how to debug the parser. @tellyworth [it looks like you were the author of some of the original code](https://github.com/WordPress/wporg-developer/blame/old/source/wp-content/themes/wporg-developer/inc/template-tags.php#L890), any thoughts please?

Issue #206, opened 2023-02-07
PHP deprecation notice: get_page_by_title is deprecated since version 6.2.0


ryelle:
> Deprecated: Function get_page_by_title is deprecated since version 6.2.0! Use WP_Query instead.

Looks like we're using this function in a few places:

https://github.com/WordPress/wporg-developer/blob/3f5c6b8c68e88cfafac9fb5bee7b7a5d55466c16/source/wp-content/themes/wporg-developer-2023/inc/formatting.php#L415

https://github.com/WordPress/wporg-developer/blob/3f5c6b8c68e88cfafac9fb5bee7b7a5d55466c16/source/wp-content/themes/wporg-developer-2023/inc/formatting.php#L426

https://github.com/WordPress/wporg-developer/blob/3f5c6b8c68e88cfafac9fb5bee7b7a5d55466c16/source/wp-content/themes/wporg-developer-2023/inc/formatting.php#L468

https://github.com/WordPress/wporg-developer/blob/3f5c6b8c68e88cfafac9fb5bee7b7a5d55466c16/source/wp-content/themes/wporg-developer-2023/inc/template-tags.php#L879

https://github.com/WordPress/wporg-developer/blob/3f5c6b8c68e88cfafac9fb5bee7b7a5d55466c16/source/wp-content/themes/wporg-developer-2023/inc/redirects.php#L68

PR #261, opened 2023-08-10
Fix namespace for WP_Query usage

adamwoodnz:
See #206 

This fixes issues with namespaces for WP_Query, see https://github.com/WordPress/wporg-developer/pull/252#issuecomment-1672539753

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/inc/template-tags.php
+++ b/source/wp-content/themes/wporg-developer-2023/inc/template-tags.php
@@ -885,7 +885,7 @@ namespace DevHub {
 					'posts_per_page' => 1,
 				);
 			
-				$query = new WP_Query( $args );
+				$query = new \WP_Query( $args );
 			
 				if ( $query->have_posts() ) {
 					$_post = $query->posts[0];


source/wp-content/themes/wporg-developer-2023/inc/redirects.php
```diff
@@ -71,7 +71,7 @@ public static function redirect_resources() {
 				'posts_per_page' => 1,
 			);
 		
-			$query = new WP_Query( $args );
+			$query = new \WP_Query( $args );
```
adamwoodnz: I did test this one and it worked... is this fix required?

dd32: This one isn't, as it's within a class that's within the default namespace.
but there's no harm in it either.

source/wp-content/themes/wporg-developer-2023/inc/formatting.php
```diff
@@ -491,7 +491,7 @@ function ( $matches ) {
 							'posts_per_page' => 1,
 						);
 						
-						$query = new WP_Query( $args );
+						$query = new \WP_Query( $args );
```
dd32: The ones in this file don't seem needed either, as they're also within a class in the default namespace.

source/wp-content/themes/wporg-developer-2023/inc/template-tags.php
```diff
@@ -885,7 +885,7 @@ function get_param_reference( $param, $recursion_limit = 3 ) {
 					'posts_per_page' => 1,
 				);
 			
-				$query = new WP_Query( $args );
+				$query = new \WP_Query( $args );
```
dd32, line 888: I think this might be the only one that's actually needed, as this function is within the DevHub namespace

source/wp-content/themes/wporg-developer-2023/inc/formatting.php
adamwoodnz: Reverted

source/wp-content/themes/wporg-developer-2023/inc/redirects.php
adamwoodnz: Reverted

PR #282, opened 2023-09-11
Add redirect for block editor handbook theme overview

### LINKED ISSUE #281: Redirect block theme overview
Redirect https://developer.wordpress.org/block-editor/how-to-guides/themes/block-theme-overview/ to https://developer.wordpress.org/themes/block-themes/

adamwoodnz:
Closes #281 

Redirects https://developer.wordpress.org/block-editor/how-to-guides/themes/block-theme-overview/ to https://developer.wordpress.org/themes/block-themes/

As `inc/import-block-editor.php` is where we handle all redirects specifically for the block editor handbook, I think it makes sense for it to be defined here, rather than `wporg-redirects.php`, but I could be convinced otherwise.

This PR is for the `wporg-developer` theme currently running in production. I'll create a corresponding PR to `trunk` for the redesign if this is approved.

### DIFF
--- a/source/wp-content/themes/wporg-developer/inc/import-block-editor.php
+++ b/source/wp-content/themes/wporg-developer/inc/import-block-editor.php
@@ -92,10 +92,20 @@ class DevHub_Block_Editor_Importer extends DevHub_Docs_Importer {
 			'tutorials/' => 'how-to-guides/',
 		];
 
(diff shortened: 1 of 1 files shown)


iandunn (approved):
LGTM 👍🏻

PR #288, opened 2023-09-21
Update CLI commands page layout

### LINKED ISSUE #264: WP-CLI page update
New [design](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-Resources?type=design&node-id=2934-7220&mode=design&t=q9aehCWavRMKQ6KF-0):

WP-CLI Commands

### Requirements

- [x] Add search
- [x] Add 'Other Developer Resources' section
- [x] Move content into single column and add sidebar with ToC

### LINKED ISSUE #277: CLI pages are broken.
You get the following when viewing a cli command on trunk:

http://localhost:8888/cli/commands/super-admin/
`Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /var/www/html/wp-includes/class-wp-list-util.php on line 168`

Looks like an infinite loop related to loading blocks.

adamwoodnz:
Closes #264 
Fixes #277 

Depends on https://github.com/WordPress/wporg-parent-2021/pull/110 which updates the card grid styles to use LinkWrapper blocks, enabling better markup for the cards which is compatible with the ToC.

Adds search, and 'Other Developer Resources' section.

Reformats content into one column and adds sidebar with ToC.

The h1 is included in the ToC, which isn't consistent with other pages like the [Code Reference](http://localhost:8888/reference/functions/absint/). @WordPress/meta-design what's correct here please?

#### Commands archive

| Desktop | Tablet | Mobile |
|-|-|-|
| localhost_8888_cli_commands_(Desktop) (3) | localhost_8888_cli_commands_(iPad) (2) | localhost_8888_cli_commands_(Samsung Galaxy S20 Ultra) (3) |

#### Single Command

| Desktop | Tablet | Mobile |
|-|-|-|
| localhost_8888_cli_commands_admin_(Desktop) | localhost_8888_cli_commands_admin_(iPad) | localhost_8888_cli_commands_admin_(Samsung Galaxy S20 Ultra) |

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/inc/block-hooks.php
+++ b/source/wp-content/themes/wporg-developer-2023/inc/block-hooks.php
@@ -35,26 +35,6 @@ function filter_search_block( $block_content, $block ) {
 add_filter( 'render_block', __NAMESPACE__ . '\\filter_search_block', 10, 2 );
 
 /**
- * Filters the search block and updates the placeholder.
- *
- * @param string $parsed_block
- * @return array
- */
-function render_block_data( $parsed_block ) {
-	if ( 'core/search' !== $parsed_block['blockName'] ) {
-		return $parsed_block;
-	}
-
-	if ( is_parsed_post_type() || 'command' === get_post_type() ) {
-		$parsed_block['attrs']['placeholder'] = __( 'Search for commands...', 'wporg' );
-	}
-
-	return $parsed_block;
-}
-
-add_filter( 'render_block_data', __NAMESPACE__ . '\render_block_data', 10, 2 );
-
-/**
  * Replaces the action URL in a block content string with a given URL path.
  *
  * @param string $block_content The block content string to modify.
--- a/source/wp-content/themes/wporg-developer-2023/patterns/cli-commands-content.php
+++ b/source/wp-content/themes/wporg-developer-2023/patterns/cli-commands-content.php
@@ -29,15 +29,23 @@
 <!-- /wp:heading -->
 
 <!-- wp:group {"style":{"spacing":{"blockGap":"var:preset|spacing|10"}},"className":"is-style-cards-grid","layout":{"type":"grid","minimumColumnWidth":"49%"}} -->
-<div class="wp-block-group is-style-cards-grid">
-	
-	<!-- wp:heading {"level":3,"fontSize":"small","fontFamily":"inter"} -->
-	<h3 class="wp-block-heading has-inter-font-family has-small-font-size"><a href="https://make.wordpress.org/cli/"><strong><?php esc_html_e( 'CLI Blog', 'wporg' ); ?></strong><?php esc_html_e( 'Catch up on the latest on WP-CLI in the main updates blog.', 'wporg' ); ?></a></h3>
-	<!-- /wp:heading -->
+<div class="wp-block-group is-style-cards-grid"><!-- wp:wporg/link-wrapper -->
+<a class="wp-block-wporg-link-wrapper" href="https://make.wordpress.org/cli/"><!-- wp:heading {"level":3,"style":{"typography":{"fontStyle":"normal","fontWeight":"700"}},"fontSize":"small","fontFamily":"inter"} -->
+<h3 class="wp-block-heading has-inter-font-family has-small-font-size" style="font-style:normal;font-weight:700"><?php esc_html_e( 'CLI Blog', 'wporg' ); ?></h3>
+<!-- /wp:heading -->
 
-	<!-- wp:heading {"level":3,"fontSize":"small","fontFamily":"inter"} -->
-	<h3 class="wp-block-heading has-inter-font-family has-small-font-size"><a href="https://make.wordpress.org/cli/handbook/"><strong><?php esc_html_e( 'CLI Handbook', 'wporg' ); ?></strong><?php esc_html_e( 'A collection of helpful guides and resources for using WP-CLI.', 'wporg' ); ?></a></h3>
-	<!-- /wp:heading -->
+<!-- wp:paragraph {"fontSize":"small"} -->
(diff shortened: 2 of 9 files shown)


source/wp-content/themes/wporg-developer-2023/inc/block-hooks.php
```diff
@@ -34,26 +34,6 @@ function filter_search_block( $block_content, $block ) {
...
- * @param string $parsed_block
- * @return array
- */
-function render_block_data( $parsed_block ) {
-	if ( 'core/search' !== $parsed_block['blockName'] ) {
-		return $parsed_block;
-	}
-
-	if ( is_parsed_post_type() || 'command' === get_post_type() ) {
-		$parsed_block['attrs']['placeholder'] = __( 'Search for commands...', 'wporg' );
```
adamwoodnz, line 49: The design doesn't use this customization anymore

source/wp-content/themes/wporg-developer-2023/patterns/cli-commands.php
```diff
@@ -7,24 +7,55 @@
...
+<div class="wp-block-group alignwide" style="margin-top:var(--wp--preset--spacing--20);margin-bottom:var(--wp--preset--spacing--20)"><!-- wp:search {"label":"<?php esc_attr_e( 'Search', 'wporg' ); ?>","showLabel":false,"placeholder":"<?php esc_attr_e( 'Search resources', 'wporg' ); ?>","width":240,"widthUnit":"px","buttonText":"Search","buttonPosition":"button-inside","buttonUseIcon":true} /--></div>
+<!-- /wp:group -->
+
+<!-- wp:group {"align":"wide","style":{"spacing":{"margin":{"top":"var:preset|spacing|20"}}},"layout":{"type":"constrained","justifyContent":"left"}} -->
+<div class="wp-block-group alignwide" style="margin-top:var(--wp--preset--spacing--20)"><!-- wp:group {"tagName":"article","style":{"spacing":{"blockGap":"var:preset|spacing|40","margin":{"top":"0px"}}}} -->
+<article class="wp-block-group" style="margin-top:0px"><!-- wp:wporg/sidebar-container -->
+<!-- wp:group -->
+<div class="wp-block-group"><!-- wp:wporg/table-of-contents /--></div>
```
adamwoodnz: This block expects to be able to [get the content of the page using the post id in the block context](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/table-of-contents/index.php#L43). Because this is an archive template and posts are queried, the post id in the context is that of the first command post, not the page. Need to find a solution for getting the template content without creating an infinite loop.

pkevan: sounds like the main query needs resetting - do you know how the posts are queried for the commands?

ryelle: There isn't a post here to get the content from, the headings here are hardcoded in the pattern.

Could you use a styled list block? it's a little duplicate work to get the styles right, but that's probably better than trying to use the dynamic block. I guess it wouldn't collapse, but there are only two headings. hmm… Does the core table of contents block help? No, it's not dynamic at all.

There's the `wporg_table_of_contents_post_content` filter, if you could use that somehow to "load" in the pattern content, then we could probably remove the `! isset( $block->context['postId'] )` check in the ToC block?

adamwoodnz: > There isn't a post here to get the content from, the headings here are hardcoded in the pattern.

Very good point 😂 

Thanks, I'll have a play with these ideas.

adamwoodnz: > There's the `wporg_table_of_contents_post_content` filter, if you could use that somehow to "load" in the pattern content

This works nicely, see https://github.com/WordPress/wporg-developer/pull/288/commits/74d955d69e9bea632555e017c94521e87de1903b where I split out the content to be parsed (static content minus the h1) and feed it through the filter.

source/wp-content/themes/wporg-developer-2023/patterns/cli-commands-content.php
```diff
@@ -0,0 +1,51 @@
...
+		'https://make.wordpress.org/cli/'
+	)
+); ?></p>
+<!-- /wp:paragraph -->
+
+<!-- wp:wporg/cli-command-table {"style":{"spacing":{"margin":{"top":"var:preset|spacing|20"}}}} /-->
+
+<!-- wp:heading {"className":"is-toc-heading"} -->
+<h2 class="wp-block-heading is-toc-heading" id="other-developer-resources"><a href="#other-developer-resources"><?php esc_html_e( 'Other Developer Resources', 'wporg' ); ?></a></h2>
+<!-- /wp:heading -->
```
adamwoodnz, line 33: Manual class, id and anchor needed here as the ToC block doesn't add them. The [id injection filters the content](https://github.com/WordPress/wporg-mu-plugins/blob/build/mu-plugins/blocks/table-of-contents/index.php#L89), which again doesn't work in this scenario.

jasmussen:
Nice work. Looks right.

A question mostly for curiosity:
Screenshot 2023-09-22 at 09 59 37

Is that `wp admin` with a spacer? I guess that's a wp cli command called `wp`? 😅 

I've not yet had a chance to use wp cli that much, so I see `wp admin` and think it should be `wp-admin`, so just checking to be sure!

adamwoodnz:
> Is that `wp admin` with a spacer? I guess that's a wp cli command called `wp`?

Yes, and that looks weird for me too 🙂 

The wp cli command is `admin` but it's a sub command to the base `wp` command (I think that's the correct terminology)

jasmussen:
Thanks!

PR #292, opened 2023-09-26
Update Since page layout

### LINKED ISSUE #263: Reference/Since page update
New [design](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-Resources?type=design&node-id=2934-6854&mode=design&t=q9aehCWavRMKQ6KF-0):

Since

### Requirements

1. Move Release number to seperate filter
2. Restyle post type filters
3. Add Search box

adamwoodnz:
Closes #263 

Depends on https://github.com/WordPress/wporg-parent-2021/pull/112 for button and checkbox toggle button styles.

Updates the taxonomy page for `wp-parser-since` to match the [design](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-Resources?node-id=2934%3A6854&mode=dev).

### Changes

- Add Search field
- Split existing filters into seperate release selector and post type filters
- Change the query title prefix to 'New and updated in ...'
- Minor font style and spacing updates in the post content items

The only complex part of this was reworking the post type filters; adding the 'All' control and restyling the checkboxes to display like buttons. I initially tried to add the 'All' control as another checkbox with no value, but submitting the form resulted in a 404 as the post_type query var was empty. My solution is to use a button which simply clears the other checkboxes so that the form is submitted with no post_type query vars. The logic for all this is in a new [view script](https://github.com/WordPress/wporg-developer/pull/292/files#diff-df917505c8b2f284644e1b9b2c5beb3e5c04c061664bb79d80710a7be6ca0f3a).

## Screenshots

| Desktop | Tablet | Mobile |
|-|-|-|
| localhost_8888_reference_since_6 1 0_page_2__wp-parser-since=6-1-0(Desktop) | localhost_8888_reference_since_6 1 0_page_2__wp-parser-since=6-1-0(iPad) | localhost_8888_reference_since_6 1 0_page_2__wp-parser-since=6-1-0(Samsung Galaxy S20 Ultra) |

| State | |
|-|-|
| Filters checked | Screenshot 2023-09-28 at 2 06 00 PM |
| Hover filters and buttons | Screenshot 2023-09-28 at 1 57 33 PM |
| Focus filters and buttons | Screenshot 2023-09-28 at 1 58 22 PM |
| Active filters and buttons | Screenshot 2023-09-28 at 2 05 43 PM |

## Testing

1. Use wp-env to set your parent theme to the [button styles branch](https://github.com/WordPress/wporg-parent-2021/pull/112)
2. Navigate to http://localhost:8888/reference/since/6.3.0/
3. Test changing release version using the dropdown
4. Test filtering the results by post type, including using the 'All' button
5. Test with keyboard and screen reader
6. Check interactive states, which are based on [Button/Small in the Design Library](https://www.figma.com/file/HOJTpCFfa3tR0EccUlu0CM/WordPress.org-Design-Library?type=design&node-id=2710-958&mode=design&t=pXEyBCRAQ1yeAbV4-0)


jasmussen:
At a glance, looks good. 

Two things, first off, the border radius of the focus style (which I assume to be 1.5px) looks a bit round, and notably in this on it looks like it's clashing with the inner radius. 

Screenshot 2023-09-28 at 10 49 36

Functionally these styles should be _mostly_ identical to what's in [the block editor](https://wordpress.github.io/gutenberg/?path=/docs/components-button--docs).

The second thing, these buttons should ideally be identical across all of wp.org, with no hover/active/focus styles that are unique to each section. I'm not in the loop on what's technically possible as far as re-use across the site, but if there's a way to do that, we definitely want to. 

Let me know if that helps.

source/wp-content/themes/wporg-developer-2023/functions.php
```diff
@@ -231,6 +232,19 @@ function init() {
 	add_filter( 'wporg_canonical_url', '__return_false' );
 }
 
+/**
+ * Change the prefix of the query title on the `reference/since/x.x.x` page
+ */
+function get_reference_since_title_prefix( $prefix ) {
+	$queried_object = get_queried_object();
+
+	if ( is_archive() && $queried_object && 'wp-parser-since' === $queried_object->taxonomy ) {
```
ryelle: Can you use [is_tax( 'wp-parser-since' )](https://developer.wordpress.org/redesign-test/reference/functions/is_tax/) here to simplify this?

source/wp-content/themes/wporg-developer-2023/src/search-filters/index.php
```diff
@@ -43,20 +43,26 @@ function render( $attributes, $content, $block ) {
...
+	);
+	$content .= '</div>';
 
 	foreach ( $search_post_types as $post_type => $label ) {
 		$input_id = esc_attr( $post_type );
 		$checked = checked( in_array( $post_type, $qv_post_type ), true, false );
 		$content .= '<div>';
 		$content .= sprintf( '<input id="%1$s" type="checkbox" name="post_type[]" value="%1$s" %2$s />', $input_id, $checked );
-		$content .= sprintf( '<label for="%s">%s</label>', $input_id, $label );
+		$content .= sprintf( '<label for="%s" tabindex="0">%s</label>', $input_id, $label );
```
ryelle: When I listen to this in VoiceOver, I'm told that I'm on a text element, so there's no indication that I can interact with this.

I think you want the checkbox to be visible to be in the tab order, and then you can use a sibling selector to style the label when an input is focused. That's the approach I used [in the query filters](https://github.com/WordPress/wporg-mu-plugins/blob/f6f13cf34e308c2c5553676909c61f1d8b7c271c/mu-plugins/blocks/query-filter/postcss/style.pcss#L128), at least.

source/wp-content/themes/wporg-developer-2023/src/search-filters/view.js
```diff
@@ -0,0 +1,41 @@
...
+			input.checked = false;
+		} );
+		updateAllButtonState();
+	} );
+
+	container.addEventListener( 'change', () => {
+		updateAllButtonState();
+	} );
+
+	container.addEventListener( 'keydown', ( event ) => {
```
ryelle: If you switch the label/checkbox behavior, you shouldn't need this either.

source/wp-content/themes/wporg-developer-2023/src/version-select/style.scss
```diff
@@ -16,24 +54,29 @@
...
-		height: 100%;
-		width: 0.8em;
-		mask-image: url(../../images/chevron.svg);
-		mask-repeat: no-repeat;
-		mask-position: center;
-		mask-size: contain;
-		transform: rotate(180deg);
-		background-color: var(--wp--preset--color--charcoal-1);
-		pointer-events: none; // Allows clicking on the select element.
+	button {
```
ryelle: Could you set up some helper classes in the parent theme for button styles? Those classes would be able to use the `button` mixins, and then just use that in these custom blocks. That could cut down on the CSS needed in the child themes.

source/wp-content/themes/wporg-developer-2023/functions.php
adamwoodnz: Yes! [Done](https://github.com/WordPress/wporg-developer/pull/292/commits/f7f5134e6e8429e008dfb87ae265f56a92a2b44a), thanks.

source/wp-content/themes/wporg-developer-2023/src/search-filters/view.js
adamwoodnz: Removed

source/wp-content/themes/wporg-developer-2023/src/search-filters/index.php
adamwoodnz: Yeah that's much better, [updated](https://github.com/WordPress/wporg-developer/pull/292/commits/8f015a3992f944554b6654ed1791e132a93a7dfb), thanks

source/wp-content/themes/wporg-developer-2023/src/version-select/style.scss
adamwoodnz: I've made a start on this and updated the testing notes above.

adamwoodnz: Parent PR is now ready for review.

source/wp-content/themes/wporg-developer-2023/src/search-filters/style.scss
```diff
@@ -1,43 +1,22 @@
...
+		border: 0;
+		clip: rect(1px, 1px, 1px, 1px);
+		clip-path: inset(50%);
+		height: 1px;
+		margin: -1px;
+		overflow: hidden;
+		padding: 0;
+		position: absolute;
+		top: 0;
+		left: 0;
```
ryelle, line 18: Can you add a position relative to the container element here? the toggle is at the top of the page, so if you're scrolled down a little tabbing into the checkbox moves the page back up. You can also see the strange highlight in VoiceOver.

Screenshot 2023-10-03 at 5 57 43 PM

ryelle (approved):
💯  the checkboxes are working well with VoiceOver & keyboard nav. I know the button PR would still block this, but everything here looks good now.

source/wp-content/themes/wporg-developer-2023/src/search-filters/style.scss
adamwoodnz: Added to the [toggle styles](https://github.com/WordPress/wporg-parent-2021/pull/112/files#diff-01daabce597632cf10fb204a48422d91c0dfa480fd49f22cb0324afdb0b32a5cR23)

PR #297, opened 2023-09-28
Update search results pages

### LINKED ISSUE #267: Search results update
New [design](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-Resources?type=design&node-id=2934-7608&mode=design&t=q9aehCWavRMKQ6KF-0):

| Common APIs | Code reference |
|-|-|
| Search results | Search results (1) |

adamwoodnz:
Closes #267 

1. Adds a resource selector in the top right. Selecting an option redirects to that resource search page with the current query preserved
2. Styles non - code reference results differently with a border and Inter font for title
3. Updates spacing
4. Updates breadcrumb, search content and pagination copy
5. Adds a search field to the single handbook template

### Screenshots

localhost_8888__s=block(Desktop) (2)

#### Code Reference

| Desktop | Tablet | Mobile |
|-|-|-|
| localhost_8888__s=block post_type%5B%5D=wp-parser-hook post_type%5B%5D=wp-parser-class(Desktop) | localhost_8888__s=block post_type%5B%5D=wp-parser-hook post_type%5B%5D=wp-parser-class(iPad) | localhost_8888__s=block post_type%5B%5D=wp-parser-hook post_type%5B%5D=wp-parser-class(Samsung Galaxy S20 Ultra) |

#### Common APIs (and everything else except Code Reference)

| Desktop | Tablet | Mobile |
|-|-|-|
| localhost_8888_apis__s=block(Desktop) (1) | localhost_8888_apis__s=block(iPad) (1) | localhost_8888_apis__s=block(Samsung Galaxy S20 Ultra) (1) |

### Testing

Search!
Switch resources, try filtering, try searches producing varying numbers of results (0, < 25, > 25,  hundreds)


source/wp-content/themes/wporg-developer-2023/patterns/search-content.php
```diff
@@ -7,41 +7,53 @@
...
-	<!-- wp:query-no-results -->
-<!-- wp:paragraph {"placeholder":"Add text or blocks that will display when a query returns no results.","style":{"spacing":{"margin":{"top":"var:preset|spacing|40"}}}} -->
-<p style="margin-top:var(--wp--preset--spacing--40)"><?php esc_attr_e( 'Sorry, but nothing matched your search terms.', 'wporg' ); ?></p>
-<!-- /wp:paragraph -->
-<!-- /wp:query-no-results --></div>
-<!-- /wp:group -->
+		<!-- wp:post-template -->
+			<?php
+			// Code reference search results at the root are styled differently.
+			if ( preg_match('#^/\?s=#', $_SERVER['REQUEST_URI']) ) { ?>
```
iandunn: Would `is_search()` work here?

iandunn: Or maybe something like:

```php
<?php global $post; ?>
<!-- wp:template-part {"slug":"search-content-<?php echo $post->post_type; ?>"} /-->
```

...and then name the template part files to match?

adamwoodnz: `$post` is null in here unfortunately.

adamwoodnz: I've now switched to [using a block to check if it's a parsed post type](https://github.com/WordPress/wporg-developer/pull/297/files#diff-75a145525c228b99a03ea331f3ae0c0e1cf6fc8487280677d761a749bfaf06e0R80). Keeps the pattern clean other than translations.

source/wp-content/themes/wporg-developer-2023/src/resource-select/index.php
```diff
@@ -0,0 +1,104 @@
...
+ * @param string   $content    Block default content.
+ * @param WP_Block $block      Block instance.
+ *
+ * @return string Returns the block markup.
+ */
+function render( $attributes, $content, $block ) {
+	global $wp_query;
+	$search_query = $wp_query->query_vars['s'];
+
+	$sources = array(
```
adamwoodnz, line 42: This list is probably not complete

source/wp-content/themes/wporg-developer-2023/src/resource-select/index.php
```diff
@@ -0,0 +1,112 @@
...
+ * @param string   $content    Block default content.
+ * @param WP_Block $block      Block instance.
+ *
+ * @return string Returns the block markup.
+ */
+function render( $attributes, $content, $block ) {
+	global $wp_query;
+	$search_query = $wp_query->query_vars['s'];
+
+	$resources = array(
```
adamwoodnz, line 42: Matches order on current [homepage](https://developer.wordpress.org/)

jasmussen:
Looks good. 

The only slightly off thing is the code reference dropdown, how it wraps on tablet and mobile:

Screenshot 2023-10-06 at 08 51 18

Screenshot 2023-10-06 at 08 51 29

The alignment (so the text inside the naked button lines up with the items above) is non-trivial to fix, and doesn't feel blocking. Showcase does the same:

Screenshot 2023-10-06 at 08 52 31

However the code reference dropdown itself looks different; it spans the full width, and uses the large chevron icon, instead of like the showcase filter bar as shown.

I recognize there are lots of tricky pieces in working across those sites at the same time, but just conceptually the filter component as shown in the Learn category mockups, with its first outing on the showcase, is meant to be the same component (give or take) across all sections. In case that helps the implementation.

adamwoodnz:
> However the code reference dropdown itself looks different; it spans the full width, and uses the large chevron icon, instead of like the showcase filter bar as shown.
> 
> I recognize there are lots of tricky pieces in working across those sites at the same time, but just conceptually the filter component as shown in the Learn category mockups, with its first outing on the showcase, is meant to be the same component (give or take) across all sections. In case that helps the implementation.

Short answer: In the short term I can update the chevron and full width behaviour 👍 

Longer answer to explain the different behaviour; the way search works on this site is that each handbook has it's own search page, eg. Code Reference at https://developer.wordpress.org/redesign-test/?s=block and Common APIs at https://developer.wordpress.org/redesign-test/apis/?s=block, so rather than filtering the results, this selector redirects the search to the appropriate page. I can see if the filter component can be adapted to fire a redirect event rather than adding a query parameter to the search, but I think this should be a later iteration, and for now I'll style the dropdown to better match, with the smaller chevron and correct spacing.

adamwoodnz:
Select styles updated to better match the Query Filter component, and new [issue](https://github.com/WordPress/wporg-developer/issues/301) created to follow up.

Issue #300, opened 2023-10-06
Missing indentation on code block


petitphp:
Initially reported in slack [#meta channel](https://wordpress.slack.com/archives/C02QB8GMM/p1695971223487939).

## Description
On the current [developer documentation](https://developer.wordpress.org/), some code blocks are missing indentation when displaying multiple lines (see attached screenshot).
Examples where this issue can be seen : [WP_Query reference](https://developer.wordpress.org/reference/classes/wp_query/#taxonomy-parameters) or [register_post_type](https://developer.wordpress.org/reference/functions/register_post_type/#flushing-rewrite-on-activation)

## Screenshot
![missing_spacing](https://github.com/WordPress/wporg-developer/assets/158572/46428c27-ec57-4b52-af88-208c57f60049)

adamwoodnz:
Unfortunately the examples given above aren't in the content on the staging site, eg. the https://developer.wordpress.org/redesign-test/reference/functions/register_post_type/#flushing-rewrite-on-activation section doesn't exist.

@StevenDufresne do you know if/how we can get the staging content to be on par with prod please?

StevenDufresne:
Running the cron jobs passing the`--url=developer.wordpress.org/redesign-test` should work.

adamwoodnz:
https://github.com/WordPress/wporg-developer/blob/1892b5001515d4521e34ca1de5161396d86f8dc0/env/setup.sh#L16

Docs: https://developer.wordpress.org/cli/commands/cron/event/run/

adamwoodnz:
Looks like we fixed something in the parsers because https://developer.wordpress.org/redesign-test/reference/functions/register_post_type/#flushing-rewrite-on-activation is there now, and indentation does still look broken


![Image](https://github.com/WordPress/wporg-developer/assets/1017872/f9494966-20ef-4271-b217-5c57e5bce4d2)

outdoor2kode:
> Looks like we fixed something in the 

How it was fixed was mentioned in this [PR description](https://github.com/WordPress/wporg-developer/pull/369#issue-1992603604) Note I Section, if anyone is interested in it, please take a look.

outdoor2kode:
Summary of the issue:

[...]

PR #308, opened 2023-10-30
Replace `jetpack_require_lib` with a direct file include

adamwoodnz:
I noticed locally that the docs weren't being imported properly, and I was seeing warnings about `jetpack_require_lib() is missing on system.`, however looking at the source this is actually an error, and [`jetpack_require_lib` has been deprecated and removed](https://github.com/Automattic/jetpack/pull/28866).

This PR implements the [suggested replacement](https://github.com/Automattic/jetpack/pull/28866#issuecomment-1548467922) following what has been done in [`wporg-markdown`](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/plugins/wporg-markdown/inc/class-importer.php#L423).

Also similar to https://github.com/WordPress/wordcamp.org/commit/ca0b71b4a53b1ac44dca4c49c8fa9f5d0cfb63cf

### Testing

Run `yarn setup:wp` and check for warnings about `jetpack_require_lib`
Ensure the CLI import completes successfully with something like:
```
Success: Successfully updated 414 of 418 CLI command pages.
Executed the cron event 'devhub_cli_markdown_import' in 180.088s.
```

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/inc/cli.php
+++ b/source/wp-content/themes/wporg-developer-2023/inc/cli.php
@@ -207,12 +207,14 @@ class DevHub_CLI {
 		if ( is_wp_error( $markdown_source ) ) {
 			return $markdown_source;
 		}
-		
-		if ( ! class_exists( 'WPCom_GHF_Markdown_Parser' ) && defined( 'JETPACK__PLUGIN_DIR' ) ) {
-			include JETPACK__PLUGIN_DIR . '/_inc/lib/markdown.php';
-		}
+
+		// Load Jetpack Markdown if it's not already loaded, for transforming markdown to HTML.
 		if ( ! class_exists( 'WPCom_GHF_Markdown_Parser' ) ) {
-			return new WP_Error( 'missing-jetpack-markdown', 'Jetpack Markdown is missing on system.' );
+			if ( defined( 'JETPACK__PLUGIN_DIR' ) ) {
+				require_once JETPACK__PLUGIN_DIR . '/_inc/lib/markdown.php';
+			} else {
+				return new WP_Error( 'missing-jetpack-markdown', 'Jetpack Markdown is missing on system.' );
+			}
 		}
 
 		// Transform GitHub repo HTML pages into their raw equivalents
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-developer-2023/inc/cli.php
```diff
@@ -207,8 +207,12 @@ private static function update_post_from_markdown_source( $post_id ) {
 		if ( is_wp_error( $markdown_source ) ) {
 			return $markdown_source;
 		}
-		if ( ! function_exists( 'jetpack_require_lib' ) ) {
-			return new WP_Error( 'missing-jetpack-require-lib', 'jetpack_require_lib() is missing on system.' );
+		
+		if ( ! class_exists( 'WPCom_GHF_Markdown_Parser' ) && defined( 'JETPACK__PLUGIN_DIR' ) ) {
```
pkevan: I can help feel that combining the two `if ( ! class_exists( 'WPCom_GHF_Markdown_Parser' )` would be better here, particularly to avoid any future issue whereby we either do or don't need to check that class_exists statement.

pkevan: Maybe something like this:
```
// Comment saying why we are doing this.
if ( ! class_exists( 'WPCom_GHF_Markdown_Parser' ) {
    if ( defined( 'JETPACK__PLUGIN_DIR' ) ) {
        include JETPACK__PLUGIN_DIR . '/_inc/lib/markdown.php';
    } else {
        return new WP_Error( 'missing-jetpack-markdown', 'Jetpack Markdown is missing on system.' );
    }
}
```

adamwoodnz: Updated 🙂

StevenDufresne:
Relevant PR: https://github.com/WordPress/wordcamp.org/pull/859

StevenDufresne:
Does this fix the [missing data problem](https://github.com/WordPress/wporg-developer/issues/300#issuecomment-1780375363)?

adamwoodnz:
> Does this fix the [missing data problem](https://github.com/WordPress/wporg-developer/issues/300#issuecomment-1780375363)?

It only relates to the cli handbook, but it may have been blocking other jobs from completing, not sure

pkevan (approved):

PR #334, opened 2023-11-06
Update header behaviour

### LINKED ISSUE #298: Fix sticky local nav and breadcrumbs after global header update
When https://github.com/WordPress/wporg-mu-plugins/pull/466 merges, it will unstick the global header and only stick the local nav (and breadcrumbs). This will affect the way the local navigation bar stickiness works on the Developers site, because it uses the `is-sticky` helper class. Instead, it can now use `"position":{"type":"sticky"}` to use the gutenberg sticky behavior.

Screen Shot 2023-10-04 at 16 06 31

See https://github.com/WordPress/wporg-documentation-2022/pull/71 for a similar fix on documentation.

adamwoodnz:
Fixes #298 
Depends on https://github.com/WordPress/wporg-mu-plugins/pull/466

Very similar to [changes required for Documentation](https://github.com/WordPress/wporg-documentation-2022/pull/71) after the global header becomes non-sticky:

> Other changes I made:
> 
> * Removing the padding on `local-navigation-bar`, this is set correctly by default now
> * Move the border from `local-navigation-bar` to the global header, this looks better when the page scrolls
> * Since the `local-navigation-bar` was moved out of the group, I moved the `style` attribute to `local-navigation-bar` directly

With these changes **only the local nav is sticky** (global header, breadcrumbs and search are not).

### Screenshots

| Page | Top | Scrolled |
|-|-|-|
| Home | localhost_8888_(Desktop) (10) | localhost_8888_(Desktop) (11) |
| With breadcrumbs | localhost_8888_coding-standards_wordpress-coding-standards_css_(Desktop) | localhost_8888_coding-standards_wordpress-coding-standards_css_(Desktop) (1) |
| Without breadcrumbs | localhost_8888_cli_commands_(Desktop) (4) | localhost_8888_cli_commands_(Desktop) (5) |

### Testing 

1. Ensure your mu-plugins is on https://github.com/WordPress/wporg-mu-plugins/pull/466
2. Navigate to the pages in the screenshots above at desktop size, and check the header behaviour.
3. Resize for tablet and mobile; no headers should be sticky below 890px (tablet landscape will still have sticky local nav).

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/parts/header-no-breadcrumbs.html
+++ b/source/wp-content/themes/wporg-developer-2023/parts/header-no-breadcrumbs.html
@@ -1,9 +1,3 @@
-<!-- wp:wporg/global-header /-->
+<!-- wp:wporg/global-header {"style":{"border":{"bottom":{"color":"var:preset|color|white-opacity-15","style":"solid","width":"1px"}}}} /-->
 
-<!-- wp:group {"align":"full","style":{"spacing":{"blockGap":"0"}},"backgroundColor":"charcoal-2","className":"is-sticky","layout":{"type":"constrained"}} -->
-<div class="wp-block-group alignfull is-sticky has-charcoal-2-background-color has-background">
-
-	<!-- wp:template-part {"slug":"local-nav","className":"has-display-contents"} /-->
-
-</div>
-<!-- /wp:group -->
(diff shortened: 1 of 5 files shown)


outdoor2kode (approved):
Behaves as described. LGTM 👍

Issue #312, opened 2023-10-31
Spacing and style in search bar


fcoveram:
As described [here](https://github.com/WordPress/wporg-developer/pull/305#issuecomment-1784796554), the search bar has new changes:

## Style
* In Article post-type, the search bar always have a border bottom of 1px in Light Grey
* When search bar is below the readcrumb, padding-top is 10px. When is not, padding-top is 20px.

![New design of search bar area in Article](https://github.com/WordPress/wporg-developer/assets/895819/9d9ad89e-b36b-4ec3-8769-c088ef97f51a)

---

## Design
[Developer mockups](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-Resources?type=design&node-id=2934-6590&mode=design&t=7w1GtutWNO2sHH72-4). Updated with these changes.

adamwoodnz:
@fcoveram so is the search bar sticky, and is it only on this page?

Currently all parts of the header are sticky, so this addition will result in a large chunk of the page being taken up. Have we decided if that's to change?

fcoveram:
> the search bar sticky, and is it only on this page?

No sticky search bar. The only sticky component in this page is the table of content.

I suggested [here](https://github.com/WordPress/wporg-mu-plugins/issues/465#issuecomment-1754693237) to only keep the local nav fixed, as in [the example](https://user-images.githubusercontent.com/541093/271435777-aa35957e-87d6-4e81-b0d2-15c8c12d9bd9.mp4) shared by @ryelle. But not sure whether that's being addressed or not.

adamwoodnz:
> No sticky search bar. The only sticky component in this page is the table of content.

Ah ok. [This design](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-Resources?node-id=3735%3A19404&mode=dev) seems to contradict that, it shows header, chapter list and ToC sticky, unless I'm mistaken.

> I suggested [here](https://github.com/WordPress/wporg-mu-plugins/issues/465#issuecomment-1754693237) to only keep the local nav fixed, as in [the example](https://user-images.githubusercontent.com/541093/271435777-aa35957e-87d6-4e81-b0d2-15c8c12d9bd9.mp4) shared by @ryelle. But not sure whether that's being addressed or not.

Yeah I think maybe we need to get the header changes shipped before we can move forward with Developer and Documentation.

jasmussen:
[...]

PR #336, opened 2023-11-06
Add breadcrumbs to cli/commands.

### LINKED ISSUE #327: CLI: Breadcrumb missing.
The breadcrumb is missing from the WP-CLI handbook.

## Screenshot

| Current | Staging |
|--------|--------|
| Screenshot 2023-11-03 at 2 46 08 PM | Screenshot 2023-11-03 at 2 45 05 PM |

StevenDufresne:
This PR adds the breadcrumbs to `cli/commands`. 

I realized after logging an issue that the [designs](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-Resources?type=design&node-id=3745-10313&mode=design) don't have the breadcrumb. However, I don't understand why this page would not have it when the block-editor handbook _landing page_ and other landing pages have it. 
It's also linked in deeper nested pages as the parent and last [production](https://developer.wordpress.org/cli/commands/) currently has the breadcrumb there.

## Screenshot
localhost_8888_cli_commands_ (1)

## Nested Page
localhost_8888_cli_commands_admin_

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/templates/archive-command.html
+++ b/source/wp-content/themes/wporg-developer-2023/templates/archive-command.html
@@ -1,4 +1,4 @@
-<!-- wp:template-part {"slug":"header-no-breadcrumbs","className":"has-display-contents"} /-->
+<!-- wp:template-part {"slug":"header","className":"has-display-contents"} /-->
 
 <!-- wp:template-part {"slug":"search","className":"has-display-contents"} /-->
 
--- a/source/wp-content/themes/wporg-developer-2023/functions.php
+++ b/source/wp-content/themes/wporg-developer-2023/functions.php
@@ -228,7 +228,6 @@ function init() {
 
 	// Modify default breadcrumbs.
 	add_filter( 'breadcrumb_trail_items', __NAMESPACE__ . '\\breadcrumb_trail_items_for_hooks', 10, 2 );
-	add_filter( 'breadcrumb_trail_items', __NAMESPACE__ . '\\breadcrumb_trail_items_remove_reference', 11, 2 );
 	add_filter( 'breadcrumb_trail_items', __NAMESPACE__ . '\\breadcrumb_trail_items_for_handbook_root', 10, 2 );
 	add_filter( 'breadcrumb_trail_items', __NAMESPACE__ . '\\breadcrumb_trail_for_note_edit', 10, 2 );
 	add_filter( 'breadcrumb_trail_items', __NAMESPACE__ . '\\breadcrumb_trail_for_since_view', 10, 2 );
@@ -281,29 +280,6 @@ function breadcrumb_trail_items_for_hooks( $items, $args ) {
 	unset( $items[4] );
 
 	return $items;
-}
-
-/**
- * Remove the 'Reference' part of the breadcrumb trail.
- *
- * @param  array $items The breadcrumb trail items.
- * @param  array $args  Original args.
- * @return array
- */
-function breadcrumb_trail_items_remove_reference( $items, $args ) {
-	if ( ! is_parsed_post_type() ) {
-		return $items;
-	}
-
-	return array_filter(
-		$items,
-		function( $item ) {
-			// Remove the 'reference' parent based on the presence of its URL.
-			// We can't use the label because of internationalization.
-			$result = (bool) preg_match( '!href="[^"]+/reference/"!', $item );
-			return ( false === $result );
-		}
-	);
 }
 
 /**
--- a/source/wp-content/themes/wporg-developer-2023/parts/header-no-breadcrumbs.html
+++ b/source/wp-content/themes/wporg-developer-2023/parts/header-no-breadcrumbs.html
@@ -1,3 +1,0 @@
-<!-- wp:wporg/global-header {"style":{"border":{"bottom":{"color":"var:preset|color|white-opacity-15","style":"solid","width":"1px"}}}} /-->
-
-<!-- wp:template-part {"slug":"local-nav","className":"has-display-contents"} /-->
--- a/source/wp-content/themes/wporg-developer-2023/parts/header.html
+++ b/source/wp-content/themes/wporg-developer-2023/parts/header.html
@@ -1,6 +1,13 @@
 <!-- wp:wporg/global-header {"style":{"border":{"bottom":{"color":"var:preset|color|white-opacity-15","style":"solid","width":"1px"}}}} /-->
 
-<!-- wp:template-part {"slug":"local-nav","className":"has-display-contents"} /-->
+<!-- wp:wporg/local-navigation-bar {"className":"has-display-contents","backgroundColor":"charcoal-2","style":{"elements":{"link":{"color":{"text":"var:preset|color|white"},":hover":{"color":{"text":"var:preset|color|white"}}}}},"textColor":"white","fontSize":"small"} -->
+
+	<!-- wp:site-title {"level":0,"fontSize":"small"} /-->
+
+	<!-- wp:navigation {"icon":"menu","overlayBackgroundColor":"charcoal-2","overlayTextColor":"white","layout":{"type":"flex","orientation":"horizontal"},"fontSize":"small","menuSlug":"developer"} /-->
+		
+<!-- /wp:wporg/local-navigation-bar -->
+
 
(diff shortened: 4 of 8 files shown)


StevenDufresne:
@ndiego Thoughts?

adamwoodnz:
If we do add this we'll need to adjust the vertical position of the ToC; it should align with the title.

StevenDufresne:
I'm getting no table of contents on the [staging env](https://developer.wordpress.org/redesign-test/cli/commands/) (well, i'm getting the back to top).

Screenshot 2023-11-06 at 2 59 39 PM

StevenDufresne:
Also, if we implement https://github.com/WordPress/wporg-developer/issues/312, then I imagine it will just work...?

adamwoodnz:
> Also, if we implement #312, then I imagine it will just work...?

That search bar with bottom border is only for handbook pages if I understand it correctly.

fcoveram:
> …I don't understand why this page would not have it when the block-editor handbook landing page and other landing pages have it.

Breadcrumbs appear from the third level and beyond (Home / Level 2 / Level 3). The idea was to avoid duplicative content as in level 2's pages local nav shows the current page in bold. So in the [nested page example](https://user-images.githubusercontent.com/1657336/280603800-4aa3eab4-74d4-4449-8b0a-bdf1b8333716.png), it's correct to show it.

> That search bar with bottom border is only for handbook pages if I understand it correctly.

~~Yes~~. ← Updating this response as [this comment](https://github.com/WordPress/wporg-developer/issues/341#issuecomment-1794462249) informs about an update of this component. The bottom border in Articles should be removed.

adamwoodnz:
> So in the [nested page example](https://user-images.githubusercontent.com/1657336/280603800-4aa3eab4-74d4-4449-8b0a-bdf1b8333716.png), it's correct to show it.

Ah ok, so I think we only need to add this to single-command.html

StevenDufresne:
> Breadcrumbs appear from the third level and beyond (Home / Level 2 / Level 3). The idea was to avoid duplicative content as in level 2's pages local nav shows the current page in bold. So in the [nested page example](https://user-images.githubusercontent.com/1657336/280603800-4aa3eab4-74d4-4449-8b0a-bdf1b8333716.png), it's correct to show it.

@fcoveram Does that mean it shouldn't be in the [block editor handbook](https://developer.wordpress.org/redesign-test/block-editor/)?

Screenshot 2023-11-07 at 2 22 54 PM

fcoveram:
> @fcoveram Does that mean it shouldn't be in the [block editor handbook](https://developer.wordpress.org/redesign-test/block-editor/)?

Exactly.

To assess the change impact, which site sections will have this change? I'm thinking of Developers, Showcase, About WordPress, and News.

The idea of showing it from level 3 is to diminish location redundancy. But in the case of Articles, it feels weird landing on its main page without breadcrumb and then seeing it in all internal pages.

outdoor2kode:
Summarizing the breadcrumb behaviors I've observed while working on another ticket.

**Handbook** ✅
Only put two here. All handbooks have the same look.
image

| level2 | level3 |
|-|-|
| Screenshot 2023-11-10 at 12 42 30 AM | Screenshot 2023-11-10 at 12 42 37 AM | 
| Screenshot 2023-11-10 at 12 43 21 AM | Screenshot 2023-11-10 at 12 43 28 AM |

**Resource** ✅
If clicking on Resource in the breadcrumbs, it would redirect back to Dashicons.
Same in production, visiting Resource page would redirect to Dashicons page.
And there's no breadcrumbs on production, but there's a title though.

| local/staging | production |
|-|-|
| Screenshot 2023-11-10 at 12 45 28 AM | Screenshot 2023-11-10 at 12 45 35 AM |

**Reference** ✅

| local/staging | production |
|-|-|
| Screenshot 2023-11-10 at 12 46 02 AM | Screenshot 2023-11-10 at 12 46 08 AM |

**Functions**  ✅
Only put Functions CPT here. Other CPTs have the same look.
image

In production, it has 4 levels.
We don't need the 'Reference' level? => **Done Adding back.**

| | level2 | level3 |
|-|-|-|
| local/staging | Screenshot 2023-11-10 at 12 46 58 AM | Screenshot 2023-11-10 at 12 47 25 AM |
| production | Screenshot 2023-11-10 at 12 47 06 AM | Screenshot 2023-11-10 at 1 31 58 AM |

**Taxonomy** ✅
There are overall three taxonomies - since, packages, files
image

Same here, in production, it has 4 levels.
We don't need the 'Reference' level? =>  **Done Adding back.** ✅

| local/staging, files | production |
|-|-|
| Screenshot 2023-11-10 at 12 47 59 AM | Screenshot 2023-11-10 at 12 48 07 AM |

There's breadcrumbs on packages page, but isn't on since page. => **Done Adding back.** ✅

| packages | since |
|-|-|
| Screenshot 2023-11-10 at 12 48 21 AM | Screenshot 2023-11-10 at 1 45 06 AM |

The files and since pages can be accessed through the links as indicated below.
Not sure how to access the packages page.

image

**CLI/Commands** ✅

I guess this is the most standard one in terms of design.
It only renders breadcrumbs when it's level 3.
It also makes sense not to include Reference in breadcrumbs on CLI/Commands page because there's a button on homepage linking to the page.

| | level2 | level3 |
|-|-|-|
| local/staging | Screenshot 2023-11-10 at 12 54 49 AM | Screenshot 2023-11-10 at 12 55 07 AM |
| production | Screenshot 2023-11-10 at 12 54 58 AM | Screenshot 2023-11-10 at 12 55 11 AM |

outdoor2kode:
Currently, most of the pages on developer sites display breadcrumbs at 2 levels. 
And to be more specific, [CLI-Commands page](https://developer.wordpress.org/redesign-test/cli/commands/) is the only one that doesn’t display at 2 levels.

And after chatting with @fcoveram, we agree that it makes sense to display the breadcrumbs from `Home / Level 1`. This is also what we have in Showcase for the site details pages.

Note: In our discussion, it was mentioned that how we describe the levels is slightly off. 
All the comments above are labeled as `Home / Level 2 / Level 3`, but they should actually be `Home / Level 1 / Level 2`.

outdoor2kode:
@adamwoodnz 

I have already added the pages missing the `Home/Level1` breadcrumbs. 
To test, you can refer to the comments above, where I have marked them with ✅.

Additionally, I have temporarily re-added the 'reference' to the breadcrumbs for now. 
Without 'reference', the process becomes a bit odd, as users won't be able to navigate back to the reference page from any CPT pages. 

Screenshot 2023-11-29 at 1 37 46 AM

~However, [this PR](https://github.com/WordPress/wporg-developer/pull/166) mentions the original reason for its removal. ([Ref](https://github.com/WordPress/wporg-developer/pull/166/files#diff-1cf9dd600383a4c89ba09a878ca8691359b6f6dcece2b9c9b90f3de810f274a4R234))~

~@StevenDufresne, is this still a concern? Is the reason we need to remove the 'reference' from the breadcrumbs due to the i18n issue, such as the breadcrumbs being too long?~

**Update**
According to Steve's feedback,
It was removed because the page was gone.
Now it can be brought back.

fcoveram:
Thanks @renintw for addressing this.

> The idea was to avoid duplicative content as in level 1’s pages local nav shows the current page in bold.

That statement shared by me days ago is slightly incorrect. There are L1 pages not displayed in the local nav menu, so displaying them is safer than not. That's the rationale behind this decision.

source/wp-content/themes/wporg-developer-2023/templates/archive-command.html
```diff
@@ -1,4 +1,4 @@
-<!-- wp:template-part {"slug":"header-no-breadcrumbs","className":"has-display-contents"} /-->
```
adamwoodnz, line 1: Looks like this template part `header-no-breadcrumbs.html` is unused after these changes and can be removed.

Also looks like the `local-nav.html` template part, which was used in both header parts, can be moved inline within `header.html` and the template part removed.

adamwoodnz (changes requested):
Nice work, I like how this simplifies the template setup.

A couple of comments inline.

One more adjustment required now that the CLI Command archive has breadcrumbs; we need to remove this custom ToC spacing rule: https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer-2023/src/style/style.scss#L407

Removing that should fix the misaligned ToC and H1:

Screenshot 2023-11-30 at 10 06 48 AM

source/wp-content/themes/wporg-developer-2023/templates/archive-command.html
outdoor2kode: Thanks. Done!

outdoor2kode:
## Updated Screenshot

image

adamwoodnz (approved):
LGTM :shipit:

PR #337, opened 2023-11-06
Add the ability to pass an id to the `code-table`.

### LINKED ISSUE #324: Code Reference: Anchor tags not working for "Used by" or "Uses"
From the [search results](https://developer.wordpress.org/redesign-test/?s=hook&post_type%5B%5D=wp-parser-function&post_type%5B%5D=wp-parser-hook&post_type%5B%5D=wp-parser-class), we link to the function references. When I click through, it doesn't activate the anchor and bring me to the right spot.

Screenshot 2023-11-03 at 2 32 58 PM

## Steps to reproduce
- Starts on a [search result](https://developer.wordpress.org/redesign-test/?s=hook&post_type%5B%5D=wp-parser-function&post_type%5B%5D=wp-parser-hook&post_type%5B%5D=wp-parser-class).
- Click on `Uses 1 function`. 
- Expect to start halfway down the page where the "Used By" or "Uses" section is.
- It doesn't.

StevenDufresne:
Fixes #324.

This PR adds the ability to pass an id for the table which can be used for `#{id}` links which is used by the "Uses" and "Used by" tables.

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/code-table/block.php
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-table/block.php
@@ -26,13 +26,7 @@ function init() {
  */
 function render( $attributes ) {
 	$table  = '<!-- wp:table {"className":"' . $attributes['className'] . '"} --><figure class="wp-block-table ' . $attributes['className'] . '">';
-
-	if ( isset( $attributes['id'] ) ) {
-		$table .= '<table id="' . esc_html( $attributes['id'] ) . '">';
-	} else {
-		$table .= '<table>';
-	}
-
+	$table .= '<table>';
 	$table .= '<thead>';
 	$table .= '<tr>';
 	foreach ( $attributes['headings'] as $heading ) {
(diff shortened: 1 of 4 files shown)


source/wp-content/themes/wporg-developer-2023/src/code-table/block.php
```diff
@@ -26,7 +26,13 @@ function init() {
  */
 function render( $attributes ) {
 	$table  = '<!-- wp:table {"className":"' . $attributes['className'] . '"} --><figure class="wp-block-table ' . $attributes['className'] . '">';
-	$table .= '<table>';
+
+	if ( isset( $attributes['id'] ) ) {
+		$table .= '<table id="' . esc_html( $attributes['id'] ) . '">';
```
adamwoodnz:
```suggestion
		$table .= '<table id="' . esc_attr( $attributes['id'] ) . '">';
```
?

adamwoodnz (approved):
Works for me after a small change to adjust the scroll position, thanks!

PR #346, opened 2023-11-07
Remove title edit links from handbooks

adamwoodnz:
Removes the GitHub edit links from the titles on handbook pages. Achieved by making the `WordPressdotorg\Markdown\Editor` instances available globally so that the filter can be removed.

### Screenshots

| Before | After |
|-|-|
| developer wordpress org_redesign-test_block-editor_(Desktop) | developer wordpress org_redesign-test_block-editor_(Desktop) (1) |

### Testing

Easiest in a sandbox. The edit links don't show for me locally.

Open each of the GitHub handbooks and check that the edit links don't appear: [Best Practices](https://developer.wordpress.org/redesign-test/coding-standards/), [Block Editor Handbook](https://developer.wordpress.org/redesign-test/block-editor/), [Rest API](https://developer.wordpress.org/redesign-test/rest-api/), [Adv Admin](https://developer.wordpress.org/redesign-test/advanced-administration/)

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/functions.php
+++ b/source/wp-content/themes/wporg-developer-2023/functions.php
@@ -198,9 +198,14 @@ add_filter( 'wporg_table_of_contents_post_content', __NAMESPACE__ . '\filter_cod
 add_filter( 'the_content', __NAMESPACE__ . '\filter_command_content', 4 );
 add_filter( 'wporg_table_of_contents_post_content', __NAMESPACE__ . '\filter_command_content' );
 
-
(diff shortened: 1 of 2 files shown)


StevenDufresne:
I'm okay with this approach, although does it still show up anywhere?

adamwoodnz:
> I'm okay with this approach, although does it still show up anywhere?

Not on this site as far as I can see. My assumption is that other sites use `WordPressdotorg\Markdown` so the filter can't be removed entirely. Definitely could be mistaken 🙂

PR #351, opened 2023-11-08
Fix searchbox autocomplete missing

### LINKED ISSUE #321: Searchbox Autocomplete missing.
The autocomplete for searches when using the code reference is not working. It was working when we did a first pass of this design. Was it removed purposefully? That is a very well used feature.

## Screenshots
Screenshot 2023-11-02 at 3 55 10 PM

should be:
Screenshot 2023-11-02 at 3 54 28 PM

### LINKED ISSUE #360: Search bar: The input should extend longer and the magnifying glass should be aligned to the left.
In testing: https://github.com/WordPress/wporg-developer/pull/351, I noticed the search bar layout is not correct.

## Screenshot
Screenshot 2023-11-10 at 10 13 43 AM

outdoor2kode:
Fixes #321

| **Page** |
|-|
| Reference |
| image |
| CLI-Commands |
| image |
| Single Page |
| image |
| Search |
| image |

| **Taxonomy** |
|-|
| **Since** |
| image |

| **Remain no autocomplete on the handbook** |
|-|
| image |

## TODO
**missing search bar** -> ✅ (See https://github.com/WordPress/wporg-developer/pull/359)
- [archive](http://localhost:8888/reference/functions/) 
- [tax: file](http://localhost:8888/reference/files/wp-activate.php/)
- [tax: package](http://localhost:8888/reference/package/atomlib/)

**missing breadcrumbs** -> summarized [here](https://github.com/WordPress/wporg-developer/pull/336#issuecomment-1804227066)
- http://localhost:8888/cli/commands/

**fix search bar style** -> Found ticket: https://github.com/WordPress/wporg-developer/pull/340
@adamwoodnz I guess the fix doesn't include the search bars outside handbooks? (margin -> see https://github.com/WordPress/wporg-developer/pull/359, quirky right padding -> img below, etc)
image

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/inc/autocomplete.php
+++ b/source/wp-content/themes/wporg-developer-2023/inc/autocomplete.php
@@ -68,8 +68,9 @@ class DevHub_Search_Form_Autocomplete {
 			'autocomplete',
 			'autocomplete',
 			array(
-				'ajaxurl' => admin_url( 'admin-ajax.php' ),
-				'nonce'   => wp_create_nonce( 'autocomplete_nonce' ),
+				'ajaxurl'   => admin_url( 'admin-ajax.php' ),
+				'nonce'     => wp_create_nonce( 'autocomplete_nonce' ),
+				'post_type' => get_post_type(),
 			)
 		);
 
@@ -91,7 +92,6 @@ class DevHub_Search_Form_Autocomplete {
 		$parser_post_types = DevHub\get_parsed_post_types();
 		$defaults          = array(
 			's'         => '',
-			'post_type' => $parser_post_types,
 			'posts'     => array(),
 		);
 
@@ -108,13 +108,9 @@ class DevHub_Search_Form_Autocomplete {
 			wp_send_json_error( $defaults );
 		}
 
-		foreach ( $form_data['post_type'] as $key => $post_type ) {
-			if ( ! in_array( $post_type, $parser_post_types ) ) {
-				unset( $form_data['post_type'][ $key ] );
-			}
-		}
-
-		$post_types = ! empty( $form_data['post_type'] ) ? $form_data['post_type'] : $parser_post_types;
+		$post_types = isset( $_POST['post_type'] ) && 'command' === $_POST['post_type'] ?
+			array( 'command' ) :
+			$parser_post_types;
 
 		$args = array(
 			'posts_per_page'       => -1,
--- a/source/wp-content/themes/wporg-developer-2023/js/autocomplete.js
+++ b/source/wp-content/themes/wporg-developer-2023/js/autocomplete.js
@@ -11,7 +11,7 @@
 		return;
 	}
 
-	var form = $( '.awesomeplete-form-wrap > form' );
+	var form = $( 'main form.wp-block-search' );
 	if ( ! form.length ) {
 		return;
 	}
--- a/source/wp-content/themes/wporg-developer-2023/js/autocomplete.js
+++ b/source/wp-content/themes/wporg-developer-2023/js/autocomplete.js
@@ -96,6 +96,7 @@
 			action: "autocomplete",
 			data: form.serialize(),
 			nonce: autocomplete.nonce,
+			post_type: autocomplete.post_type
 		};
 
 		$.post( autocomplete.ajaxurl, data )
--- a/source/wp-content/themes/wporg-developer-2023/inc/template-tags.php
+++ b/source/wp-content/themes/wporg-developer-2023/inc/template-tags.php
@@ -1765,8 +1765,7 @@ namespace DevHub {
 		$post_types = get_parsed_post_types();
 		$taxonomies = array( 'wp-parser-since', 'wp-parser-package', 'wp-parser-source-file' );
 
-		return ! ( is_search() || is_404() ) &&
-			( is_singular( $post_types ) || is_post_type_archive( $post_types ) || is_tax( $taxonomies ) || is_page( 'reference' ) || is_front_page() );
+		return ! ( is_search() || is_404() ) && ( is_singular( $post_types ) || is_post_type_archive( $post_types ) || is_tax( $taxonomies ) || is_page( 'reference' ) );
 	}
 
 	/**
--- a/source/wp-content/themes/wporg-developer-2023/inc/template-tags.php
+++ b/source/wp-content/themes/wporg-developer-2023/inc/template-tags.php
@@ -1765,7 +1765,8 @@ namespace DevHub {
 		$post_types = get_parsed_post_types();
 		$taxonomies = array( 'wp-parser-since', 'wp-parser-package', 'wp-parser-source-file' );
 
-		return ! ( is_search() || is_404() ) && ( is_singular( $post_types ) || is_post_type_archive( $post_types ) || is_tax( $taxonomies ) || is_page( 'reference' ) );
+		return ! ( is_search() || is_404() ) &&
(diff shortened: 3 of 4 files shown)


source/wp-content/themes/wporg-developer-2023/js/autocomplete.js
```diff
@@ -11,7 +11,7 @@
 		return;
 	}
 
-	var form = $( '.awesomeplete-form-wrap > form' );
+	var form = $( 'main form.wp-block-search' );
```
outdoor2kode, line 14: https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer-2023/inc/autocomplete.php#L40 already takes care of handbook searchbar

StevenDufresne (approved):

adamwoodnz (changes requested):
Great work fixing this!

I don't think autocomplete should be enabled on [WP CLI commands](http://localhost:8888/cli/commands/), because it only includes code reference content from what I can see. I believe the expectation when you search on that page is that you're searching commands, which is what happens if you submit the search and end up on that search page, eg. http://localhost:8888/cli/commands/?s=wp+admin

adamwoodnz:
It looks like the fix for https://github.com/WordPress/wporg-developer/issues/360 is fairly simple, so you could include it in this PR. Something like:

```css
.wp-block-search__inside-wrapper .awesomeplete {
  flex: 1;

  .wp-block-search__input {
    width: 100%;
  }
}
```

outdoor2kode:
> I don't think autocomplete should be enabled on [WP CLI commands](http://localhost:8888/cli/commands/), because it only includes code reference content from what I can see.

Thanks, It's a nice catch!
I've adjusted it so that it only displays commands, while maintaining the code reference content on other custom post type pages.

**WP CLI commands**
image

**Other CPT**
image

outdoor2kode:
Broken search bar style fixed:

image

source/wp-content/themes/wporg-developer-2023/inc/autocomplete.php
```diff
@@ -108,13 +108,9 @@ public function autocomplete_data_update() {
...
-		foreach ( $form_data['post_type'] as $key => $post_type ) {
-			if ( ! in_array( $post_type, $parser_post_types ) ) {
-				unset( $form_data['post_type'][ $key ] );
-			}
-		}
-
-		$post_types = ! empty( $form_data['post_type'] ) ? $form_data['post_type'] : $parser_post_types;
+		$post_types = isset( $_POST['post_type'] ) && 'command' === $_POST['post_type'] ?
+			array( 'command' ) :
+			$parser_post_types;
```
adamwoodnz, line 113: Nitpick but I think this format (typically enforced by eslint) is clearer
```suggestion
		$post_types = isset( $_POST['post_type'] ) && 'command' === $_POST['post_type']
		        ? array( 'command' )
		        : $parser_post_types;
```

source/wp-content/themes/wporg-developer-2023/inc/autocomplete.php
```diff
@@ -108,13 +108,9 @@ public function autocomplete_data_update() {
 			wp_send_json_error( $defaults );
 		}
 
-		foreach ( $form_data['post_type'] as $key => $post_type ) {
-			if ( ! in_array( $post_type, $parser_post_types ) ) {
-				unset( $form_data['post_type'][ $key ] );
-			}
-		}
-
-		$post_types = ! empty( $form_data['post_type'] ) ? $form_data['post_type'] : $parser_post_types;
```
adamwoodnz, line 117: Removing this logic is a bit concerning. It looks like line 117 was intended to do what we want... is `$form_data['post_type']` not `command` like expected? Or is it stripped by line 113 because it's not in `$parser_post_types`?

adamwoodnz (commented):
This is working for me, left a question about the logic change though.

source/wp-content/themes/wporg-developer-2023/inc/autocomplete.php
outdoor2kode: The code here is a bit tricky and confusing.
1. Originally it used  `$form_data['post_type']`, but according to [this line](https://github.com/WordPress/wporg-developer/pull/351/files/b0a77cfda6d111fcd46895648109cfa476f137c2#diff-9133227a1757e0f7ff2996d7bc952b9db0445049a6984bbd91830d7c6107e43aL104) and [this line](https://github.com/WordPress/wporg-developer/pull/351/files/b0a77cfda6d111fcd46895648109cfa476f137c2#diff-5e0d1972031d52467147da1a95b1eaf3929b263e66640ad2b369130ab4789ef5R97),  `$form_data` is likely to be `$parser_post_types` all the time. (coming from [L94](https://github.com/WordPress/wporg-developer/pull/351/files/b0a77cfda6d111fcd46895648109cfa476f137c2#diff-9133227a1757e0f7ff2996d7bc952b9db0445049a6984bbd91830d7c6107e43aL94))
2. So I was thinking of adding `command` to `get_parsed_post_types()`, but it would affect quite a few other parts of the code base.
3. and since `$form_data['post_type']` is always `$parser_post_types` (or is there a case that a user needs to pass in the `post_type` through a form? I don't think this is common for search bar), this check seems extra and `$post_types` in line 117 can just be set as `$parser_post_types` (`post_type` in [defaults](https://github.com/WordPress/wporg-developer/pull/351/files/b0a77cfda6d111fcd46895648109cfa476f137c2#diff-9133227a1757e0f7ff2996d7bc952b9db0445049a6984bbd91830d7c6107e43aL94) is also removed accordingly)
4. Then, because of 2., I didn't add `command` directly to that function, I created an additional condition check for it. ([this](https://github.com/WordPress/wporg-developer/pull/351/files/b0a77cfda6d111fcd46895648109cfa476f137c2#diff-5e0d1972031d52467147da1a95b1eaf3929b263e66640ad2b369130ab4789ef5R99) and [this](https://github.com/WordPress/wporg-developer/pull/351/files/b0a77cfda6d111fcd46895648109cfa476f137c2#diff-9133227a1757e0f7ff2996d7bc952b9db0445049a6984bbd91830d7c6107e43aR111))
5. And yes, stripped by line 113 is one of the reasons.

Overall, my idea comes from that `$form_data['post_type']` has always been `$parser_post_types`, and it doesn't make much sense to get the `$post_type` [from](https://github.com/WordPress/wporg-developer/pull/351/files/b0a77cfda6d111fcd46895648109cfa476f137c2#diff-9133227a1757e0f7ff2996d7bc952b9db0445049a6984bbd91830d7c6107e43aL103-L104) `$_POST['data']`.

outdoor2kode: ps. you might need to expand the code sections to view the reference lines for some of the links.

adamwoodnz: OK I've stepped through the code now and it makes sense to me. It does make me wonder if we can extend the autocomplete to all the other handbooks in the same manner (`/themes`, `/rest-api`, etc). That can be a separate issue though.

adamwoodnz (approved):
:shipit: LGTM

source/wp-content/themes/wporg-developer-2023/inc/autocomplete.php
outdoor2kode: Ticket for it was created. https://github.com/WordPress/wporg-developer/issues/367

PR #369, opened 2023-11-14
Fix missing Methods section for Class

### LINKED ISSUE #325: Code Reference: Methods missing for WP_Hook.
I would expect to see the methods on this page:
https://developer.wordpress.org/redesign-test/reference/classes/wp_hook/

| Current | Staging |
|--------|--------|
| Screenshot 2023-11-03 at 2 37 25 PM | Screenshot 2023-11-03 at 2 37 20 PM |

outdoor2kode:
Fixes #325.

This PR adds a custom block to render the missing Methods section on Class pages.

### Screenshot
image

### Note I
For the missing "flushing-rewrite-on-activation" section mentioned [here](https://github.com/WordPress/wporg-developer/issues/300#issuecomment-1780375363), since there are some sort of broken linkings between the Explanation and the Function on staging env, the whole [More Information](https://developer.wordpress.org/redesign-test/reference/functions/register_post_type/#more-information) is not rendered. I've fixed the page by manually adding an explanation to the function `register_post_type` on the staging env, with same the content as the previous one.

| **The synced explanation doesn't link correctly to the corresponding function** (post name: (no title)) | **Manually added one** (post name: Explanation: register_post_type) |
|-|-|
| image | image |

And because we're not going to touch any data on the current developer site and we're going to apply the 2023 theme to the current site, we don't need to do anything on this issue, once the new theme is applied, `More Information` will be there. 

Put simply: The logic is there, it's just the data linking is broken on staging and local env.

### Note II
There are also some kinds of data linking problems between the Class and Methods, but only on local env.
Locally, [this](https://github.com/WordPress/wporg-developer/compare/fix/missing-Methods-section-for-Class?expand=1#diff-ae29e15b24374e527ce977eb53b1fad259230fb058568e95575ac4e968c3ebb3R67) would return an empty array, making Methods section not rendered.
If you want to test this PR, you can either test directly on the staging env while sandboxed, 
or you can run the following code once on your local env, it will designate every method of [(WP_Hook)](http://localhost:8888/reference/classes/wp_hook/#description) as a child of WP_Hook.
```
$class_post = get_posts(array(
    'post_type'   => 'wp-parser-class',
    'title'       => 'WP_Hook',
    'numberposts' => 1
));

if ( count($class_post) ) {
    $class_post_id = $class_post[0]->ID;

    $method_posts = get_posts(array(
        'post_type'   => 'wp-parser-method',
        'numberposts' => -1,
        's'           => 'WP_Hook::',
    ));

    foreach ( $method_posts as $method_post ) {
        wp_update_post(array(
            'ID'          => $method_post->ID,
            'post_parent' => $class_post_id,
        ));
    }
}
```
I've been trying to locate where method posts are designated as child posts of a class post with keywords like 'parent_post' on either the old codebase or the current one, but there's no luck finding it. If anyone gets any idea about this, please let me know, thanks.

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/code-methods/style.scss
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-methods/style.scss
@@ -1,1 +1,0 @@
-/* css styles */
--- a/source/wp-content/themes/wporg-developer-2023/src/code-methods/style.scss
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-methods/style.scss
@@ -1,0 +1,1 @@
+/* css styles */
--- a/source/wp-content/themes/wporg-developer-2023/functions.php
+++ b/source/wp-content/themes/wporg-developer-2023/functions.php
@@ -155,6 +155,7 @@ require_once __DIR__ . '/src/cli-command-table/block.php';
 require_once __DIR__ . '/src/code-changelog/block.php';
 require_once __DIR__ . '/src/code-deprecated/block.php';
 require_once __DIR__ . '/src/code-description/block.php';
+require_once __DIR__ . '/src/code-methods/block.php';
 require_once __DIR__ . '/src/code-explanation/block.php';
 require_once __DIR__ . '/src/code-hooks/block.php';
 require_once __DIR__ . '/src/code-parameters/block.php';
@@ -569,7 +570,9 @@ function add_handbook_templates( $templates ) {
 /**
  * Filters content for the code reference blocks so Table of Contents can be added.
  *
- * Note: This filter is added and removed in src/code-description/block.php to prevent infinite loops.
+ * Note: This filter is added and removed in the files below to prevent infinite loops.
+ * - src/code-description/block.php
+ * - src/code-methods/block.php
  * Any update to the function name should be reflected there.
  *
  * @param string $content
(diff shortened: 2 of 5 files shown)


source/wp-content/themes/wporg-developer-2023/src/code-methods/style.scss
adamwoodnz, line 1: Can you omit this if it's not needed?

adamwoodnz (commented):
Nice work!

We don't have this is in the designs, and I wonder if the table layout we use for Related might be appropriate, @WordPress/meta-design thoughts?

Screenshot 2023-11-16 at 4 16 16 PM

adamwoodnz:
> Locally, [this](https://github.com/WordPress/wporg-developer/compare/fix/missing-Methods-section-for-Class?expand=1#diff-ae29e15b24374e527ce977eb53b1fad259230fb058568e95575ac4e968c3ebb3R67) would return an empty array, making Methods section not rendered.

I have methods locally, not sure what's different 🤷 

Screenshot 2023-11-16 at 4 35 18 PM

jasmussen:
That seems useful to me, yes, if it's easy.

source/wp-content/themes/wporg-developer-2023/src/code-methods/style.scss
outdoor2kode: Done ✅

outdoor2kode:
> I have methods locally, not sure what's different 🤷

hmm.. weird, how did you import the data? Ran the `wp cron event run --all`?

adamwoodnz:
> how did you import the data? Ran the `wp cron event run --all`

Yeah I think so. I manually imported some data, but pretty sure this wasn't part of it.

adamwoodnz:
> yes, if it's easy

@ren this was regarding using the code table; are you planning to make that change on this PR or should we merge and do a follow up PR?

outdoor2kode:
> @ren this was regarding using the code table; are you planning to make that change on this PR or should we merge and do a follow up PR?

I'm going to merge this PR first and make the UI change on this [ticket](https://github.com/WordPress/wporg-developer/issues/384).

Issue #368, opened 2023-11-14
Accessibility issues with handbook layout


adamwoodnz:
From [Alex Stine on Slack](https://wordpress.slack.com/archives/C04U953K77A/p1699932257988309?thread_ts=1699558771.994199&cid=C04U953K77A):

Here is a quick list:

- [x] The Improve it on GitHub link is not descriptive enough.
- [x] Same for See list of changes link.
- [x] The link NextGetting Started has no spacing and is announced as 2 words instead of 3.
- [x] The handbook chapters heading 2 is the first heading on the page. This design needs to be reworked, no questions asked.
- [x] The skip to main content link does not land you in the main content, it lands you in all the items before the main heading 1.
- [x] The main landmark should not wrap navigational elements such as the Chapters heading 2 or the search form above it.

adamwoodnz:
cc @fcoveram @ndiego, not sure how hard this will be to fix yet

ndiego:
> - The handbook chapters heading 2 is the first heading on the page. This design needs to be reworked, no questions asked.
> - The skip to main content link does not land you in the main content, it lands you in all the items before the main heading 1.
> - The main landmark should not wrap navigational elements such as the Chapters heading 2 or the search form above it.

Are these three items due to the fact that the Chapters list is inside `<main>` and not wrapped in an `<aside>` tag? I was comparing the current layout with other sites like [Mozilla](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Selectors/Attribute_selectors) and the [React docs](https://react.dev/learn/reacting-to-input-with-state). In both cases, they have a heading that is not an H1 that is the first on the page, but, it's not in `<main>` and is wrapped in an `<aside>`.

cc @alexstine

alexstine:
@ndiego It would be more valid if it were wrapped in an `<aside>` tag but it would be even better if the heading 1 came first. Headings should represent importance on a page so I always have argued that the `<aside>` should appear directly after `<main>` in the DOM.

adamwoodnz:
Pretty sure I'll be able to reorder the content and keep the visual layout. It's next on my list after the ToC styling.

adamwoodnz:
Should the Table of Contents also be an aside?

Something like

```html
<header { including breadcrumbs and search } >
<main>
<aside class="chapter-list">
<aside class="table-of-contents">
<footer>
```

The visual layout will be tough to achieve if so. It would help greatly if it was immediately after the h1 within main.

[...]

Issue #345, opened 2023-11-06
Remove 'Last updated' from Block Editor Handbook


adamwoodnz:
@adamwoodnz this is unrelated to this issue, but note that the "Last updated" date for articles in the Block Editor Handbook are not correct since the entire handbook resyncs periodically throughout the day. If it's possible to remove the  "Last updated" date for the Block Editor Handbook, that would be great.

_Originally posted by @ndiego in https://github.com/WordPress/wporg-developer/issues/339#issuecomment-1795035899_

PR #376, opened 2023-11-15
Fix accessibility issues with page layouts

adamwoodnz:
See #368 

Several issues with the handbook layouts are fixed, and some of these fixes have also been applied to other layouts (remove search field from main landmark):

- [x] The handbook chapters heading 2 is the first heading on the page. This design needs to be reworked, no questions asked.
- [x] The skip to main content link does not land you in the main content, it lands you in all the items before the main heading 1.
- [x] The main landmark should not wrap navigational elements such as the Chapters heading 2 or the search form above it.

**The visual layout should not change.**

### Testing

Use dev tools to test and check the following:

1. Almost all the templates have updated layouts, with the Search field moving out of `main`. The skip to content link should take you directly to the main content, with no search or chapters in main before the h1.
2. On handbook pages the chapter list aside has been moved after main, so check the headings hierarchy and accessibility tree

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/templates/single-handbook-github.html
+++ b/source/wp-content/themes/wporg-developer-2023/templates/single-handbook-github.html
@@ -1,12 +1,6 @@
 <!-- wp:template-part {"slug":"header","className":"has-display-contents"} /-->
 
-<!-- wp:group {"style":{"spacing":{"padding":{"left":"var:preset|spacing|edge-space","right":"var:preset|spacing|edge-space"}}},"className":"alignfull","layout":{"type":"default"}} -->
-<div class="wp-block-group alignfull" style="padding-right:var(--wp--preset--spacing--edge-space);padding-left:var(--wp--preset--spacing--edge-space)">
-
-	<!-- wp:pattern {"slug":"wporg-developer-2023/search-field"} /-->
-
-</div>
-<!-- /wp:group -->
+<!-- wp:template-part {"slug":"search-wide","className":"has-display-contents"} /-->
 
 <!-- wp:group {"align":"full","style":{"spacing":{"padding":{"right":"var:preset|spacing|edge-space","left":"var:preset|spacing|edge-space"}}},"layout":{"type":"constrained"}} -->
 <div class="wp-block-group alignfull" style="padding-right:var(--wp--preset--spacing--edge-space);padding-left:var(--wp--preset--spacing--edge-space)">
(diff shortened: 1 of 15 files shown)


source/wp-content/themes/wporg-developer-2023/src/handbook-pagination/view.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Add screen reader text to handbook pagination links.
+ * This is needed because the post-navigation-link block cannot accept rich content via the label attribute.
```
adamwoodnz, line 8: I tried passing this content through the label attribute. Passing the HTML string in directly as `<span aria-hidden="true">Next </span><span class="screen-reader-text">Next chapter: </span>` worked, but using `wp_kses_post` didn't, I think because it broke the block markup. The rendered block was missing all the passed attributes.

```php
wp_kses_post(__(<span aria-hidden="true">${ visualLabel }</span><span class="screen-reader-text">${ screenReaderText }</span>)
```

source/wp-content/themes/wporg-developer-2023/templates/single-handbook-github.html
adamwoodnz, line 1: No longer required now that the pagination has built in logic to determine whether to show GitHub links

source/wp-content/themes/wporg-developer-2023/src/article-meta-github/block.php
```diff
@@ -0,0 +1,86 @@
...
+		do_blocks( '<!-- wp:paragraph {"className":"external-link"} --><p class="external-link"><a href="[article_edit_link]">%s</a></p><!-- /wp:paragraph -->' ),
+		sprintf(
+			/* translators: %s: article title */
+			__( 'Improve it on GitHub<span class="screen-reader-text">: %s</span>', 'wporg' ),
+			esc_html( $title )
+		)
+	);
+	$content .= do_blocks( '</div><!-- /wp:group -->' );
+
+	if ( 'blocks-handbook' !== $post_type ) {
```
adamwoodnz, line 68: See https://github.com/WordPress/wporg-developer/issues/345

StevenDufresne:
This PR is hard to follow as there are abstractions/refactoring and format updates. Is there an easier way to look at this? Otherwise, I don't think I can provide a useful code-specific review.

adamwoodnz:
> This PR is hard to follow as there are abstractions/refactoring and format updates. Is there an easier way to look at this? Otherwise, I don't think I can provide a useful code-specific review.

Yeah that's fair, there's a lot going on. I'll split it in 3; the 2 new blocks and the layout changes.

adamwoodnz:
> there are abstractions/refactoring and format updates. Is there an easier way to look at this?

The scope of this has been reduced to only include the layout changes, see updated description 🙏

PR #377, opened 2023-11-16
Use IBM Plex Sans for headings

### LINKED ISSUE #374: Use IBM Plex Sans for the heading font-family
In reviewing the new Developer Resources design, I noticed that the title on the homepage and a few headings at the bottom are in EB Garamond. However, this font is used nowhere else on the site. 

image

In discussing this issue with @jasmussen and @fcoveram and reviewing alternatives ([see Figma](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-Resources?type=design&node-id=4147-7431&mode=design&t=Ew3TwwR0vKOpuD5D-0)), we also identified that the use of IBM Plex Mono for all headings (even heading that are not code) was not ideal. This was only apparent after seeing the staging site live. 

So, to address the EB Garamond issue and the extensive use of IBM Plex Mono for non-code text, the proposed solution is to use IBM Plex Sans for all non-code headings, which includes the homepage title. This would be a new font but would provide some contrast to Inter (used for all other text) while keeping some connective tissue with the `code` font (IBM Plex Mono) used throughout Developer Resources.

### Homepage

| Current | Proposed |
|-|-|
| image | image |

### Internal Page

| Current | Proposed |
|-|-|
| image | image |

### Code reference page (a heading that is also code)

| Current | Proposed |
|-|-|
| image | image |

### Implementation

IBM Plex Sans will need to be added to the theme, but in some initial tests, the only additional step will be to set `    --wp--custom--heading--typography--font-family: var(--wp--preset--font-family--ibm-plex-sans);` (or whatever the name of the preset variable is).

The IBM Plex Sans will also be used in the redesign of the Developer Blog, so if this issue should instead be logged in the parent theme, let me know, and I'll move it over. Thanks!

adamwoodnz:
Closes #374 
Depends on https://github.com/WordPress/wporg-mu-plugins/issues/506
Depend on https://github.com/WordPress/wporg-parent-2021/pull/116

Use IBM Plex Sans for:
- [x] Anything currently using EB Garamond (eg. home page h1 and h2s)
- [x] All non code titles (eg. exclude code reference titles)

### Screenshots

| Home | New and Updated | Code | Handbook |
|-|-|-|-|
| localhost_8888_(Desktop) | localhost_8888_reference_since_6 3 0_(Desktop) (2) | localhost_8888_reference_functions__prime_site_caches_(Desktop) | localhost_8888_block-editor_(Desktop) |


jasmussen:
Thanks for the screenshots. IMO this adds a nice bit of character to the section, and is more legible for headings than the monospaced font. I think this is a good way to start.

fcoveram:
It looks great

PR #385, opened 2023-11-20
Fix first/last updates font size to match adjacent links.

### LINKED ISSUE #382: First/Last Published font size should be smaller (or match edit article/changelog)
The font sizes for First/Last published seem to be larger than the "Edit Article/Changelog" links. They should be the same.

Example: https://developer.wordpress.org/redesign-test/rest-api/reference/taxonomies/

## Screenshot
Screenshot 2023-11-20 at 12 44 03 PM

StevenDufresne:
Fixes #382.

The "Improve" links are inheriting a style of `14px` from the [parent theme](https://github.com/WordPress/wporg-parent-2021/blob/aa715a7bc1d9b5a63edd636e550f7791772568fb/source/wp-content/themes/wporg-parent-2021/sass/page/_single.scss#L3C1-L3C1). In this case, the smaller text seems to work better as it's secondary information. As a result, I've change the first/last to be the same size of `14px`. 

Noting we can't set the `fontSize` to `small` because it will change the line-height. So we remove it allow the parent styles to style these controls.

| Before  | After |
|--------|--------|
| Screenshot 2023-11-20 at 3 38 21 PM | Screenshot 2023-11-20 at 3 48 25 PM |

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/patterns/article-meta-github.php
+++ b/source/wp-content/themes/wporg-developer-2023/patterns/article-meta-github.php
@@ -15,7 +15,7 @@
 		<p style="font-style:normal;font-weight:700"><?php esc_html_e( 'First published', 'wporg' ); ?></p>
 		<!-- /wp:paragraph -->
 
-		<!-- wp:post-date {"fontSize":"small"} /-->
+		<!-- wp:post-date /-->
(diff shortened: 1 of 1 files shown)


outdoor2kode (approved):
LGTM 👍

PR #388, opened 2023-11-20
Fix accessibility of GitHub links

### LINKED ISSUE #345: Remove 'Last updated' from Block Editor Handbook

adamwoodnz:
See #368 
Closes #345 

Handbook article meta GitHub links now include the page title for context, eg. 'See list of changes: WordPress Coding Standards'.

To achieve this the article meta GitHub pattern has been migrated to a block so that it has access to post data via block context.

The Block Editor handbook pages no longer have a Changelog link in article meta, see #345 

There are no visual changes.

### Testing

Screenshot 2023-11-21 at 9 26 29 AM

Use the accessibility tree in dev tools to see the text content for the links, and listen to how it is read with Voiceover, NVDA or JAWS.

Check that the Changelog link is present on other handbooks like [WPCS](http://localhost:8888/coding-standards/wordpress-coding-standards/), but not the [Block Editor](http://localhost:8888/block-editor/) handbook.


source/wp-content/themes/wporg-developer-2023/templates/single-handbook-github.html
adamwoodnz, line 1: This is no longer required now that there is [conditional rendering logic in the block](https://github.com/WordPress/wporg-developer/pull/388/files#diff-fffd9ab28f1182151553823e544586e4ad7186670afea891e11c66614ff3e34cR46). The inclusion of the github links was the only difference with `single-handbook.html`, so we have removed a lot of duplicate markup.

source/wp-content/themes/wporg-developer-2023/src/article-meta-github/block.php
```diff
@@ -0,0 +1,86 @@
...
+		! isset( $post_id )
+		|| ! isset( $post_type )
+		|| ! in_array( $post_type, $github_handbooks, true )
+	) {
+		return '';
+	}
+
+	$title = get_the_title( $post_id );
+
+	$content = do_blocks( '<!-- wp:group {"style":{"spacing":{"blockGap":"0"}},"layout":{"type":"constrained"}} --><div class="wp-block-group">' );
```
StevenDufresne: Does this group need to be in this block?

StevenDufresne: Ah, nevermind. I see, this is the container for the individual block.


Could we connect the first and last concatenation and `do_blocks` only once?
```
$inner_content = "<!-- wp:paragraph....";
$content .= do_blocks(
    sprintf( '<!-- wp:group {"style":{"spacing":{"blockGap":"0"}},"layout":{"type":"constrained"}} --><div class="wp-block-group">%s</div><!-- /wp:group -->'  , $inner_content ) 
)
```

StevenDufresne (approved):
I can confirm that there is screen reader text in the links that include the article title.

source/wp-content/themes/wporg-developer-2023/src/article-meta-github/block.php
adamwoodnz: Thanks, cleaned up the markup in https://github.com/WordPress/wporg-developer/pull/388/commits/6e6761a363e11cc1f3056c479146dc9dd4c8a304

PR #389, opened 2023-11-20
Fix accessibility of handbook pagination

### LINKED ISSUE #368: Accessibility issues with handbook layout

adamwoodnz:
Closes #368 

Handbook pagination links now include more context, including the page title of the post to be navigated to, eg. 'Next chapter: Get Started'.

To achieve this the handbook pagination pattern has been migrated to a block so that it has access to post data via block context.

There are no visual changes.

#### Next

Screenshot 2023-11-21 at 9 37 50 AM

#### Previous

Screenshot 2023-11-21 at 9 38 46 AM

Use the accessibility tree in dev tools to see the text content for the links, and listen to how it is read with Voiceover, NVDA or JAWS.


source/wp-content/themes/wporg-developer-2023/parts/search-wide.html
```diff
@@ -0,0 +1,7 @@
+<!-- wp:group {"style":{"spacing":{"padding":{"left":"var:preset|spacing|edge-space","right":"var:preset|spacing|edge-space"}}},"className":"alignfull","layout":{"type":"default"}} -->
+<div class="wp-block-group alignfull" style="padding-right:var(--wp--preset--spacing--edge-space);padding-left:var(--wp--preset--spacing--edge-space)">
```
StevenDufresne: I saw these changes on the other a11y branch fix: https://github.com/WordPress/wporg-developer/pull/388/files#diff-64e65ccda0f0264641251aef7432f56f0916d0f922c6312b3387553ab6be2d25R2

source/wp-content/themes/wporg-developer-2023/src/handbook-pagination/block.php
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @return string Returns the block markup.
+ */
+function render( $attributes, $content, $block ) {
+	$content = do_blocks(
+		'<!-- wp:group {"align":"full","style":{"border":{"top":{"color":"var:preset|color|light-grey-1","width":"1px","style":"solid"}},"spacing":{"margin":{"top":"var:preset|spacing|20","bottom":"var:preset|spacing|20"},"padding":{"top":"var:preset|spacing|30","bottom":"var:preset|spacing|30"}}},"layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"space-between"}} -->
+		<div class="wp-block-group alignfull" style="border-top-color:var(--wp--preset--color--light-grey-1);border-top-style:solid;border-top-width:1px;margin-top:var(--wp--preset--spacing--20);margin-bottom:var(--wp--preset--spacing--20);padding-top:var(--wp--preset--spacing--30);padding-bottom:var(--wp--preset--spacing--30)">'
+	);
+	$content .= sprintf(
+		do_blocks( '<!-- wp:post-navigation-link {"type":"previous","label":"%s","showTitle":true,"linkLabel":true} /-->' ),
```
StevenDufresne: Same as in #388, do we need to call `do_blocks` this often? Can we not call it once?

source/wp-content/themes/wporg-developer-2023/src/handbook-pagination/block.json
```diff
@@ -0,0 +1,18 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/block.json",
```
StevenDufresne: Can we not use the filters to output the html? 

https://github.com/WordPress/wporg-developer/blob/dea1a77ba1a665ba25eeffb438d2bb7471ae0d3a/source/wp-content/themes/wporg-developer-2023/functions.php#L190

source/wp-content/themes/wporg-developer-2023/parts/search-wide.html
adamwoodnz: Ah both these branches need rebasing after the layout branch was merged, thanks

source/wp-content/themes/wporg-developer-2023/functions.php
```diff
@@ -709,11 +710,33 @@ function get_adjacent_handbook_post_link( $output, $format, $link, $post, $adjac
...
+			: /* translators: %s: post title */
+			__( 'Next chapter: %s', 'wporg' ),
+		$title
+	);
+
+	$content = str_replace(
+		'%title',
+		sprintf(
+			'<span aria-hidden="true" class="post-navigation-link__label">%1$s</span>
+			<span aria-hidden="true" class="post-navigation-link__title">%2$s</span>
```
adamwoodnz, line 726: Adding these classes, which would usually only be applied if the `wp:post-navigation-link` had a label configured, means that the [existing styles from the parent theme](https://github.com/WordPress/wporg-parent-2021/blob/trunk/source/wp-content/themes/wporg-parent-2021/sass/blocks/_post-navigation-link.scss#L12) continue to work:

Screenshot 2023-11-22 at 11 16 34 AM

source/wp-content/themes/wporg-developer-2023/src/handbook-pagination/block.json
adamwoodnz: Yes! Great catch, updated in https://github.com/WordPress/wporg-developer/pull/389/commits/e9e7740cf80d021730efadbfd6d3fcda0fa483b0

source/wp-content/themes/wporg-developer-2023/src/handbook-pagination/block.php
adamwoodnz: No longer using a block for this

StevenDufresne:
This is doing what it is supposed to. 

I'm not certain that calling everything a "chapter" is technically correct. The outermost items are chapters but are the inner pages chapters? This isn't a sticking point for me but I wanted to mention it.

adamwoodnz:
> I'm not certain that calling everything a "chapter" is technically correct.

Fair point. 'Next article:' actually might be more accurate overall, or just 'Next:'

PR #404, opened 2023-11-22
Fix broken code block sticky header

### LINKED ISSUE #373: Broken code block sticky header
sticky

outdoor2kode:
Fixes #373

This PR gives local-navigation-bar a 60px height so that the sticky header can set its `top` accurately.
Also make the shape of the code block rounded corners on the bottom complete
Depends on https://github.com/WordPress/wporg-mu-plugins/pull/519.

## Screencast

https://github.com/WordPress/wporg-developer/assets/18050944/83469890-4f98-44fd-ada7-5092d2dce1e0

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/scss/prism.scss
+++ b/source/wp-content/themes/wporg-developer-2023/scss/prism.scss
@@ -209,8 +209,6 @@ pre.line-numbers > code {
 
 /* Copy button */
 .wp-code-block-button-container {
-	position: sticky;
-	top: calc(var(--wp--custom--local-navigation-bar--spacing--height) + var(--wp-admin--admin-bar--height, 0px));
 	z-index: 1;
 	display: flex;
 	justify-content: right;
--- a/source/wp-content/themes/wporg-developer-2023/scss/prism.scss
+++ b/source/wp-content/themes/wporg-developer-2023/scss/prism.scss
@@ -264,13 +264,13 @@ pre.line-numbers > code {
 	&::after {
 		content: "";
 
-		/* Since the margin-bottom of the pre defined in _global.scss is 1.6em,
-		 * -1.63em is set here to make the shape of the rounded corners at the bottom of
-		 * the code block complete when the sticky wp-code-block-button-container is scolled there.
+		/* Since the margin-bottom of the pre defined in _global.scss is 1em,
+		 * -1.1em is set here to make the shape of the rounded corners at the bottom of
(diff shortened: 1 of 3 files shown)


source/wp-content/themes/wporg-developer-2023/scss/prism.scss
```diff
@@ -210,7 +210,7 @@ pre.line-numbers > code {
 /* Copy button */
 .wp-code-block-button-container {
 	position: sticky;
-	top: calc(var(--wp-global-header-height) + var(--wp-admin--admin-bar--height));
+	top: calc(var(--wp--custom--local-navigation-bar--spacing--height) + var(--wp-admin--admin-bar--height));
```
adamwoodnz: I noticed this doesn't work when I'm logged out because `--wp-admin--admin-bar--height` is undefined, might need a fallback
```suggestion
	top: calc(var(--wp--custom--local-navigation-bar--spacing--height) + var(--wp-admin--admin-bar--height, 0px));
```

renintw: Good catch! Fixed.

outdoor2kode: Good catch! Fixed.

outdoor2kode:
Based on our discussion in the ticket, I've removed the sticky feature. However, I've kept the other modifications intact. These changes won't affect the UI due to the absence of sticky, and will actually make it easier to add back the sticky feature in the future if needed.

source/wp-content/themes/wporg-developer-2023/theme.json
```diff
@@ -151,6 +151,11 @@
 						"top": "150px"
 					}
 				}
+			},
+			"local-navigation-bar": {
+				"spacing": {
+					"height": "60px"
```
adamwoodnz: As discussed on https://github.com/WordPress/wporg-mu-plugins/pull/519, prefer to set this value as the default there, then all these themes (Showcase, Documentation and Developer) don't need to set it.

outdoor2kode: Showcase: https://github.com/WordPress/wporg-showcase-2022/pull/264
Developer: [Remove the height setting](https://github.com/WordPress/wporg-developer/pull/404/commits/5a503a23fb94386a8fe9f45716d070a7b58271d3)
Documentation: No padding was set at the beginning. I've built the local environment and checked out production, the 60px from mu-plugins was correctly applied.

PR #408, opened 2023-11-26
Fix broken TOC when showing more Used by

### LINKED ISSUE #372: Broken TOC when showing more Used by
ezgif com-optimize

outdoor2kode:
Fixes #372

This PR introduces modifications to the collapse table functionality, ensuring that the sidebar repositions correctly after collapsing.

## Screencast
Screen Capture on 2023-11-27 at 05-29-25

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/code-table/view.js
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-table/view.js
@@ -4,6 +4,7 @@ const init = () => {
 		tables.forEach( ( table ) => {
 			const showMoreLink = table.querySelector( '.wp-block-wporg-code-table-show-more' );
 			const showLessLink = table.querySelector( '.wp-block-wporg-code-table-show-less' );
+			const sidebarContainer = document.querySelector( '.wp-block-wporg-sidebar-container' );
 
 			if ( showMoreLink ) {
 				showMoreLink.addEventListener( 'click', ( event ) => {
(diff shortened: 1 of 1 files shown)


adamwoodnz (commented):
This works for me, however I don't think it the responsibility of the code table to reset the sidebar state. This fix will also need to be implemented on other sites like Documentation.

I feel like the root cause of the problem is that the [onScroll logic](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/sidebar-container/src/view.js#L29) isn't running when the table is collapsed and the height of the page changes. Could you see if we can fix it there instead?

outdoor2kode:
> I feel like the root cause of the problem is that the [onScroll logic](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/sidebar-container/src/view.js#L29) isn't running when the table is collapsed and the height of the page changes. Could you see if we can fix it there instead?

I was initially considering modifying the onScroll logic to address the issue but realized that the changes might make the logic a bit difficult to understand. So I went for a more intuitive fix. I have opened a [PR](https://github.com/WordPress/wporg-mu-plugins/pull/522) in the mu-plugins to modify the logic within onScroll. Please check it out. Thanks!

--

Other than that, do you know about the toggle here:
https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/sidebar-container/src/view.js#L88
Is this toggle still in use anywhere?

adamwoodnz:
> Other than that, do you know about the toggle here:
> https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/sidebar-container/src/view.js#L88
> Is this toggle still in use anywhere?

Yep, still in use below 1200px

Screenshot 2023-11-28 at 9 19 32 AM

outdoor2kode:
> Yep, still in use below 1200px

Oh, right. Obviously I remembered it incorrectly; I had thought the toggle was at the original position but removed due to a design decision. Thanks!

outdoor2kode:
Fixed by https://github.com/WordPress/wporg-developer/issues/372.

PR #410, opened 2023-11-27
Add styles for comment link modal

### LINKED ISSUE #392: Link action broken on user notes and feedback forms once there is a note saved
Steps to reproduce

1. Login and find a post which already has a note, eg. https://developer.wordpress.org/redesign-test/reference/classes/atomentry/#comment-12
2. Start to add a new note or leave feedback on an existing note
3. Attempt to add a link by clicking the link button

Expect: modal to open to add the link.

Actual: nothing happens, no modal and no errors in console

Could be related to 2 forms and scripts being loaded on the page.

adamwoodnz:
Fixes #392 

Fixes an issue where the modal for adding a link in a user note or feedback didn't appear when there is already a note added to the code reference item being viewed.

I discovered that the JS correctly activates the modal but the editor and dashicons styles are missing so the modal is shown at the bottom of the page, rather than over the top:

The dashicons in the admin bar are also broken, only on this page:

Screenshot 2023-11-29 at 11 50 28 AM

I haven't been able to work out why the styles aren't enqueued correctly, (they should be added [here](https://github.com/WordPress/wordpress-develop/blob/a21f5a3cc054137f3ace4760db43845158fb3cc2/src/wp-includes/class-wp-editor.php#L209)), so this PR ensures they are loaded if the `code-comments` block is loaded. It isn't an issue if there aren't any notes submitted yet and only the `code-comment-form` block is loaded, so we don't need to do anything in that block.

I've also added styles to bring the modal more inline with the parent theme, using colors and spacing.

### Screenshots

| Desktop | Tablet | Mobile |
|-|-|-|
| localhost_8888_reference_functions_is_admin_(Desktop) | localhost_8888_reference_functions_is_admin_(iPad) | localhost_8888_reference_functions_is_admin_(Samsung Galaxy S20 Ultra) |

### Testing

1. Add a user note on a code reference single page with no existing notes. Include a link to ensure it works.
2. Once the page reloads check in the devtools network tab that both editor.css and dashicons.css are loaded. The dashicons in the admin bar should be ok.
3. Add feedback on the first note, and include a link.
4. Add a new note and include a link.

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
+++ b/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
@@ -532,6 +532,11 @@ body[class] {
 	#wp-link-close {
 		top: var(--wp--preset--spacing--20);
 		right: var(--wp--preset--spacing--20);
+
+		&:hover,
+		&:focus {
+			color: var(--wp--custom--button--hover--color--background);
+		}
 	}
 
 	#wp-link input[type="text"] {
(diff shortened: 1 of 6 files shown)


source/wp-content/themes/wporg-developer-2023/src/style/style.scss
```diff
@@ -507,3 +507,14 @@ body[class] {
...
+
+#wp-link-wrap {
+	height: 250px !important;
+	margin-top: -125px !important;
+}
+
+#link-selector #search-panel,
+#link-selector #wplink-link-existing-content,
+#link-selector #link-options .link-target {
+	display: none;
```
outdoor2kode: Do we just temporarily set it as "none" and will bring it back once its layout is fixed?

outdoor2kode (commented):
Look good 👍 I can see the modal popping out at the right place.

Just a few questions and comments:

1. The style `outline--focus--color` doesn't seem to exist.
Screenshot 2023-12-04 at 1 49 02 AM
Screenshot 2023-12-04 at 1 51 11 AM
2. Style for `:focus` looks inconsistent, maybe that's expected? There's no extra borderline for the cancel button.
Screenshot 2023-12-04 at 1 53 59 AM
3. By clicking the `Edit`, we can enter the admin edit page. Do we also look to adjust the modal styles there?
image

source/wp-content/themes/wporg-developer-2023/src/style/style.scss
adamwoodnz: This UI is currently hidden in prod, and I copied [this CSS from the old theme](https://github.com/WordPress/wporg-developer/blob/old/source/wp-content/themes/wporg-developer/scss/main.scss#L1917). I'm not aware of any desire to support this functionality.

adamwoodnz:
Thanks @renintw 

> 1. The style `outline--focus--color` doesn't seem to exist.

[I've added this to parent](https://github.com/WordPress/wporg-parent-2021/commit/3e4525555177b25a2905a0b1587d72970bc1ccf5), you may need to update your composer dependency

> 2\. Style for `:focus` looks inconsistent, maybe that's expected? There's no extra borderline for the cancel button.

👍 I'll take a look at this

> 3\. By clicking the `Edit`, we can enter the admin edit page. Do we also look to adjust the modal styles there?

Yes, I think we should probably hide those extra fields like the frontend. Button styles and spacing I'm not so worried about as it matches the older style of the admin.

adamwoodnz:
> 2\. Style for `:focus` looks inconsistent, maybe that's expected? There's no extra borderline for the cancel button.

Updated styles for all buttons in the modal:

Screenshot 2023-12-04 at 1 46 02 PM

adamwoodnz:
> 3. By clicking the Edit, we can enter the admin edit page. Do we also look to adjust the modal styles there?

Added admin styles

Screenshot 2023-12-04 at 2 05 52 PM

outdoor2kode (approved):
LGTM 👍

PR #411, opened 2023-11-27
Add back <p> tags via wpautop

### LINKED ISSUE #396: Paragraph tags are missing in the Code Reference docs
I was looking at the doc for [`WP_Query`](https://developer.wordpress.org/redesign-test/reference/classes/wp_query/) and was wondering why all the content seemed to be very close together. For example, there is no spacing between text and code blocks. I then realized that text content is not wrapped in paragraph tags like on the current site. Adding paragraph tags should fix the spacing issue.

| Current | Redesign |
|-|-|
|image|image|

outdoor2kode:
Fixes #396 

This PR uses `add_filter( 'the_content', 'wpautop' );` to add back the missing `<p>` for the Explanation content.

## Screenshot

| Before | After |
|-|-|
| image | image |
| image | image |

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/inc/template-tags.php
+++ b/source/wp-content/themes/wporg-developer-2023/inc/template-tags.php
@@ -1880,6 +1880,9 @@ namespace DevHub {
 		// Restore original global post.
 		$post = $orig;
 
+		// Automatic paragraph and line break formatting in content.
+		add_filter( 'the_content', 'wpautop' );
+
 		// Restore filter.
 		add_filter( 'the_content', array( 'DevHub_Formatting', 'fix_unintended_markdown' ), 1 );
 


adamwoodnz (commented):
I expected it would be a problem with the parser, but this works pretty well, how did you know to use this filter?

I notice we do have a lot of extra `p` tags though. They actually don't affect the layout, but will significantly increase the size of the dom. Maybe the [filter settings](https://developer.wordpress.org/reference/functions/wpautop/) might be able to fix this, or maybe it will require changes in the parser?

Screenshot 2023-11-28 at 11 49 58 AM

outdoor2kode:
I first tried regex then found out `wpautop` and saw it's used [here](https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer-2023/src/code-description/block.php#L102).

And I've noticed that actually the `the_content` filter already adds `<p>` tags to `$content`, which is why the layout didn't break in production (Empty lines are also replaced with `<p></p>`). However, here, since `$content` is used for rendering a block in the end, it seems that the `<p>` tags disappear under Gutenberg's processing.

What do you refer to as the 'parser' in this context?

adamwoodnz:
> What do you refer to as the 'parser' in this context?

I assumed the explanation was generated from some other content, but I see now that it looks like content created in the admin, ignore 🙂 

> And I've noticed that actually the `the_content` filter already adds `<p>` tags to `$content`, which is why the layout didn't break in production (Empty lines are also replaced with `<p></p>`). However, here, since `$content` is used for rendering a block in the end, it seems that the `<p>` tags disappear under Gutenberg's processing.

So you don't think there's any way to not get the empty p tags? Doesn't even look possible to edit the content, but that seems buggy for a line break in the editor to become an extra tag. It doesn't happen in current prod:

Screenshot 2023-11-28 at 1 01 30 PM

outdoor2kode:
> I assumed the explanation was generated from some other content, but I see now that it looks like content created in the admin, ignore 🙂

Ah ok!

> So you don't think there's any way to not get the empty p tags? Doesn't even look possible to edit the content, but that seems buggy for a line break in the editor to become an extra tag. It doesn't happen in current prod:

I tried using regular expressions to directly add `<p>` tags to paragraphs while excluding `<br>`, empty lines, etc. But I still end up with `<p></p>`. I even tried to remove `<p></p>` directly with regular expressions before using `sprintf`, but it didn’t work either. It seems likely that the issue is with the `render_callback`. Do you have any ideas?

At the moment, I can’t think of a good solution to this problem. I’m inclined to merge first and then open a ticket to record the issue, and then go through the source code of `render_callback`.

adamwoodnz:
> I’m inclined to merge first and then open a ticket to record the issue, and then go through the source code of `render_callback`.

👍 Sounds good

adamwoodnz (approved):

outdoor2kode:
Issue opened: https://github.com/WordPress/wporg-developer/issues/413

PR #415, opened 2023-11-29
Fix Code Reference: Missing parsed code styles  

### LINKED ISSUE #322: Code Reference: Missing parsed code styles.
On this page:
 https://developer.wordpress.org/redesign-test/reference/classes/wpdb/parse_db_host/

I would expect the return parameters to be styled differently.

| Current | Staging |
|--------|--------|
| Screenshot 2023-11-02 at 3 58 35 PM | Screenshot 2023-11-02 at 3 57 56 PM |

outdoor2kode:
Fixes #322

Apply `syntax-red` to the parameter type.

## Screenshot

image

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/code-return-value/style.scss
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-return-value/style.scss
@@ -1,10 +1,5 @@
 .wp-block-wporg-code-reference-return-value {
-	.return-type {
-		font-family: var(--wp--preset--font-family--monospace);
-		font-size: var(--wp--preset--font-size--small);
-		color: var(--wp--preset--color--syntax-red);
-	}
-
+	.return-type,
 	.type {
 		font-family: var(--wp--preset--font-family--monospace);
 		font-size: var(--wp--preset--font-size--small);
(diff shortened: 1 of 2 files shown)


adamwoodnz (changes requested):
👍 works for Parameters.

Looks like we need to do it for Return though as well

Screenshot 2023-11-30 at 9 52 40 AM

outdoor2kode:
Cleaned up the CSS a bit so that the 'type' under both parameter and return has the same color, font, and font size.

## Updated Screenshot
image

source/wp-content/themes/wporg-developer-2023/src/code-return-value/style.scss
```diff
@@ -2,5 +2,12 @@
 	.return-type {
 		font-family: var(--wp--preset--font-family--monospace);
 		font-size: var(--wp--preset--font-size--small);
+		color: var(--wp--preset--color--syntax-red);
+	}
+
+	.type {
+		font-family: var(--wp--preset--font-family--monospace);
+		font-size: var(--wp--preset--font-size--small);
+		color: var(--wp--preset--color--syntax-red);
```
adamwoodnz: Combine these? The rules are the same.
```suggestion
	.return-type,
	.type {
		font-family: var(--wp--preset--font-family--monospace);
		font-size: var(--wp--preset--font-size--small);
		color: var(--wp--preset--color--syntax-red);
```

adamwoodnz (commented):
Looks good, one suggestion

adamwoodnz (approved):
LGTM 🚀

PR #416, opened 2023-11-29
Fix redundant empty result set messages

### LINKED ISSUE #338: Redundant empty result set messages
We show 2 messages that speak to the search results:

- `0 results found for “23rasdfasdf”.`
- `Sorry, but nothing matched your search terms.`. 

We probably don't need the first one when the result set is empty.

# Screenshot
Screenshot 2023-11-06 at 3 48 58 PM

outdoor2kode:
Fixes #338 

This PR fixes redundant empty result set messages by returning earlier when `$results_count === 0`.
`0 results found for “xxxxxxx”.` is removed.

## Screenshot
image

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/search-results-context/index.php
+++ b/source/wp-content/themes/wporg-developer-2023/src/search-results-context/index.php
@@ -22,8 +22,13 @@ function render( $attributes ) {
 		return '';
 	}
 
-	$posts_per_page = get_query_var( 'posts_per_page' );
 	$results_count = $wp_query->found_posts;
+
+	if ( 0 === $results_count ) {
+		return;
+	}
+
+	$posts_per_page = get_query_var( 'posts_per_page' );
 	$current_page = get_query_var( 'paged' ) ?: 1;
 	$first_result = ($current_page - 1) * $posts_per_page + 1;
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-developer-2023/src/search-results-context/index.php
```diff
@@ -22,8 +22,13 @@ function render( $attributes ) {
 		return '';
 	}
 
-	$posts_per_page = get_query_var( 'posts_per_page' );
 	$results_count = $wp_query->found_posts;
+
+	if ( 0 <= $results_count ) {
```
adamwoodnz: I think this should be changed otherwise the count will not be shown for any count over 0
```suggestion
	if ( 0 === $results_count ) {
```

adamwoodnz (changes requested):
Message is removed for all result counts at present, see suggestion inline 🙂

source/wp-content/themes/wporg-developer-2023/src/search-results-context/index.php
outdoor2kode: ah... it was `$results_count <= 0` in the first place, and then yoda standard came in..
Thanks.

adamwoodnz (approved):
LGTM :shipit:

PR #419, opened 2023-11-30
Fix content overflowing code blocks

### LINKED ISSUE #417: Content overflowing code blocks in Themes handbook
See https://developer.wordpress.org/redesign-test/themes/advanced-topics/theme-json/#theme-json-elements

Screenshot 2023-11-30 at 4 02 10 PM

outdoor2kode:
Fixes #417

I've added `<pre>` at the top so that by default, `<pre>` under the developer site will have `overflow:auto`, just like it was previously in [global](https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer-2023/scss/_global.scss#L96).

However, it seems like the SCSS files in this [directory](https://github.com/WordPress/wporg-developer/tree/trunk/source/wp-content/themes/wporg-developer-2023/scss) are no longer effective. The CSS doesn't change even after modifications. Are we looking to fix it?
Also, the `main.css` in the [CSS folder](https://github.com/WordPress/wporg-developer/tree/trunk/source/wp-content/themes/wporg-developer-2023/stylesheets) can probably be removed. It looks like we've replaced it with [style/style.scss](https://github.com/WordPress/wporg-developer/compare/fix/Content-overflowing-code-blocks?expand=1#diff-5a967de023463491bebbfaca0909be4be783855fcace73f96a9a0e9c8060224a).

## Screenshots

image

---

**User Notes Preview is Also Broken**
| Before | After |
|-|-|
| image | image |


adamwoodnz (commented):
👍 Works for me 

I notice that prism.css uses this rule too

Screenshot 2023-12-04 at 2 21 48 PM

adamwoodnz:
> However, it seems like the SCSS files in this [directory](https://github.com/WordPress/wporg-developer/tree/trunk/source/wp-content/themes/wporg-developer-2023/scss) are no longer effective. The CSS doesn't change even after modifications. Are we looking to fix it?
> Also, the `main.css` in the [CSS folder](https://github.com/WordPress/wporg-developer/tree/trunk/source/wp-content/themes/wporg-developer-2023/stylesheets) can probably be removed.

Indeed none of these are included in the build anymore, but we do load the prism stylesheet so maybe that should be:  https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer-2023/functions.php#L500

@StevenDufresne any context here please; did you leave the rest of these files in source control intentionally? We've come this far without them, seems like they're YAGNI now to me.

> It looks like we've replaced it with [style/style.scss](https://github.com/WordPress/wporg-developer/compare/fix/Content-overflowing-code-blocks?expand=1#diff-5a967de023463491bebbfaca0909be4be783855fcace73f96a9a0e9c8060224a).

👍 that's my understanding

adamwoodnz (approved):

StevenDufresne:
It's probably a relic of converting this from a classic theme.

outdoor2kode:
I've just cleaned up our style assets by removing all unused stylesheets, covering both CSS and SCSS files, and kept only those in use. 
Although the SCSS files of the in-use ones are not currently integrated into our build process, I've retained them. This is to facilitate their future reintegration into our build workflow, ensuring a smoother development experience.

PR #421, opened 2023-11-30
Improve UX of Related section

### LINKED ISSUE #371: Improve UX of Related section
The font size of file paths can be smaller.
A word should not be truncated and then wrapped to the next line.

What do you think @adamwoodnz, it doesn't appear that there's a clear [design instruction](https://www.figma.com/file/2WxlJFzMJvqPfZL1EkAOVp/Developer-Resources?type=design&node-id=2934-7281&mode=design&t=WHhMuRzmY0pMRv0q-0) about this. 

image

outdoor2kode:
Fixes #371

As suggested [here](https://github.com/WordPress/wporg-developer/issues/371#issuecomment-1812000401), set the file path font size as `14px` and the link as `word-break: break-word`.

## Screenshot

image

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/code-table/style.scss
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-table/style.scss
@@ -30,7 +30,7 @@
 
 		td a {
 			text-decoration: underline;
-			word-break: break-all;
+			word-break: break-word;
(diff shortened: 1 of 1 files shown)


adamwoodnz (approved):
:shipit: LGTM

PR #423, opened 2023-12-04
Fix the last updated date is not displaying

### LINKED ISSUE #398: The Last Updated date is not displaying on Code Reference docs
I'm not sure if this is a bug specifically with the Staging site, but I noticed that the Last Updated date is not displayed on Code Reference docs. Here's an [example](https://developer.wordpress.org/redesign-test/reference/hooks/edited_taxonomy/). 

image

outdoor2kode:
Fixes #398

Since direct use of PHP conditions in the pattern isn't possible, I've implemented a shortcode solution. Now, the "Last updated" text only appears when the modified date is later than the publish date.

## Test Steps
1. Find a page that has no "Last updated" shows up.
2. Update the page.
3. "Last updated" should be there.

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/inc/shortcodes.php
+++ b/source/wp-content/themes/wporg-developer-2023/inc/shortcodes.php
@@ -54,6 +54,9 @@ add_shortcode(
 	}
 );
 
+/**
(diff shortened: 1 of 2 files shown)


adamwoodnz (approved):
👍 Works for me in sandbox

PR #430, opened 2023-12-13
Remove extra p and a tags from the explanation content

### LINKED ISSUE #413: Remove extra <p> tags from the Explanation content
See discussion https://github.com/WordPress/wporg-developer/pull/411#issuecomment-1828830833 & https://github.com/WordPress/wporg-developer/pull/411#issuecomment-1830416747.

image

outdoor2kode:
Fixes #413 

Initially, I tried using regex to parse content, only processing `<p>` tags. Despite this, although many `<p>` tags were eliminated, there were still `<p></p>` tags generated above and below `<h2>`, `<h3>` tags. I think this is because of an attempt to render `<heading>` that's from classic editor in the block theme. Although I could also use another regex to process these `<heading>` tags, I think that would complicate the code. 

So, I started looking for a Gutenberg API that could directly convert content into blocks, but the `$parsed_blocks = parse_blocks($content);` + `$block_content = serialize_blocks($parsed_blocks);` approach didn't work.

Therefore, in the end, I thought about directly converting the classic editor into a GB editor and then using the 'convert to blocks' feature for the conversion.

image

This indeed successfully removed all the redundant `<p>` tags

--

Next, I addressed the issue of redundant `<a>` tags. The reason for these extra `<a>` tags primarily come from adding `$content = apply_filters( 'the_content', apply_filters( 'get_the_content', $content ) );`, and also from using the `@see` tag from phpDocumentor in the editor. 

So I removed the filter as well as got rid of the `{@see}` and left the original function name, the content can then be parsed normally without generating extra <a> tags.

## Screencast

In axeDevTools, all related issues have been resolved.

Screenshot 2023-12-13 at 21 30 05

Then, in Lighthouse, the score seems to have improved, but this is on a local basis. I notice there is still an excessive DOM size.

image
image

## Note

1. This method requires going into posts and clicking the Convert to Blocks button to switch the content. I am not aware of any function that can achieve this. I've browsed around but didn't find anything. Perhaps it's possible to refer to the Gutenberg source code [here](https://github.com/WordPress/gutenberg/blob/a51534b6b374879e602c41ee171563452ac9115f/packages/block-library/src/freeform/convert-to-blocks-button.js#L22) for this, but it might be faster to convert manually than to understand this code.

2. Then, in transitioning from the classic editor to the GB editor, the only difference I see is the disappearance of the 'Associated with' at the top of the edit screen. I feel this is okay as the 'Preview in new tab' feature of the Block Editor still allows viewing the complete page. Additionally, the code can be adjusted later to add it back.

image

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/inc/explanations.php
+++ b/source/wp-content/themes/wporg-developer-2023/inc/explanations.php
@@ -106,6 +106,7 @@ class WPORG_Explanations {
 			'supports'          => array( 'editor', 'revisions' ),
 			'rewrite'           => false,
 			'query_var'         => false,
+			'show_in_rest'      => true,
 		) );
 	}
 
--- a/source/wp-content/themes/wporg-developer-2023/inc/template-tags.php
+++ b/source/wp-content/themes/wporg-developer-2023/inc/template-tags.php
@@ -1848,9 +1848,6 @@ namespace DevHub {
 	function get_explanation_content( $_post ) {
 		global $post;
 
-		// Temporarily remove filter.
-		remove_filter( 'the_content', array( 'DevHub_Formatting', 'fix_unintended_markdown' ), 1 );
-
 		// Store original global post.
 		$orig = $post;
 
@@ -1874,17 +1871,14 @@ namespace DevHub {
 			$content = get_explanation_field( 'post_content', $_post );
 		}
 
-		// Pass the content through expected content filters.
-		$content = apply_filters( 'the_content', apply_filters( 'get_the_content', $content ) );
+		// Sanitize message content.
+		$content = wp_kses_post( $content );
+
+		// Remove the {@see} tag, leaving only the function name, so that there'd be no extra a tags being rendered.
+		$content = preg_replace( '/\{@see (.*?)\}/', '$1', $content );
 
 		// Restore original global post.
 		$post = $orig;
-
-		// Automatic paragraph and line break formatting in content.
-		add_filter( 'the_content', 'wpautop' );
-
-		// Restore filter.
-		add_filter( 'the_content', array( 'DevHub_Formatting', 'fix_unintended_markdown' ), 1 );
 
 		return $content;
 	}


adamwoodnz:
> 1. This method requires going into posts and clicking the Convert to Blocks button to switch the content

So I'm clear about the implications; we'd need to go into every explanation post and do this?

adamwoodnz (commented):
This works for me. I was able to convert my WP_Query explanation to blocks and the extra p and a tags were gone 👍 

Migrating the classic content to blocks seems like a good move too. We currently have 620 published explanations so it will take some time to get these all migrated though.

One thing to keep in mind is what happens if we have to switch back to the old theme (like we had to for Showcase)? Are blocks going to be compatible? I think this is unlikely, but we should at least consider what a roll back would look like.

adamwoodnz:
> 2\. Then, in transitioning from the classic editor to the GB editor, the only difference I see is the disappearance of the 'Associated with' at the top of the edit screen. I feel this is okay as the 'Preview in new tab' feature of the Block Editor still allows viewing the complete page. Additionally, the code can be adjusted later to add it back.

We could add this to the page details sidebar presumably.

adamwoodnz:
> One thing to keep in mind is what happens if we have to switch back to the old theme (like we had to for Showcase)? Are blocks going to be compatible? I think this is unlikely, but we should at least consider what a roll back would look like.

Thinking about this; as I understand it blocks are supposed to be saved as standard HTML so maybe it wouldn't be an issue. We could test by converting one production explanation then reverting it.

outdoor2kode:
> So I'm clear about the implications; we'd need to go into every explanation post and do this?

Yes. That's right. 👍 

> One thing to keep in mind is what happens if we have to switch back to the old theme (like we had to for Showcase)? Are blocks going to be compatible? I think this is unlikely, but we should at least consider what a roll back would look like.

I have tested it and found that blockized content can be rendered without errors in the current production theme (old theme). However, I think the best approach is to backup all the original classic content from the DB, and then run a script to update it back if necessary in case we need to roll back. 

I'll check to see if I have the permissions to do this part. The worst-case scenario is that we have to manually revert each Explanation post one by one. So I will also backup a copy of the classic content locally.

> We could add this to the page details sidebar presumably.

Yeah, that's what I'm thinking. I'll see if I can do this today. If not, I'll log it.

outdoor2kode:
> Thinking about this; as I understand it blocks are supposed to be saved as standard HTML so maybe it wouldn't be an issue. We could test by converting one production explanation then reverting it.

Ah, yes. This is the way I tested it. You can see the Revision [here](https://developer.wordpress.org/wp-admin/post.php?post=142980&action=edit).

outdoor2kode:
> We could add this to the page details sidebar presumably.
> > Yeah, that's what I'm thinking. I'll see if I can do this today. If not, I'll log it.

Issue opened https://github.com/WordPress/wporg-developer/issues/436

I've noticed that it can't be added to the GB editor sidebar simply through `add_meta_box` - it would be rendered on the bottom. It needs to enqueue a script and leverage GB API.

I'll continue working on it, just opened that issue so it won't be a blocker.

outdoor2kode:
I'll merge this PR first as we're going to launch tmr and it seems there'd be no further feedback coming in.

PR #435, opened 2023-12-14
Modify callout icons and spacing.

ndiego:
In reviewing the updated callouts, I completely overlooked the icons. Also the icons in [Figma](https://www.figma.com/file/HOJTpCFfa3tR0EccUlu0CM/WordPress.org-Design-Library?type=design&node-id=1264-4116&mode=design&t=thQtAIFDBWAgbIdR-0) do not appear to be correct. 

The info callout should use the `i` icon, which means we can then use the `?` icon for the default. I also added a bit more spacing between the icon and the text.

| Before | After |
|-|-|
|image|image|

> [!NOTE]
> A similar PR will be needed in `wporg-mu-plugins` to update the [Notice block styles](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/notice/postcss/style.pcss) if this PR is merged.


fcoveram:
Thanks for noticing this. I fixed the error in Figma, and the spacing update looks good to me 🚀

outdoor2kode (approved):
Thanks for catching this! Looks good to me 👍

PR #446, opened 2023-12-15
Add redirects for docs changes and update formatting

ndiego:
This PR adds redirects for a few recent changes to the Block Editor Handbook. 

- https://github.com/WordPress/gutenberg/pull/56931
- https://github.com/WordPress/gutenberg/pull/56584
- https://github.com/WordPress/gutenberg/pull/57120

> [!NOTE]
> Many of the redirects go to the same destination, so if there is a more efficient way of doing this, just let me know, and I'll update the PR.

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/inc/import-block-editor.php
+++ b/source/wp-content/themes/wporg-developer-2023/inc/import-block-editor.php
@@ -49,24 +49,44 @@ class DevHub_Block_Editor_Importer extends DevHub_Docs_Importer {
 
 		// Any handbook pages where the slug changes should be listed here.
 		$redirects = [
-			'components/server-side-render' => 'packages/packages-server-side-render',
-			'explanations/faq' => 'getting-started/faq',
-			'explanations/glossary' => 'getting-started/glossary',
-			'how-to-guides/block-theme' => 'how-to-guides/themes/create-block-theme',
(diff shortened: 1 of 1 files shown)


adamwoodnz (approved):
Tested in sandbox; existing redirects work going to new locations. New redirects won't work until the old pages 404, but new locations look good.

pkevan:
> Many of the redirects go to the same destination, so if there is a more efficient way of doing this, just let me know

For redirects it's usually best to be explicit with them, rather than do some fancy regex or wildcard, so the method was good 👍

ndiego:
@pkevan do you know if this has been deployed yet? I have tested setting https://developer.wordpress.org/block-editor/getting-started/create-block/finishing/ to draft. This should redirect to https://developer.wordpress.org/block-editor/getting-started/tutorial, but I am still getting a 404.

pkevan:
Not sure - will take a look.

pkevan:
I think it might be the structure of the handbook page - presumably they all need to be draft in order for the redirects to work. i.e. `create-block` parent item?

adamwoodnz:
Debugged and found the issue; the paths in $redirects shouldn't have a trailing slash. I'll push a fix.

ndiego:
Ah, thanks @adamwoodnz. Sorry about that!

adamwoodnz:
Deployed, appears to be fixed

PR #447, opened 2023-12-16
Update inline code padding and use border-radius variable

ndiego:
This is a minor tweak to inline code styles. Previously, the padding was hardcoded and did not adjust based on the size of the font. There are a few instances where inline code is used in headings. 

This PR updates the left and right padding to `0.25em`. This translates to a padding increase of `1px` to `4px` for article copy. 

While adjusting the padding, I also updated all instances of `border-radius: 2px` to use the custom variable.

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
+++ b/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
@@ -182,10 +182,7 @@ pre {
 
 // Style code tags but not ones inside of the code block.
 .wporg-has-embedded-code *:not(.wp-block-code):not(.wp-code-block-button-container) > code {
-	padding: 4px 6px;
-	background-color: var(--wp--preset--color--light-grey-2);
 	font-size: var(--wp--preset--font-size--small);
-	border-radius: var(--wp--custom--button--border--radius);
 }
 
 .wporg-dot-link-list {
--- a/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
+++ b/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
@@ -32,7 +32,7 @@ pre {
 		padding: 20px;
 		background-color: #f7f7f7;
 		border: 1px solid var(--wp--preset--color--light-grey-1);
-		border-radius: var(--wp--custom--button--border--radius);
+		border-radius: var(--wp--custom--code--border--radius);
 		overflow: scroll;
 	}
 
@@ -40,7 +40,7 @@ pre {
 		display: inline-block;
 		line-height: var(--wp--custom--body--extra-small--typography--line-height);
 		background: var(--wp--preset--color--light-grey-2);
-		border-radius: var(--wp--custom--button--border--radius);
+		border-radius: var(--wp--custom--code--border--radius);
 		padding-inline-start: 0.25em;
 		padding-inline-end: 0.25em;
 		max-width: 100%;
@@ -432,7 +432,7 @@ pre {
 
 .wporg-developer-code-block {
 	$half_padding: calc(var(--wp--preset--spacing--10) / 2);
-	$border_radius: var(--wp--custom--button--border--radius);
+	$border_radius: var(--wp--custom--code--border--radius);
 
 	.wp-code-block-button-container {
 		background-color: var(--wp--preset--color--light-grey-2);
--- a/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
+++ b/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
@@ -32,7 +32,7 @@ pre {
 		padding: 20px;
 		background-color: #f7f7f7;
 		border: 1px solid var(--wp--preset--color--light-grey-1);
-		border-radius: 2px;
+		border-radius: var(--wp--custom--button--border--radius);
 		overflow: scroll;
 	}
 
@@ -40,9 +40,9 @@ pre {
 		display: inline-block;
 		line-height: var(--wp--custom--body--extra-small--typography--line-height);
 		background: var(--wp--preset--color--light-grey-2);
-		border-radius: 2px;
-		padding-inline-start: 3px;
-		padding-inline-end: 3px;
+		border-radius: var(--wp--custom--button--border--radius);
+		padding-inline-start: 0.25em;
+		padding-inline-end: 0.25em;
 		max-width: 100%;
 	}
 
@@ -185,7 +185,7 @@ pre {
 	padding: 4px 6px;
 	background-color: var(--wp--preset--color--light-grey-2);
 	font-size: var(--wp--preset--font-size--small);
-	border-radius: 2px;
+	border-radius: var(--wp--custom--button--border--radius);
 }
 
 .wporg-dot-link-list {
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-developer-2023/src/style/style.scss
```diff
@@ -32,17 +32,17 @@ pre {
 		padding: 20px;
 		background-color: #f7f7f7;
 		border: 1px solid var(--wp--preset--color--light-grey-1);
-		border-radius: 2px;
+		border-radius: var(--wp--custom--button--border--radius);
```
adamwoodnz: While this is the same value, semantically it's a different use, and I think we should add a new custom var
```suggestion
		border-radius: var(--wp--custom--code--border--radius);
```

source/wp-content/themes/wporg-developer-2023/src/style/style.scss
```diff
@@ -32,17 +32,17 @@ pre {
...
 	code {
 		display: inline-block;
 		line-height: var(--wp--custom--body--extra-small--typography--line-height);
 		background: var(--wp--preset--color--light-grey-2);
-		border-radius: 2px;
-		padding-inline-start: 3px;
-		padding-inline-end: 3px;
+		border-radius: var(--wp--custom--button--border--radius);
+		padding-inline-start: 0.25em;
+		padding-inline-end: 0.25em;
```
adamwoodnz, line 45: [This rule](https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer-2023/src/style/style.scss#L185) is overriding these values for inline code, eg. in the [More Information section](https://developer.wordpress.org/reference/classes/wp_query/#more-information). Could you try removing it?

outdoor2kode: How does 
```
.wp-block-search div.awesomplete > ul {
	left: unset;
	inset-inline-start: 0;
}
```
affect the value here?

outdoor2kode: Oh, wait. I guess you're referring to [this one](https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer-2023/src/style/style.scss#L192)
```
padding: 4px 6px;
```
Must be the code has been updated.

source/wp-content/themes/wporg-developer-2023/src/style/style.scss
outdoor2kode: Are you suggesting adding style in the parent theme [here](https://github.com/WordPress/wporg-parent-2021/blob/trunk/source/wp-content/themes/wporg-parent-2021/theme.json#L142-L146)? If so, do you think we should also include the complete styling for the code here, like the color of the code, padding, etc.? Perhaps we can ask the design team to provide a standard design for the code.

And what do you think we should call [this](https://github.com/WordPress/wporg-developer/blob/16766c3e87ceed66393f5f84ff4a31f442f74801/source/wp-content/themes/wporg-developer-2023/src/style/style.scss#L636)? I reckon a callout is not a typical button or code?

```
"code": {
	"border": {
		"color": "var(--wp--preset--color--light-grey-1)",
		"radius": "2px",
		"style": "solid",
		"width": "1px"
	}
},
```

Or you're just referring to adding something like [these two](https://github.com/WordPress/wporg-developer/blob/16766c3e87ceed66393f5f84ff4a31f442f74801/source/wp-content/themes/wporg-developer-2023/src/style/style.scss#L629-L630).

source/wp-content/themes/wporg-developer-2023/src/style/style.scss
renintw: I've just replaced the `padding: 4px 6px;` with ```padding-inline-start: 0.25em;
		padding-inline-end: 0.25em;```

You can look for `.wporg-has-embedded-code *:not(.wp-block-code):not(.wp-code-block-button-container) > code` elements on http://localhost:8895/reference/classes/wp_query/ to see if those affected are expected.

outdoor2kode: I've removed it as well as other redundant styles in d04799c.

source/wp-content/themes/wporg-developer-2023/src/style/style.scss
adamwoodnz:
```javascript
"code": {
	"border": {
		"radius": "2px",
	}
},
``` 

Yep exactly that. I guess it makes sense to add it to parent, even if most child themes might not use it.

adamwoodnz: > And what do you think we should call [this](https://github.com/WordPress/wporg-developer/blob/16766c3e87ceed66393f5f84ff4a31f442f74801/source/wp-content/themes/wporg-developer-2023/src/style/style.scss#L636)? I reckon a callout is not a typical button or code?

Callouts are definitely different, not sure what you're asking here sorry... can you expand on it?

adamwoodnz:
> There are a few instances where inline code is used in headings.

Could you please give an example for testing?

source/wp-content/themes/wporg-developer-2023/src/style/style.scss
outdoor2kode: Ah, yeah. I'm referring to the `button` in the `--wp--custom--button--border--radius`.
Since this border radius belongs to a callout, I assume we also want it to be something like `--wp--custom--callout--border--radius`.

outdoor2kode: The custom code block style has been added to [parent](https://github.com/WordPress/wporg-parent-2021/pull/125). And the button has also been replaced with code here through 39fb3f1.

## Results
image
image

outdoor2kode:
> Could you please give an example for testing?

@ndiego icymi.

source/wp-content/themes/wporg-developer-2023/src/style/style.scss
adamwoodnz: > Ah, yeah. I'm referring to the `button` in the `--wp--custom--button--border--radius`. Since this border radius belongs to a callout, I assume we also want it to be something like `--wp--custom--callout--border--radius`.

Sorry I get it now. Yeah I don't think this change should be included, unless we're going to update the [mu-plugins callout styles](url) to also use a var.

source/wp-content/themes/wporg-developer-2023/src/style/style.scss
```diff
@@ -633,7 +630,7 @@ pre {
 	color: var(--wp--custom--wporg-callout--color--text);
 	background-color: var(--wp--custom--wporg-callout--color--background);
 	border-width: 0;
-	border-radius: 2px;
+	border-radius: var(--wp--custom--button--border--radius);
```
adamwoodnz, line 633: I think we should revert this, as it's a value copied from mu-plugins, see [comment](https://github.com/WordPress/wporg-developer/pull/447#discussion_r1495163913).

source/wp-content/themes/wporg-developer-2023/src/style/style.scss
```diff
@@ -435,7 +432,7 @@ pre {
 
 .wporg-developer-code-block {
 	$half_padding: calc(var(--wp--preset--spacing--10) / 2);
-	$border_radius: 2px;
+	$border_radius: var(--wp--custom--code--border--radius);
```
adamwoodnz, line 435: This isn't inline code, rather it's the container, so I don't think it should use this var. Reason being; if I were to change this variable to increase the inline code style, I don't think I'd expect these container styles to also change.

Screenshot 2024-02-20 at 3 21 02 PM

adamwoodnz (changes requested):
Sorry this has gone back and forth, but I think we should reduce the scope of this. Variablizing all 2px values doesn't make sense to me without creating more granular vars.

PR #448, opened 2023-12-16
Add blockquote styles

### LINKED ISSUE #390: Add `<blockquote>` styling
Currently, `<blockquote>` only have default browser styles. A good example is the [Deprecations page](https://developer.wordpress.org/redesign-test/block-editor/reference-guides/block-api/block-deprecation/). 

@jasmussen @fcoveram do we have a standard design that could be applied here?

We have a design on Showcase, but not sure that would work well here. Perhaps just the blue line with no additional text styles?

| Current | Showcase |
|-|-|
|image|image|

ndiego:
Fixes #390 

Adds `blockquote` styling. 

| Before | After |
|-|-|
|image|image|

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
+++ b/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
@@ -28,6 +28,13 @@ pre {
 		}
 	}
 
+	blockquote {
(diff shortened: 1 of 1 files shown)


adamwoodnz (approved):
Works for me 👍

PR #449, opened 2023-12-16
Increase banner font-size on Front Page

ndiego:
Given the font-size, I have found the "See what has changed in the WordPress 6.4 API." banner on the Front Page very hard to read. 

This PR bumps up the size to `small` instead of `extra-small`.

|Before|After|
|-|-|
|image|image|


adamwoodnz (approved):
LGTM

PR #454, opened 2023-12-22
Abstract article meta GitHub for use in different contexts

StevenDufresne:
I added this abstraction so we can use this component in other themes (documentation/data-liberation). 

Ideally, once this merges, we could then move it to the [wporg-mu-plugin](https://github.com/wordpress/wporg-mu-plugins) repository. I didn't want to make those changes in one large PR.

## Changes
- Pass in the `heading`, `text` and `linkURL`.
- Remove logic related to conditional displaying of the block.
- Add block hooks to handle the conditional rendering.

## Considerations
In order to check if the block is being used for the changelog, which doesn't exist on the block-editor handbook, I had to use the url in the check:

https://github.com/WordPress/wporg-developer/pull/454/files#diff-d2b871ec9cde93c4a3003771e7339d18ebe322f91a523dd3c9405aa95b61b5f3R81

I don't think that is ideal since that could change in the future. I thought about adding an ID attribute but that could get out of sync as well. I'm figuring that it should be good enough.

What do ya'll think?

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/article-meta-github/block.json
+++ b/source/wp-content/themes/wporg-developer-2023/src/article-meta-github/block.json
@@ -8,7 +8,16 @@
 	"icon": "smiley",
 	"description": "",
 	"usesContext": [ "postId", "postType" ],
-	"attributes": {},
+	"attributes": {
+		"handbooks": {
+			"type": "array",
+			"default": []
+		},
+		"editLinkURL": {
+			"type": "string",
+			"default": ""
+		}
+	},
 	"supports": {
 		"inserter": false
 	},
--- a/source/wp-content/themes/wporg-developer-2023/src/article-meta-github/block.json
+++ b/source/wp-content/themes/wporg-developer-2023/src/article-meta-github/block.json
@@ -9,15 +9,15 @@
 	"description": "",
 	"usesContext": [ "postId", "postType" ],
 	"attributes": {
-		"linkUrl": {
-			"type": "string",
-			"default": ""
-		},
 		"heading": {
 			"type": "string",
 			"default": ""
 		},
-		"text": {
+		"linkText": {
+			"type": "string",
+			"default": ""
+		},
+		"linkUrl": {
 			"type": "string",
 			"default": ""
 		}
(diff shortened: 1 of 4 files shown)


source/wp-content/themes/wporg-developer-2023/src/article-meta-github/block.php
```diff
@@ -60,45 +51,19 @@ function render( $attributes, $content, $block ) {
...
 				<!-- /wp:paragraph -->
 
 			</div>
 			<!-- /wp:group -->'
 		),
-		esc_html__( 'Edit article', 'wporg' ),
+		esc_html__( $attributes['heading'], 'wporg' ),
+		esc_url( $attributes['linkURL'] ),
 		sprintf(
 			/* translators: %s: article title */
```
adamwoodnz: Comment needs updating

source/wp-content/themes/wporg-developer-2023/src/article-meta-github/block.json
```diff
@@ -8,7 +8,20 @@
...
+	"attributes": {
+		"linkUrl": {
+			"type": "string",
+			"default": ""
+		},
+		"heading": {
+			"type": "string",
+			"default": ""
+		},
+		"text": {
```
adamwoodnz: Maybe this would make the purpose of the content clearer
```suggestion
		"linkText": {
```

source/wp-content/themes/wporg-developer-2023/src/article-meta-github/block.php
```diff
@@ -60,45 +51,19 @@ function render( $attributes, $content, $block ) {
...
-			__( 'Improve it on GitHub<span class="screen-reader-text">: %s</span>', 'wporg' ),
+			__( '%1$s<span class="screen-reader-text">: %2$s"</span>', 'wporg' ),
+			esc_html( $attributes['text'] ),
```
adamwoodnz:
```suggestion
			esc_html( $attributes['linkText'] ),
```

adamwoodnz (approved):
> I'm figuring that it should be good enough.

> What do ya'll think?

👍 Yeah I agree, LGTM

Issue #456, opened 2023-12-27
The "Last updated" date is incorrect in Handbooks that source their content from GitHub


ndiego:
In handbooks where the content lives in GitHub, the "Last updated" displays the last time the content was synced from GitHub. The [Block Editor Handbook](https://developer.wordpress.org/block-editor/getting-started/quick-start-guide/) is a good example. 

The image below is from the [Quick Start Guide](https://developer.wordpress.org/block-editor/getting-started/quick-start-guide/). The actual content was last updated on 11/19, but the date is showing today.

<img width="509" alt="image" src="https://github.com/WordPress/wporg-developer/assets/4832319/ac615162-8124-41de-a5a7-7b7348e71db5">

adamwoodnz:
We had a previous [issue](https://github.com/WordPress/wporg-developer/issues/345) to remove this info for the Block Editor handbook (although this should probably apply to all GitHub handbooks), but there must have been a regression and it has been added back. Do we want to try to get the actual last edit date from GitHub, or just remove it again?

ndiego:
It would be great to get the actual date if possible. But perhaps we just remove it for the time being while an alternative implementation is explored? I don't know how hard it would be to get the actual date.

adamwoodnz:
> But perhaps we just remove it for the time being while an alternative implementation is explored? I don't know how hard it would be to get the actual date.

Yeah I think this sounds like a plan. Looks like we can [query for commits using the GitHub API](https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#list-commits-on-a-repository), and get the date for the last commit.

StevenDufresne:
A couple questions to help identify the solution:
- Why does that matter to the user who assumes this is the canonical information source?
- Does the sync update all dates or only the documents that were modified?

ndiego:
> Why does that matter to the user who assumes this is the canonical information source?

Since we have a lot of outdated docs (this is being worked on), the fact that the "Last updated" date shows the current date is quite misleading. 

> Does the sync update all dates or only the documents that were modified?

To my knowledge, the entire Handbook is synced every 15 min, so the "Last updated" date always shows the current date. Another option would be to only sync if a change is detected, but not sure what's the easiest solution.

adamwoodnz:
> Does the sync update all dates or only the documents that were modified?

> Another option would be to only sync if a change is detected, but not sure what's the easiest solution.

[...]

PR #462, opened 2024-01-03
Remove GitHub handbook last updated date 

adamwoodnz:
See https://github.com/WordPress/wporg-developer/issues/456

The article meta Last Updated field currently uses the modified date, but for GitHub handbooks this is the import date, not the GitHub edited date. This PR temporarily removes it, until we can get the actual edited date of the markdown file from GitHub. 

It follows the approach from https://github.com/WordPress/wporg-developer/pull/454, creating a block to be used for both dates, then filtering to remove the last updated date on GitHub handbooks only. The local handbooks for Plugins and Themes are unaffected.

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/inc/block-hooks.php
+++ b/source/wp-content/themes/wporg-developer-2023/inc/block-hooks.php
@@ -52,32 +52,37 @@ function get_block_content_by_home_url( $block_content, $replacement_home_url = 
 }
 
 /**
- * Filters the article meta block and conditionally removes it.
+ * Filters an article meta block and conditionally removes it.
  *
  * @param string $block_content
  * @param array  $block
  * @return string
  */
-function filter_article_meta_block( $block_content, $block ) {
-	if ( 'wporg/article-meta-github' === $block['blockName'] ) {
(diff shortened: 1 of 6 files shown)


source/wp-content/themes/wporg-developer-2023/inc/block-hooks.php
```diff
@@ -52,26 +52,20 @@ function get_block_content_by_home_url( $block_content, $replacement_home_url =
...
-		$github_handbooks = array(
-			'wpcs-handbook',
-			'blocks-handbook',
-			'rest-api-handbook',
-			'cli-handbook',
-			'adv-admin-handbook',
-		);
+function filter_article_meta_block( $block_content, $block ) {	
+	// Not all handbooks come from GitHub.
+	$local_handbooks = array( 'plugin-handbook', 'theme-handbook' );
```
adamwoodnz, line 63: Reverse this logic based on the assumption that new handbooks added would be GitHub based, and this logic would not have to be changed. GitHub handbooks allow community updates.

StevenDufresne (approved):
LGTM.

PR #467, opened 2024-01-09
Search Results Block: Strip markdown from excerpts

### LINKED ISSUE #465: Blockquote notices break on the search results page
When an article contains a blockquote, there is a problem with how these are presented on the search results page.
The excerpt includes the HTML for the notice, which interrupts the text and breaks the layout.

[Example link](https://developer.wordpress.org/block-editor/?s=deprecation)

Screenshot:
Search results with broken excerpt

Markup:

```
<div class="wp-block-post-excerpt">
<p class="wp-block-post-excerpt__excerpt">
&gt; This page provides a comprehensive guide to the principles and usage of the Deprecation API. For an introduction check out the 
</p><div class="wp-block-wporg-notice is-tip-notice">
<div class="wp-block-wporg-notice__icon"></div>
<div class="wp-block-wporg-notice__content"></div></div>
(https://developer.wordpress.org/news/2023/03/block-deprecation-a-tutorial/) which can be found on the [Developer Blog](https://developer.wordpress.org/news/). <p></p></div>
```

Expected:

This page provides a comprehensive guide to the principles and usage of the Deprecation API. For an introduction check out the [tutorial on the basics of block deprecation](https://developer.wordpress.org/news/2023/03/block-deprecation-a-tutorial/) which can be found on the [Developer Blog](https://developer.wordpress.org/news/).

ryelle:
Fixes #465 — The markdown rendered in the excerpt matches one of the legacy notice shortcodes (`[tutorial]`), which is causing the shortcode processing to output here. It's also unexpected that the markdown would output in the search results, excerpts are typically plain text. This update parses the markdown and strips the resulting HTML, so that these excerpts behave more like native WordPress excerpts.

| Before | After |
|---|---|
| search-results-before | search-results-after |
| ref-search-before | ref-search-after |

To test:

- View a handbook search result, for example `/block-editor/?s=deprecation`
- There should be no broken-looking notices
- There should be no markdown

Try on the code reference too, there should be no change.

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/search-post/index.php
+++ b/source/wp-content/themes/wporg-developer-2023/src/search-post/index.php
@@ -11,6 +11,7 @@ namespace WordPressdotorg\Theme\Developer_2023\Dynamic_Search_Post;
 use function DevHub\is_parsed_post_type;
 
 add_action( 'init', __NAMESPACE__ . '\init' );
+add_filter( 'get_the_excerpt', __NAMESPACE__ . '\strip_markdown_in_excerpt' );
 
 /**
  * Registers the block using the metadata loaded from the `block.json` file.
@@ -26,6 +27,24 @@ function init() {
 			'render_callback' => __NAMESPACE__ . '\render',
 		)
 	);
+}
+
+/**
+ * Parse the markdown in a given string, then strip the resulting HTML.
+ *
+ * @param string $excerpt The post excerpt.
+ *
+ * @return string A string without markdown or HTML.
+ */
+function strip_markdown_in_excerpt( $excerpt ) {
+	// Load Jetpack Markdown if it's not already loaded.
+	if ( ! class_exists( 'WPCom_GHF_Markdown_Parser' ) && defined( 'JETPACK__PLUGIN_DIR' ) ) {
+		require_once JETPACK__PLUGIN_DIR . '/_inc/lib/markdown.php';
+	}
+
+	$parser = new \WPCom_GHF_Markdown_Parser();
+	$html = $parser->transform( $excerpt );
+	return wp_strip_all_tags( $html );
 }
 
 /**


adamwoodnz (commented):
Works for me, but I have a question (maybe my understanding of excerpt is off); why is the excerpt based on the markdown? The post content has already be parsed to html, I think this happens on import, should we also be generating an excerpt on import?

ryelle:
Good question, I also spent some time tracking down where the excerpt is generated. The import parsing starts [around here](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/plugins/wporg-markdown/inc/class-importer.php#L472), and on line 490 the first sentence of markdown content is used. Maybe it was assumed it will always be plain text? Only after that is the markdown content is parsed to HTML, and that's saved as the post content.

I suppose we could flip those, and pull out the first sentence after parsing as HTML, instead. Strip out the tags, then save it.

adamwoodnz:
Ah nice, yeah that assumption sounds accurate.

> I suppose we could flip those, and pull out the first sentence after parsing as HTML, instead. Strip out the tags, then save it.

Yeah imo it would be more efficient to save the expected content once, at import time, rather than on every page load.

ryelle:
Okay, I've made an update to [wporg-markdown](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/plugins/wporg-markdown/inc/class-importer.php), so that the excerpt processing runs after the HTML parsing.

```diff
Index: inc/class-importer.php
===================================================================
--- inc/class-importer.php	(revision 13104)
+++ inc/class-importer.php	(working copy)
@@ -486,12 +486,6 @@
 		}
 		$markdown = trim( $markdown );
 
-		// Use the first sentence as the excerpt.
-		$excerpt = '';
-		if ( preg_match( '/^(.+)/', $markdown, $matches ) ) {
-			$excerpt = $matches[1];
-		}
-
 		// Transform to HTML and save the post
 		$parser = new WPCom_GHF_Markdown_Parser();
 		$parser->preserve_shortcodes = false;
@@ -499,6 +493,12 @@
 
 		$html = apply_filters( 'wporg_markdown_after_transform', $html, $this->get_post_type() );
 
+		// Use the first line as the excerpt, but first strip any HTML.
+		$excerpt = '';
+		if ( preg_match( '/^(.+)/', wp_strip_all_tags( $html ), $matches ) ) {
+			$excerpt = $matches[1];
+		}
+
 		add_filter( 'wp_kses_allowed_html', [ $this, 'wp_kses_allow_links' ], 10, 2 );
 
 		$post_data = array(
```

You can re-parse all handbook pages by adding `add_filter( 'wporg_markdown_check_etags', '__return_false' );`, then running `yarn wp-env run cli "wp cron event run --all"`

adamwoodnz:
> Okay, I've made an update to [wporg-markdown](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/plugins/wporg-markdown/inc/class-importer.php), so that the excerpt processing runs after the HTML parsing.

I tried to update my local plugin but couldn't so I just applied this patch and it worked 👍 

I don't see the change on [trunk](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/plugins/wporg-markdown/inc/class-importer.php#L489), assume you still need to ship it?

ryelle:
> I don't see the change on [trunk](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/plugins/wporg-markdown/inc/class-importer.php#L489), assume you still need to ship it?

That's right, I added the diff here for "review" since that plugin is trac-based. I figured if that looks good, I'd ship that patch & close this PR.

adamwoodnz (approved):
Patch for wporg-markdown LGTM

ryelle:
Merged into wporg-markdown: https://meta.trac.wordpress.org/changeset/13119

PR #469, opened 2024-01-12
Chapters and ToC: Internal scroll

### LINKED ISSUE #453: ToC and Chapter list iteration 1
There are multiple existing issues open for these components, so this issue aims to address all these together, settle on a new design and implement the changes. Summary of issues:

1. ToC isn't sticky when it's longer than the viewport https://github.com/WordPress/wporg-developer/issues/445 https://github.com/WordPress/wporg-developer/issues/365
2. Chapter list doesn't scroll with the page https://github.com/WordPress/wporg-developer/issues/341
3. Header hit area to expand on mobile is only the caret https://github.com/WordPress/wporg-developer/issues/434

## Technical considerations

Currently the [Chapter list](https://github.com/WordPress/wporg-developer/tree/trunk/source/wp-content/themes/wporg-developer-2023/src/chapter-list) is a custom block in the Developer theme. The [Table of Contents](https://github.com/WordPress/wporg-mu-plugins/tree/trunk/mu-plugins/blocks/table-of-contents) (and the [Sidebar Container](https://github.com/WordPress/wporg-mu-plugins/tree/trunk/mu-plugins/blocks/sidebar-container), which actually controls the sticky behaviour) are custom blocks in the wporg-mu-plugins repo. The Sidebar Container and Table of Contents blocks are also used on the [Documentation single pages](https://wordpress.org/documentation/article/hosting-wordpress/). To streamline development it would be worth consolidating the scrolling behaviour and header in a new or refactored mu-plugins block which is used for both.

adamwoodnz:
Closes #453 

Depends on https://github.com/WordPress/wporg-mu-plugins/pull/554, which makes the sidebar container scroll internally once it becomes fixed.

Wraps the Chapters List in a sidebar container so that it behaves the same as the ToC, but without the 'Back to top' link.

On the Code Reference the comment form rules have been moved under the form so that they no longer clash with the ToC. It now only has to reduce in height when the footer becomes visible, see [discussion](https://github.com/WordPress/wporg-developer/issues/453#issuecomment-1867192556).

The scrolling areas have custom scrollbars so that they appear on hover. This has been inspired by the React docs ([example](https://react.dev/reference/react/legacy)), and should help users discover that these are scrollable more easily. With the typical Mac OS setting of hidden scrollbars, it is not immediately obvious. Note that this does not work in Firefox.

#### Code reference

https://github.com/WordPress/wporg-developer/assets/1017872/aa029416-67eb-4ec0-84e4-a93122193758

#### Block Editor

https://github.com/WordPress/wporg-developer/assets/1017872/f72896fc-a8f3-41da-b259-23e1e671a1a5

### Testing

1. Ensure your mu-plugins is on https://github.com/WordPress/wporg-mu-plugins/pull/554
2. Test pages with short ToC and/or Chapter List
3. Test pages with long ToC and/or Chapter List
4. Test for regressions on small screens, there should be no change
5. Test in different browsers (check scrollbars)

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
+++ b/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
@@ -85,7 +85,7 @@ pre {
 		flex-direction: row !important;
 		column-gap: var(--local--column-gap);
 
-		/* Left sidebar */
+		/* Left sidebar: Chapter List */
 		> .wp-block-wporg-sidebar-container {
 			width: var(--local--sidebar--width);
 			inset-inline-start: var(--wp--preset--spacing--edge-space);
@@ -109,7 +109,7 @@ pre {
 			margin-left: auto;
 			margin-right: auto;
 
-			/* Right sidebar */
+			/* Right sidebar: ToC */
 			.wp-block-wporg-sidebar-container {
 				inset-inline-end: var(--wp--preset--spacing--edge-space);
 			}
--- a/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
+++ b/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
@@ -71,6 +71,7 @@ pre {
 .has-three-columns {
 	--local--sidebar--width: 232px;
 	--local--column-gap: 40px;
+	justify-content: flex-end;
 
 	> * {
 		width: 100%;
@@ -84,8 +85,10 @@ pre {
 		flex-direction: row !important;
 		column-gap: var(--local--column-gap);
 
-		aside {
+		/* Left sidebar */
+		> .wp-block-wporg-sidebar-container {
 			width: var(--local--sidebar--width);
+			inset-inline-start: var(--wp--preset--spacing--edge-space);
 		}
 
 		main {
@@ -105,10 +108,11 @@ pre {
 			max-width: 960px;
 			margin-left: auto;
 			margin-right: auto;
-		}
 
-		.wp-block-wporg-sidebar-container {
-			inset-inline-end: var(--wp--preset--spacing--edge-space);
+			/* Right sidebar */
+			.wp-block-wporg-sidebar-container {
+				inset-inline-end: var(--wp--preset--spacing--edge-space);
+			}
 		}
 	}
 }
--- a/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
+++ b/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
@@ -69,7 +69,7 @@ pre {
 }
 
 .has-three-columns {
-	--local--sidebar--width: 232px;
+	--local--sidebar--width: 248px;
 	--local--column-gap: 40px;
 	justify-content: flex-end;
 
--- a/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
+++ b/source/wp-content/themes/wporg-developer-2023/src/style/style.scss
@@ -81,11 +81,20 @@ pre {
 		order: 1;
 	}
 
+	/* Left sidebar: Chapter List */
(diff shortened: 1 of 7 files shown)


fcoveram:
It works great. Excellent improvement 👏

ryelle:
My Chapters sidebar looks like this, it looks like there's still a position:absolute coming from somewhere. Once scrolling, the sidebars both look correct.



In the two-column layout, the chapter sidebar doesn't scroll (screenshot just to explain what layout I mean). I would expect it to continue to scroll even though the other sidebar is collapsed.

outdoor2kode:
> In the two-column layout, the chapter sidebar doesn't scroll (screenshot just to explain what layout I mean). I would expect it to continue to scroll even though the other sidebar is collapsed.

I saw this as well. 

> My Chapters sidebar looks like this, it looks like there's still a position:absolute coming from somewhere. Once scrolling, the sidebars both look correct.

But I was unable to reproduce this one, though.

source/wp-content/themes/wporg-developer-2023/src/style/style.scss
```diff
@@ -85,7 +85,7 @@ pre {
 		flex-direction: row !important;
 		column-gap: var(--local--column-gap);
 
-		/* Left sidebar */
```
renintw: 👍👍👍

source/wp-content/themes/wporg-developer-2023/src/style/style.scss
```diff
@@ -84,8 +85,10 @@ pre {
 		flex-direction: row !important;
 		column-gap: var(--local--column-gap);
 
-		aside {
+		/* Left sidebar: Chapter List */
```
outdoor2kode: 👍👍👍

adamwoodnz:
> My Chapters sidebar looks like this, it looks like there's still a position:absolute coming from somewhere. Once scrolling, the sidebars both look correct.

Fixed the postcss nesting issues you detailed in https://github.com/WordPress/wporg-mu-plugins/pull/554 so should be fixed now if you update mu-plugins

adamwoodnz:
> In the two-column layout, the chapter sidebar doesn't scroll (screenshot just to explain what layout I mean). I would expect it to continue to scroll even though the other sidebar is collapsed.

Nice catch, I'll make this behaviour independent

adamwoodnz:
> In the two-column layout, the chapter sidebar doesn't scroll (screenshot just to explain what layout I mean). I would expect it to continue to scroll even though the other sidebar is collapsed.

I have now fixed this in https://github.com/WordPress/wporg-mu-plugins/pull/554 and you should be able to retest. [More details >](https://github.com/WordPress/wporg-mu-plugins/pull/554#issuecomment-1916039040)

https://github.com/WordPress/wporg-developer/assets/1017872/14463f0a-3eb0-48c6-bfd1-eabb7fed9c08

fcoveram:
The above looks great to me ⭐

ryelle (commented):
Like the other PR, this is working well 👍🏻 

I did notice that on pages with short-content, there's no space under the chapter list. This screenshot is from [the Common APIs page](https://developer.wordpress.org/apis/). Maybe `sidebar-container` could get some bottom margin/padding?

Screenshot 2024-01-31 at 5 14 52 PM

adamwoodnz:
> I did notice that on pages with short-content, there's no space under the chapter list. This screenshot is from [the Common APIs page](https://developer.wordpress.org/apis/). Maybe `sidebar-container` could get some bottom margin/padding?

Another great find, fixed in [40f8977](https://github.com/WordPress/wporg-mu-plugins/pull/554/commits/40f897780e388f616b72383ec0f28b6e5aff487a)

Thanks for the solid review!

PR #471, opened 2024-01-15
Add 'Associated with' information to Explanation editor view

### LINKED ISSUE #436: Add 'Associated with' information to Explanation editor view
When the Explanations were using a classic block there was an 'Associated with` field at the top of the page, so the admin knew which code reference post the Explanation was for.

Image

After migrating all the Explanations to blocks this is gone. We need to add an entry in the block editor sidebar with this information.

See https://github.com/WordPress/wporg-developer/pull/430#issue-2040128903

outdoor2kode:
Close #436

This PR adds back the "Associated with" info to the Explanation block editor view, and also shows the post type explicitly for a better layout as well as for providing more knowledge.

| Class | Function | Hook |
|-|-|-|
| image | image | image |

**Test steps**

1. Go to Explanation and click on any post.
2. There should be "Associated with" on the side panel.

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/inc/explanations.php
+++ b/source/wp-content/themes/wporg-developer-2023/inc/explanations.php
@@ -51,7 +51,7 @@ class WPORG_Explanations {
 
 		// Admin.
 		add_action( 'edit_form_after_title',   array( $this, 'post_to_expl_controls'  )        );
-		add_action( 'edit_form_top',           array( $this, 'expl_to_post_controls'  )        );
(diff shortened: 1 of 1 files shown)


ryelle (approved):
Looks good, renders the connection in the sidebar 👍🏻 

Maybe a future improvement could be to move this into the "Summary" section? But that might be more work than is really necessary.

outdoor2kode:
I hadn't thought about placing it in the summary. I just checked, and it does look like there is additional work needed. I'll merge this first and try the idea out later.

PR #472, opened 2024-01-15
Fix "Improve it on GitHub" links don't go anywhere if the user is logged out.

### LINKED ISSUE #438: "Improve it on GitHub" links don't go anywhere if the user is logged out
The "Improve it on GitHub" links in the Block Editor Handbook are not correctly linking to the respective GitHub pages (https://developer.wordpress.org/block-editor/). 

image

outdoor2kode:
Fixes #438

This PR ensures that when a user clicks on [Improve it on GitHub](https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer-2023/inc/shortcodes.php#L37) and [See list of changes](https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer-2023/inc/shortcodes.php#L49), they are first checked for login status. If not logged in, they are redirected to the login page. Since both links utilize [get_edit_post_link()](https://github.com/WordPress/wordpress-develop/blob/6.4/src/wp-includes/link-template.php#L1474), this results in a situation where if not logged in, users will be redirected back to the current page after clicking them.

image
Example: http://localhost:8888/block-editor/

---

Additionally, I have removed the article-meta pattern from the single page and kept it only for the handbook, because the Date part reflects the time when the post was added/modified in the theme, not the actual time the content was modified. Furthermore, on pages other than the handbook, content is not edited through GitHub. cc @ndiego 

image
Example: http://localhost:8888/reference/hooks/get_edit_post_link/

## Note

Now, after logging in, users will be redirected back to the page where they clicked the link, instead of being directly taken to the target page (GitHub). The main reason is that `get_edit_post_link()` can only retrieve the link to the target page after logging in, so it cannot be directly used as the redirect parameter in `wp_login_url()`. Any thoughts on this? Perhaps we could modify the handbook page to not use `get_edit_post_link()` in the [shortcode function](https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer-2023/inc/shortcodes.php#L34-L39) but instead return the edit link directly, without considering whether the user is logged in or not, since editing on GitHub also requires a separate login and fork.

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/inc/shortcodes.php
+++ b/source/wp-content/themes/wporg-developer-2023/inc/shortcodes.php
@@ -27,14 +27,16 @@ add_shortcode(
 
 /**
  * Get the link to edit the page.
- *
- * The handbook code automatically swaps out the edit link for the github URL,
- * if this is a github handbook.
  */
 add_shortcode(
 	'article_edit_link',
 	function() {
-		return get_edit_post_link();
+		global $post;
+		$markdown_source = get_markdown_edit_link( $post->ID );
+		if ( $markdown_source ) {
+			return esc_url( $markdown_source );
+		}
+		return is_user_logged_in() ? get_edit_post_link() : wp_login_url( get_permalink() );
 	}
 );
 
@@ -46,9 +48,10 @@ add_shortcode(
 	function() {
 		// If this is a github page, use the edit URL to generate the
 		// commit history URL
-		$edit_url = get_edit_post_link();
-		if ( str_contains( $edit_url, 'github.com' ) ) {
-			return str_replace( '/edit/', '/commits/', $edit_url );
+		global $post;
+		$markdown_source = get_markdown_edit_link( $post->ID );
+		if ( str_contains( $markdown_source, 'github.com' ) ) {
+			return str_replace( '/edit/', '/commits/', $markdown_source );
 		}
 		return '#';
 	}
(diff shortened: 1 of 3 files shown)


ryelle:
> Perhaps we could modify the handbook page to not use get_edit_post_link() in the [shortcode function](https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer-2023/inc/shortcodes.php#L34-L39) but instead return the edit link directly, without considering whether the user is logged in or not, since editing on GitHub also requires a separate login and fork.

I think this makes more sense, especially if you're logged out on both services— you'd click "Improve", log into wporg, then click "Improve" again, only to get the github login.

adamwoodnz:
> Additionally, I have removed the article-meta pattern from the single page and kept it only for the handbook, because the Date part reflects the time when the post was added/modified in the theme, not the actual time the content was modified. Furthermore, on pages other than the handbook, content is not edited through GitHub. cc @ndiego

Ah yes, I think that will fix https://github.com/WordPress/wporg-developer/issues/473

outdoor2kode:
> I think this makes more sense, especially if you're logged out on both services— you'd click "Improve", log into wporg, then click "Improve" again, only to get the github login.

I've made the changes based on the idea.

ryelle (approved):
Looks good! Tested on github handbooks, non-github handbooks, and the code reference 👍🏻

ndiego:
Thanks all!

PR #475, opened 2024-01-16
Update UI of Methods section to the table layout used by Related.

### LINKED ISSUE #384: Update UI of Methods section to the table layout used by Related
See https://github.com/WordPress/wporg-developer/pull/369#pullrequestreview-1733426194

outdoor2kode:
Closes #384

This PR updates the UI of the Methods section to the table layout that's used by the Related section.
Dash is added for empty description. 
cc @WordPress/meta-design 

table

Do we want to show all the methods like it did previously, or only show a few (5 for the moment) and allow to expand?

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/code-methods/block.php
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-methods/block.php
@@ -74,7 +74,7 @@ function get_methods_content() {
 			$headings     = array( __( 'Name', 'wporg' ), __( 'Description', 'wporg' ) );
 
 			return sprintf(
-				'<!-- wp:wporg/code-table {"id":"uses","headings":%1$s,"rows":%2$s,"style":{"spacing":{"margin":{"top":"var:preset|spacing|20"}}}} /--> ',
+				'<!-- wp:wporg/code-table {"id":"uses","headings":%1$s,"rows":%2$s,"itemsToShow":100,"style":{"spacing":{"margin":{"top":"var:preset|spacing|20"}}}} /--> ',
 				wp_json_encode( $headings ),
 				wp_json_encode( get_row_data( $class_methods ) )
 			);
--- a/source/wp-content/themes/wporg-developer-2023/src/code-methods/block.php
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-methods/block.php
@@ -48,7 +48,7 @@ function render( $attributes, $content, $block ) {
 
 	$wrapper_attributes = get_block_wrapper_attributes();
 	return sprintf(
-		'<section %s>%s %s</section>',
+		'<section %1$s>%2$s %3$s</section>',
 		$wrapper_attributes,
 		$title_block,
 		do_blocks( $content )
--- a/source/wp-content/themes/wporg-developer-2023/src/code-methods/block.php
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-methods/block.php
@@ -111,5 +111,6 @@ function get_row_data( $class_methods ) {
 		);
 		wp_reset_postdata();
 	}
+
 	return $rows;
 }
--- a/source/wp-content/themes/wporg-developer-2023/src/code-methods/block.php
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-methods/block.php
@@ -35,7 +35,7 @@ function render( $attributes, $content, $block ) {
 		return '';
 	}
 
-	$content = get_methods_content();
+	$content = get_methods_content( $block->context['postId'] );
 
 	if ( empty( $content ) ) {
 		return '';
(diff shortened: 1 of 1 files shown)


jasmussen:
Seems fine to me as a small iteration, though I'd tend to think that editorially we should strive for every item having a description.

fcoveram:
I would show all methods.

The TOC on the right eases the navigation by showing headings and jumping between sections; it was thought for long pages. In addition, the link to expand the area is subtle and blends with the ones above.

outdoor2kode:
> I'd tend to think that editorially we should strive for every item having a description.

Thank you for raising this concern. The items missing descriptions are actually due to [a lack of function doc](https://github.com/WordPress/wordpress-develop/blob/6.4/src/wp-includes/atomlib.php#L272-L325). ([Example that has doc](https://github.com/WordPress/wordpress-develop/blob/6.4/src/wp-includes/atomlib.php#L108-L114)) It's from core, not a problem with the parser itself.

> I would show all methods. 

Since it needs to specify a number of items to show. I've made it show more rows now, at least 100 rows. I think there isn't a class that has that many methods.

source/wp-content/themes/wporg-developer-2023/src/code-methods/block.php
```diff
@@ -73,42 +71,46 @@ function get_methods_content( $post_id ) {
...
+ *
+ * @param WP_Post[] $class_methods
+ * @return array[]
+ */
+function get_row_data( $class_methods ) {
+	$rows = array();
+
+	foreach ( $class_methods as $method ) {
+		// Remove the filter that adds the code reference block to the content to avoid a possible infinite loop.
+		remove_filter( 'the_content', 'DevHub\filter_code_content', 4 );
```
ryelle, line 97: When is `the_content` filter run? Since you're using the excerpt below, I don't think it is.

source/wp-content/themes/wporg-developer-2023/src/code-methods/block.php
```diff
@@ -35,7 +35,7 @@ function render( $attributes, $content, $block ) {
 		return '';
 	}
 
-	$content = get_methods_content( $block->context['postId'] );
+	$content = get_methods_content();
```
ryelle: Why remove the `postId` context?

ryelle (commented):
Looks good to me, just some clarifying comments.

As for the 100 method limit, maybe go larger just to be safe— https://developer.wordpress.org/reference/classes/wp_customize_manager/ has 117, https://developer.wordpress.org/reference/classes/wp_xmlrpc_server/ has 106. I think those are the only two over 100, but they do exist. (I found these with some SQL queries in phpmyadmin)

Overall, I think the table layout is a big improvement 👍🏻

source/wp-content/themes/wporg-developer-2023/src/code-methods/block.php
outdoor2kode: I guess I removed it because either in the previous code or the current one, it's not used.
`get_the_ID(),` was used somehow. I have added it back and replaced `get_the_ID()` to make it consistent with the conditional check above.

source/wp-content/themes/wporg-developer-2023/src/code-methods/block.php
outdoor2kode: If not added, this results in an error.

```[18-Jan-2024 06:00:19 UTC] PHP Fatal error:  Allowed memory size of 536870912 bytes exhausted (tried to allocate 20480 bytes) in /var/www/html/wp-content/plugins/query-monitor/classes/Backtrace.php on line 137``` 

And when I changed `apply_filters( 'get_the_excerpt', $method->post_excerpt, $method )` to just `$method->post_excerpt`, everything worked fine again.

I checked which functions were hooked to `get_the_excerpt`, and noticed that `wp_trim_excerpt` is one of them. This function [triggers](https://github.com/WordPress/wordpress-develop/blob/6.4/src/wp-includes/formatting.php#L3992) 'the_content' filter, leading to an infinite loop.

outdoor2kode:
> As for the 100 method limit, maybe go larger just to be safe— https://developer.wordpress.org/reference/classes/wp_customize_manager/ has 117, https://developer.wordpress.org/reference/classes/wp_xmlrpc_server/ has 106. I think those are the only two over 100, but they do exist. (I found these with some SQL queries in phpmyadmin)

Thanks for checking this! I didn't realize I could look into the DB.
I've increased it to 150.

source/wp-content/themes/wporg-developer-2023/src/code-methods/block.php
ryelle: TIL! Thanks for explaining that.

source/wp-content/themes/wporg-developer-2023/src/code-methods/block.php
ryelle: Oh I see— it's a better practice to keep it and fix get_methods_content (like [you did here](https://github.com/WordPress/wporg-developer/pull/475/commits/ee99c7aca5e3a30a27ece53fe693a28d51feee36) 👍🏻 ), since blocks _could_ be used in a query loop, where the "current post" (in context) would be different than the global post. It's not how we're using these blocks here, so it's not critical, but good to be in the habit of doing.

ryelle (approved):
Changes look good 👍🏻

source/wp-content/themes/wporg-developer-2023/src/code-methods/block.php
outdoor2kode: Ah, I see. Thanks for clarifying this!

PR #486, opened 2024-01-25
Fix copy code text twice

### LINKED ISSUE #485: Copy code action copies twice
Migrated from [Trac](https://meta.trac.wordpress.org/ticket/7422)

> Hi there!
On your developer website there is a bug when you copy above code block, it actually copies it to clipboard twice.
I have tried it on this specific code block https://developer.wordpress.org/themes/template-files-section/page-template-files/#using-conditional-tags-in-page-templates but it also happened on some other block on same page, haven't tested it out on other pages.
I'm on macbook pro version 14.2.1 and Google Chrome version 120.0.6099.216
Let me know if you need any more info. Thanks!

outdoor2kode:
Fixes #485

The issue arises because users add an extra `<code>` tag before the `[php]` shortcode in their generated content, resulting in an additional layer of `<code>` tags. This causes `find( 'code' )` to select two elements, leading to the text being copied twice. `first()` is added here to fix the problem.
image

**How to test**
1. See if it only copies once on `themes/template-files-section/page-template-files/`. (Two layers of `<code>` tag)
2. Also check the source on the reference page as well `reference/functions/is_admin/`. (One layers of `<code>` tag)

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/js/function-reference.js
+++ b/source/wp-content/themes/wporg-developer-2023/js/function-reference.js
@@ -42,7 +42,9 @@ jQuery( function ( $ ) {
 		$copyButton.on( 'click', function ( event ) {
 			event.preventDefault();
 			clearTimeout( timeoutId );
-			const code = $element.find( 'code' ).text();
+			// first() is used here as sometimes there are additional <code> tags in user-generated content,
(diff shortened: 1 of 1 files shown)


adamwoodnz (approved):
LGTM, nice work! 🚀

PR #487, opened 2024-01-28
Add redirects for the Block Editor Handbook.

ndiego:
This PR corresponds to https://github.com/WordPress/gutenberg/pull/58358. 

Two pages in the Block Editor Handbook are being removed, and we need to implement redirects to the new sections. The [Generate Blocks with WP-CLI](https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/generate-blocks-with-wp-cli/) had been removed awhile ago, but the redirect was never created. So this PR also adds that.


ryelle (approved):
Looks good, the destination pages all exist already & redirects work as expected. Is this ready to merge, or does it need to wait on upstream content changes?

ndiego:
We can merge now. I will merge the associated PR in Gutenberg. Thanks!

PR #494, opened 2024-02-02
Tools: Update node, yarn, composer packages

ryelle:
This updates the node version & yarn packages to use the latest build scripts. `@wordpress/scripts` and `@wordpress/env` are the only packages used here, both have been updated. Since I was updating things, I also updated the composer lockfile.

I tested with my local env and a fresh install, the build works, the env works, and parsing still works (on the fresh install, my existing env is wonky).

**To test**

- Use `nvm install` or `nvm use` to switch to Node 20
- Run `yarn` to update the dependencies
- Run `yarn build` to build the theme
- Start up the env
- Everything should work as expected, you should _not_ have to reconfigure anything


outdoor2kode (commented):
I have to change [these two scripts](https://github.com/WordPress/wporg-developer/blob/trunk/package.json#L17-L18) to `"wp-env run cli bash env/setup.sh"` and `"wp-env run cli wp devhub parse --user_id=1"` to avoid this error:
```
OCI runtime exec failed: exec failed: unable to start container process: exec: "bash env/setup.sh": stat bash env/setup.sh: no such file or directory: unknown
```
Did you run into this?

Other than that, it works correctly on the fresh install as well as on my current local env provision when switching to Node 20 on this branch.

ryelle:
Good catch, I don't think I tried to run `setup`. I have run into that with other cli calls— something about passing the command in quotes doesn't work anymore. Removing the quotes entirely should also work: `"setup:wp": "wp-env run cli bash env/setup.sh"`

outdoor2kode:
Ah yeah, you're right. Just tested it. It also works after removing the quotes entirely.
And `parse` also needs updating as well.

outdoor2kode (approved):

PR #502, opened 2024-02-19
Block Editor Handbook: Add redirect for experiemtns package that no longer exists

t-hamano:
Originally reported in the Gutenberg repository: https://github.com/WordPress/gutenberg/issues/58560

The `@wordpress/experiments` package has been renamed to `@wordpress/private-apis` package in the following PR: [#47839](https://github.com/WordPress/gutenberg/pull/47839)

Perhaps due to this, there seem to be two pages in the Block Editor Handbook: a page before the rename and a page after the rename.

- https://developer.wordpress.org/block-editor/reference-guides/packages/packages-experiments/
- https://developer.wordpress.org/block-editor/reference-guides/packages/packages-private-apis/

This PR redirects access to a package reference page that no longer exists to a new package reference page.

After that, we'll probably need to remove the old page (`block-editor/reference-guides/packages/packages-experiments/`) from the WP dashboard.


adamwoodnz (approved):
👍 LGTM, tested in sandbox

PR #511, opened 2024-04-27
Convert Github repository link to a button with text

jonathanbossenger:
In the WP-CLI repository it was pointed out that the [GitHub repository icon link isn't super obvious](https://github.com/wp-cli/wp-cli/issues/5227) on the page.

This PR attempts to fix that, but wrapping the icon in a more familiar button layout, with the label "GitHub Repository".


jonathanbossenger:
Before:
wp-site-–-WP-CLI-Command-Developer-WordPress-org

After:
wp-site-–-WP-CLI-Command-Developer-WordPress-org-after

PR #512, opened 2024-05-01
Add redirect for the IAPI reference

ndiego:
The Interactivity API reference was relocated in https://github.com/WordPress/gutenberg/pull/61143. This PR adds a redirect from the old location to the new one. Once the redirect is in place, I will manually remove the old page.


adamwoodnz (approved):
Works for me ✨

adamwoodnz:
This has been deployed, but the redirect won't work until the old path 404s.

ndiego:
Thanks @adamwoodnz, I'll take care of that now.

Issue #514, opened 2024-05-07
Issues when running yarn setup:tools for the first time


juanmaguitar:
I'm getting the following error when doing `yarn setup:tools` for the first time after cloning the repo locally

```
  - Installing yoast/phpunit-polyfills (1.x-dev 3abb63e): Extracting archive
 43/43 [============================] 100%
In GitDownloader.php line 501:

  Failed to execute git checkout '693a5181a113fadb4fa6d768977cbf74e579069c' -- && git reset --har
  d '693a5181a113fadb4fa6d768977cbf74e579069c' --

  fatal: reference is not a tree: 693a5181a113fadb4fa6d768977cbf74e579069c

  It looks like the commit hash is not available in the repository, maybe the commit was removed
  from the branch? Run "composer update wporg/wporg-mu-plugins" to resolve this.


install [--prefer-source] [--prefer-dist] [--prefer-install PREFER-INSTALL] [--dry-run] [--dev] [--no-suggest] [--no-dev] [--no-autoloader] [--no-progress] [--no-install] [-v|vv|vvv|--verbose] [-o|--optimize-autoloader] [-a|--classmap-authoritative] [--apcu-autoloader] [--apcu-autoloader-prefix APCU-AUTOLOADER-PREFIX] [--ignore-platform-req IGNORE-PLATFORM-REQ] [--ignore-platform-reqs] [--] [<packages>...]

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

```

<details>
<summary>See full output when <code>yarn setup:tools</code></summary>
<pre>
yarn run v1.22.19
$ yarn && composer install && TEXTDOMAIN=wporg composer exec update-configs && composer --working-dir=./source/wp-content/plugins/phpdoc-parser install
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
warning " > @wordpress/scripts@27.1.0" has unmet peer dependency "@playwright/test@^1.39.0".
warning " > @wordpress/scripts@27.1.0" has unmet peer dependency "react@^18.0.0".
warning " > @wordpress/scripts@27.1.0" has unmet peer dependency "react-dom@^18.0.0".
warning "@wordpress/scripts > @wordpress/e2e-test-utils-playwright@0.18.0" has unmet peer dependency "@playwright/test@>=1".
[...]

PR #516, opened 2024-05-07
Update Setup instructions with prerequisites

juanmaguitar:
When running `yarn setup:tools` for the first time after cloning the repo locally, some errors may appear if `composer update` is not run first.

A specific PHP version is also a requirement that is not properly remarked on the `README.md`

See https://github.com/WordPress/wporg-developer/issues/514 

This PR propose a solution to tackle these issues and improve the contribution experience for first-timers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
 		"@wordpress/env": "9.2.0"
 	},
 	"scripts": {
-		"setup:tools": "yarn && composer install && TEXTDOMAIN=wporg composer exec update-configs && composer --working-dir=./source/wp-content/plugins/phpdoc-parser install",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -12,7 +12,7 @@
 		"@wordpress/env": "9.2.0"
 	},
 	"scripts": {
-		"setup:tools": "yarn && composer install && TEXTDOMAIN=wporg composer exec update-configs && composer --working-dir=./source/wp-content/plugins/phpdoc-parser install",
```
cbravobernal, line 15: Not critic, but you would be running this command twice
`TEXTDOMAIN=wporg composer exec update-configs`

juanmaguitar: Thanks Carlos! 
I'll remove this part from the PR

PR #517, opened 2024-05-08
Fix compatibility issues with shortcodes in blocks and Gutenberg 18.1+

### LINKED ISSUE #513: Block Editor Handbook: Improve it on Github links are broken
All links I've spot checked in the Block Editor handbook have broken links.

The `href`s are  appended with `[article_edit_link]`.

For example: https://developer.wordpress.org/block-editor/

The link goes to : https://developer.wordpress.org/block-editor/[article_edit_link]

Block_Editor_Handbook___Developer_WordPress_org

Note: requires Gutenberg 18.1+ to reproduce

adamwoodnz:
Fixes #513 

After https://github.com/WordPress/gutenberg/pull/60349, shortcodes passed via block attributes don't work.

Shortcodes in block markup in patterns do work, so this PR removes the custom block usage and puts the shortcodes directly in the pattern markup. The main reason for using these custom blocks was conditional rendering, which has been moved to the template level.

The result is a fair bit of markup duplication but the reduction of custom block usage and hidden complexity is probably positive.

### Testing

1. Ensure your Gutenberg version is 18.1+
2. Load the Plugins handbook. Only the First Published and Last Updated (if different from published) meta data should be displayed
3. Load the Advanced Admin handbook. The dates plus Edit on GitHub and Changelog meta data should be displayed. The GitHub links should include the article title for screen readers only.
4. Load the Block Editor handbook. The meta data should be the same as the Advanced Admin handbook, but without the Changelog link.
5. Test all the links


source/wp-content/themes/wporg-developer-2023/inc/block-hooks.php
```diff
@@ -52,33 +51,6 @@ function get_block_content_by_home_url( $block_content, $replacement_home_url =
...
-
-		if ( in_array( $post_type, $local_handbooks ) ) {
-			return '';
-		}
-
-		// The block editor handbook doesn't have a changelog.
-		// We only know it's the changelog because of the linkURL attribute.
-		if ( 'blocks-handbook' === $post_type && '[article_changelog_link]' === $block['attrs']['linkURL'] ) {
-			return '';
-		}
```
adamwoodnz, line 76: This logic replaced by the switch [here](https://github.com/WordPress/wporg-developer/pull/517/files#diff-1cf9dd600383a4c89ba09a878ca8691359b6f6dcece2b9c9b90f3de810f274a4R607)

source/wp-content/themes/wporg-developer-2023/inc/block-hooks.php
```diff
@@ -52,33 +51,6 @@ function get_block_content_by_home_url( $block_content, $replacement_home_url =
...
- */
-function filter_handbook_meta_link_block( $block_content, $block ) {
-	if ( 'wporg/handbook-meta-link' === $block['blockName'] ) {
-		// Not all handbooks come from GitHub.
-		$local_handbooks = array( 'apis-handbook', 'plugin-handbook', 'theme-handbook' );
-		$post_type = get_post_type();
-
-		if ( in_array( $post_type, $local_handbooks ) ) {
-			return '';
-		}
```
adamwoodnz, line 70: This logic replaced by reverting the removal of the `single-handbook-github.html` template

source/wp-content/themes/wporg-developer-2023/inc/shortcodes.php
```diff
@@ -57,6 +57,17 @@ function() {
...
 );
 
+/**
+ * Get the title of the article.
+ */
+add_shortcode(
+	'article_title',
+	function() {
+		global $post;
+		return $post->post_title;
```
ryelle: Maybe `get_the_title()` so that the title filters can be run? Not a strong suggestion.

source/wp-content/themes/wporg-developer-2023/patterns/article-meta-block-editor.php
```diff
@@ -0,0 +1,58 @@
+<?php
+/**
+ * Title: Article Meta for the Block Editor GitHub handbook
+ * Slug: wporg-developer-2023/article-meta-block-editor
+ * Inserter: no
+ */
+
+// get title fot the current post
+$title = get_the_title();
+
```
ryelle: Is this still needed?

ryelle (approved):
Links & dates all work, checked a few different handbook pages. I left two comments, but this shouldn't need a re-review to merge 👍🏻 

> The result is a fair bit of markup duplication but the reduction of custom block usage and hidden complexity is probably positive.

Yeah, it's too bad about the duplication, but this is probably closer to how core expects themes to behave, so hopefully less likely to break 🤞🏻

source/wp-content/themes/wporg-developer-2023/patterns/article-meta-block-editor.php
adamwoodnz: Oh nice catch, thanks

source/wp-content/themes/wporg-developer-2023/inc/shortcodes.php
adamwoodnz: I think I tried that but it didn't work, I'll take another look

PR #528, opened 2024-07-22
Set yarn version to v1 (classic)

sirreal:
Run the following command to ensure that the correct yarn version is
used when using the project on machines that have newer yarn versions
installed:

```sh
yarn set version classic
```

Without this change, newer yarn versions would not work well with this repo,
and would automatically attempt to upgrade the lockfile.

---

Alternatively, this project could switch to newer versions of yarn or recent versions of `npm`. Other WordPress projects are using `npm` and that may be preferable.

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,0 +1,1 @@
+yarnPath: .yarn/releases/yarn-classic.cjs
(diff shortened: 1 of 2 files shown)


ryelle (approved):
Thanks for the PR!

-----

> Alternatively, this project could switch to newer versions of yarn or recent versions of npm. Other WordPress projects are using npm and that may be preferable.

We have yarn baked in to a few places (and used on all our other projects) - i agree `npm` would be preferable, but it's a bigger change to do that, so this is a good middle step.

PR #529, opened 2024-07-22
Remove jQuery from user notes preview

cbravobernal:
## What

Remove jQuery from `user-notes-preview.js`

## Why

jQuery is known to inflate the bundle size of packages, and in most cases, it can be replaced with native language functionality.

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/js/user-notes-preview.js
+++ b/source/wp-content/themes/wporg-developer-2023/js/user-notes-preview.js
@@ -25,7 +25,7 @@
 		document.querySelector('#commentform .tablist').style.display = 'flex';
 	
 		if (textarea && preview && tabs.length > 0) {
-			// Append spinner to preview tab
+			// Append spinner to preview tab.
 			tabs[tabs.length - 1].parentNode.appendChild(spinner);
 	
 			previewContent = preview.querySelector('.preview-content');
@@ -42,7 +42,7 @@
 
 	function previewEvents() {
 		const commentFormComment = document.getElementById('comment-form-comment');
-		const tabContentHeight = commentFormComment.offsetHeight; // outerHeight equivalent in vanilla JS
+		const tabContentHeight = commentFormComment.offsetHeight;
 		tabs.forEach(tab => {
 			tab.addEventListener('keydown', handlePreviewEvent);
 			tab.addEventListener('click', handlePreviewEvent);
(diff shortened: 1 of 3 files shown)


source/wp-content/themes/wporg-developer-2023/js/user-notes-preview.js
```diff
@@ -42,86 +39,87 @@
...
 			processing = false;
 
-			// Make first child of the preview focusable
-			preview.children().first().attr( {
-				'tabindex': '0'
-			} );
-		} );
+			// Make first child of the preview focusable.
+			if ( preview.firstChild ) {
+				preview.firstChild.setAttribute( 'tabindex', '0' );
```
ryelle: I'm getting an error when I test this, it looks like `firstChild` is a text node, not an html element.

Screenshot 2024-07-25 at 3 52 51 PM

source/wp-content/themes/wporg-developer-2023/js/user-notes-preview.js
```diff
@@ -3,37 +3,34 @@
...
-		preview = $( '#comment-preview' );
-		tabs = $( '#commentform .tablist' ).find( 'a' );
-		spinner = $( '<span class="spinner" style="display:none;"></span>' );
-		text = '';
+		textarea = document.querySelector( '.comment-form textarea' );
+		preview = document.querySelector( '#comment-preview' );
+		tabs = document.querySelectorAll( '#commentform .tablist a' );
+		spinner = document.createElement( 'span' );
+		spinner.className = 'spinner';
+		spinner.style.display = 'none';
```
ryelle, line 19: This does create the spinner correctly, but apparently it was never styled when we moved to the block theme, so nothing is visible. Not a problem for this PR though.

source/wp-content/themes/wporg-developer-2023/js/user-notes-preview.js
sirreal: 👍 Looks like a place for [firstElementChild](https://developer.mozilla.org/en-US/docs/Web/API/Element/firstElementChild)

cbravobernal: Fixed in https://github.com/WordPress/wporg-developer/pull/529/commits/6762ec8d27a674f05e4517b910b0f2a2aa77515e

It's weird I didn't get the error until now 🤔

source/wp-content/themes/wporg-developer-2023/js/user-notes-preview.js
cbravobernal: Where can I check the old styles?

ryelle: I wouldn't copy over the old styles, but you could borrow [the loading animation from this CSS](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/screenshot-preview-block/src/style.scss#L51) (it's the loading state for images such as [the pattern & style variation screenshots on Themes](https://wordpress.org/themes/twentytwentyfour/)).

cbravobernal: The spinner didn't have any styled, an empty span was appearing. I added the style, but the time of the spinner is quite small.

https://github.com/user-attachments/assets/43312032-1cfb-42fa-ae86-efd20953b65b

PR #531, opened 2024-07-23
Dashicons: Implement with Interactivity API

sirreal:
Replace the [dashicon page](https://developer.wordpress.org/resource/dashicons/) shortcode with a dashicon page block.

I had to replace the shortcode with a block because `wpautop` was running on the shortcode and mangling Interactivity API HTML.

I have commented the problematic line and added a todo comment to re-enabled this when supported. ~This requires https://github.com/WordPress/wordpress-develop/pull/7075 to work correctly on the server. I expect that to be released in 6.6.2. To test this, you'll need to check out that PR and instruct wp-env to use that core version.~

I have not removed or touched the existing dashicon shortcode code for now, but there are opportunities for removing code if we go ahead with this.

Most of the page HTML is transformed from the shortcode in https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer-2023/inc/shortcode-dashicons.php.

## Testing

Load up the page in a local environment and confirm it behaves like prodoction. Note that these pages render a random icon in detail:

- https://developer.wordpress.org/resource/dashicons/
- http://localhost:8888/resource/dashicons/

"Permalinks" were rendered client site with a hash to the icon slug. This behavior has been changed to use query params so that the server can render the requested icon (hash is not available to the server). Client-side, if a fragment is found with no valid icon query param, it will be upgraded to a query param:

- http://localhost:8888/resource/dashicons/?icon=dashicons-id-alt

See the upgrade behavior, where these will behave the same (after client rewrites the URL):
- https://developer.wordpress.org/resource/dashicons/#id-alt
- http://localhost:8888/resource/dashicons/#id-alt

Filtering should work basically the same. Try using the filter bar. Non-alphanumeric characters are ignored from filtering.

The copy buttons work the same, but they are buttons instead of `a` elements. This seems more semantic as they open a prompt.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
 	"license": "GPL-2.0-or-later",
 	"private": true,
 	"dependencies": {
-		"@wordpress/scripts": "27.1.0"
+		"@wordpress/scripts": "28.3.0"
 	},
 	"devDependencies": {
 		"@wordpress/env": "9.2.0"
--- a/source/wp-content/themes/wporg-developer-2023/package.json
+++ b/source/wp-content/themes/wporg-developer-2023/package.json
@@ -7,7 +7,7 @@
 	"private": true,
 	"devDependencies": {
 		"@wordpress/eslint-plugin": "^17.7.0",
-		"@wordpress/scripts": "27.1.0"
+		"@wordpress/scripts": "28.3.0"
 	},
 	"eslintConfig": {
 		"extends": "../../../../.eslintrc.js"
--- a/source/wp-content/themes/wporg-developer-2023/package.json
+++ b/source/wp-content/themes/wporg-developer-2023/package.json
@@ -1,7 +1,7 @@
 {
-    "name": "wporg-developer-2023",
-    "version": "1.0.0",
-    "description": "Theme for WordPress Developer Resources",
+	"name": "wporg-developer-2023",
+	"version": "1.0.0",
+	"description": "Theme for WordPress Developer Resources",
 	"author": "WordPress.org",
 	"license": "GPL-2.0-or-later",
 	"private": true,
@@ -16,8 +16,8 @@
 		"extends": "../../../../.stylelintrc"
 	},
 	"scripts": {
-		"build": "wp-scripts build",
-		"start": "wp-scripts start",
+		"build": "wp-scripts build --experimental-modules",
+		"start": "wp-scripts start --experimental-modules",
 		"lint:js": "wp-scripts lint-js src",
 		"lint:css": "wp-scripts lint-style *.css src/**/*.scss",
 		"format": "wp-scripts format src"
--- a/source/wp-content/themes/wporg-developer-2023/package.json
+++ b/source/wp-content/themes/wporg-developer-2023/package.json
@@ -31,8 +31,8 @@
 		}
 	},
 	"scripts": {
-		"build": "wp-scripts build",
-		"start": "wp-scripts start",
+		"build": "wp-scripts build --experimental-modules",
+		"start": "wp-scripts start --experimental-modules",
 		"lint:js": "wp-scripts lint-js --config ../../../../eslint.config.js src",
 		"lint:css": "wp-scripts lint-style *.css src/**/*.scss",
 		"format": "wp-scripts format src"
--- a/source/wp-content/themes/wporg-developer-2023/src/dashicons-page/render.php
+++ b/source/wp-content/themes/wporg-developer-2023/src/dashicons-page/render.php
@@ -11,19 +11,19 @@ wp_enqueue_style(
 	filemtime( get_stylesheet_directory() . '/stylesheets/page-dashicons.css' )
 );
 
-$icons = array();
+$icons          = array();
 $icons_sections = array();
 
 foreach ( \DevHub_Dashicons::get_dashicons() as $section_group => $section ) {
 	$icon_section = array(
 		'label' => $section['label'],
-		'slug' => sanitize_title( $section_group ),
+		'slug'  => sanitize_title( $section_group ),
 		'icons' => array(),
 	);
 	foreach ( $section['icons'] as $k => $v ) {
-		$icons[$k] = $v;
+		$icons[$k]                 = $v;
 		$icons[$k]['sectionLabel'] = $section['label'];
-		$icon_section['icons'][] = $k;
+		$icon_section['icons'][]   = $k;
 	}
 	$icons_sections[] = $icon_section;
 }
@@ -37,8 +37,8 @@ wp_interactivity_config(
 	'wporg/developer/dashicons-page',
 	array(
 		'texts' => array(
-			'copyCss' => __( 'Copy this, then paste in your CSS :before selector.', 'wporg' ),
-			'copyHtml' =>  __( 'Copy this, then paste in your HTML.', 'wporg' ),
+			'copyCss'   => __( 'Copy this, then paste in your CSS :before selector.', 'wporg' ),
+			'copyHtml'  => __( 'Copy this, then paste in your HTML.', 'wporg' ),
 			'copyGlyph' => __( 'Copy this, then paste in your Photoshop textfield.', 'wporg' ),
 		),
 		'icons' => $icons,
@@ -53,16 +53,16 @@ wp_interactivity_state(
 		 *
 		 * All these "derived state" getters must be defined in view.js as well
 		 */
-		'iconClass' => function() {
+		'iconClass'           => function() {
 			return 'dashicons ' . wp_interactivity_get_context()['icon'];
 		},
 		'sectionAnchorTarget' => function() {
 			return 'icons-' . wp_interactivity_get_context()['section']['slug'];
 		},
-		'sectionAnchorHref' => function () {
+		'sectionAnchorHref'   => function () {
 			return '#icons-' . wp_interactivity_get_context()['section']['slug'];
 		},
-		'iconSectionLabel' => function() {
+		'iconSectionLabel'    => function() {
 			return wp_interactivity_get_context()['section']['label'];
 		},
 
@@ -72,17 +72,17 @@ wp_interactivity_state(
 		 * Expected in WordPress 6.6.2
 		 */
 		// 'eachIcon' => function() use ( $icons ) {
-		// 	return $icons[ wp_interactivity_get_context()['icon'] ];
+		//   return $icons[ wp_interactivity_get_context()['icon'] ];
 		// },
 
 		/*
 		 * END: Derived state
 		 */
 
-		'iconsSections' => $icons_sections,
-		'selectedIcon' => array( $selected_icon ),
-		'filter' => '',
-		'style' => '',
+		'iconsSections'       => $icons_sections,
+		'selectedIcon'        => array( $selected_icon ),
+		'filter'              => '',
+		'style'               => '',
 	)
 );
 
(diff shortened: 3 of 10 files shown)


source/wp-content/themes/wporg-developer-2023/package.json
```diff
@@ -16,8 +16,8 @@
 		"extends": "../../../../.stylelintrc"
 	},
 	"scripts": {
-		"build": "wp-scripts build",
-		"start": "wp-scripts start",
+		"build": "wp-scripts build --experimental-modules",
+		"start": "wp-scripts start --experimental-modules",
```
sirreal, line 20: I could pull this out to its own PR to upgrade the package and add this flag. I don't think it should affect anything else.

sirreal:
@ryelle I'd appreciate feedback on this PR, thank you!

package.json
```diff
@@ -6,7 +6,7 @@
 	"license": "GPL-2.0-or-later",
 	"private": true,
 	"dependencies": {
-		"@wordpress/scripts": "27.1.0"
+		"@wordpress/scripts": "28.3.0"
```
gziolo, line 9: I assume that the code is going to work with WordPress 6.7 because otherwise, it would fail with the missing JSX runtime script dependency.

source/wp-content/themes/wporg-developer-2023/src/dashicons-page/render.php
```diff
@@ -0,0 +1,262 @@
...
+	$icons_sections[] = $icon_section;
+}
+
+$selected_icon = $_GET['icon'] ?? '';
+if ( ! array_key_exists( $selected_icon, $icons ) ) {
+	$selected_icon = array_rand( $icons );
+}
+
+wp_interactivity_config(
+	'wporg/dashicons-page',
```
ryelle: When defining interactivity stores in themes, I've been naming them with the theme namespace too, just to differentiate from the shared interactive blocks in [wporg-mu-plugins](https://github.com/WordPress/wporg-mu-plugins/tree/trunk/mu-plugins/blocks). Would you mind updating the name to `wporg/developer/dashicons-page`?

source/wp-content/themes/wporg-developer-2023/src/dashicons-page/render.php
```diff
@@ -0,0 +1,262 @@
...
+namespace WordPressdotorg\Theme\Developer_2023\Block_Dashicons_Page;
+
+require_once dirname( dirname( __DIR__ ) ) . '/inc/dashicons.php';
+
+wp_enqueue_style(
+	'dashicons-page',
+	get_stylesheet_directory_uri() . '/stylesheets/page-dashicons.css',
+	array(),
+	get_stylesheet_directory() . '/stylesheets/page-dashicons.css'
+);
```
ryelle: You could also move this stylesheet into the block, to keep it all in one place— once this page is updated to use the block, the old shortcode & related code can be removed.

source/wp-content/themes/wporg-developer-2023/src/dashicons-page/render.php
```diff
@@ -0,0 +1,262 @@
...
+
+namespace WordPressdotorg\Theme\Developer_2023\Block_Dashicons_Page;
+
+require_once dirname( dirname( __DIR__ ) ) . '/inc/dashicons.php';
+
+wp_enqueue_style(
+	'dashicons-page',
+	get_stylesheet_directory_uri() . '/stylesheets/page-dashicons.css',
+	array(),
+	get_stylesheet_directory() . '/stylesheets/page-dashicons.css'
```
ryelle: This should be a version number, the previous code used the modified time:

```suggestion
	filemtime( get_stylesheet_directory() . '/stylesheets/page-dashicons.css' )
```

source/wp-content/themes/wporg-developer-2023/src/dashicons-page/render.php
```diff
@@ -0,0 +1,262 @@
...
+}
+
+wp_interactivity_config(
+	'wporg/dashicons-page',
+	array(
+		'texts' => array(
+			'copyCss' => _( 'Copy this, then paste in your CSS :before selector.' ),
+			'copyHtml' =>  _( 'Copy this, then paste in your HTML.' ),
+			'copyGlyph' => _( 'Copy this, then paste in your Photoshop textfield.' ),
+		),
```
ryelle: These strings should be double-underscore, and have a text domain: `__( 'Copy …', 'wporg' )`

ryelle (commented):
Overall this is working just as the current page 👍🏻  I appreciate that you've kept the old URLs working, and the server-side rendering of the icon param works nicely.

There are a handful of lint errors, I know you mentioned not having it set up locally — I don't know if this helps, but this is the output at the moment (ignoring some content escaping issues that exist on the shortcode too).

```
$ composer lint ./source/wp-content/themes/wporg-developer-2023/src/dashicons-page
> phpcs './source/wp-content/themes/wporg-developer-2023/src/dashicons-page'
.E 2 / 2 (100%)

FILE: ./source/wp-content/themes/wporg-developer-2023/src/dashicons-page/render.php
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
FOUND 11 ERRORS AFFECTING 9 LINES
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  26 | ERROR | [x] Array keys must be surrounded by spaces unless they contain a string or an integer. (WordPress.Arrays.ArrayKeySpacingRestrictions.NoSpacesAroundArrayKeys)
  27 | ERROR | [x] Array keys must be surrounded by spaces unless they contain a string or an integer. (WordPress.Arrays.ArrayKeySpacingRestrictions.NoSpacesAroundArrayKeys)
  33 | ERROR | [ ] $_GET data not unslashed before sanitization. Use wp_unslash() or similar (WordPress.Security.ValidatedSanitizedInput.MissingUnslash)
  33 | ERROR | [ ] Detected usage of a non-sanitized input variable: $_GET['icon'] (WordPress.Security.ValidatedSanitizedInput.InputNotSanitized)
  42 | ERROR | [ ] Found single-underscore "_()" function when double-underscore expected. (WordPress.WP.I18n.SingleUnderscoreGetTextFunction)
  43 | ERROR | [x] Expected 1 space after "=>"; 2 found (WordPress.WhiteSpace.OperatorSpacing.SpacingAfter)
  43 | ERROR | [ ] Found single-underscore "_()" function when double-underscore expected. (WordPress.WP.I18n.SingleUnderscoreGetTextFunction)
  44 | ERROR | [ ] Found single-underscore "_()" function when double-underscore expected. (WordPress.WP.I18n.SingleUnderscoreGetTextFunction)
  53 | ERROR | [ ] Empty line required before block comment (Squiz.Commenting.BlockComment.NoEmptyLineBefore)
  77 | ERROR | [x] Expected 1 space before comment text but found 2; use block comment if you need indentation (Squiz.Commenting.InlineComment.SpacingBefore)
 158 | ERROR | [x] Opening PHP tag must be on a line by itself (Squiz.PHP.EmbeddedPhp.ContentBeforeOpen)
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 5 MARKED SNIFF VIOLATIONS AUTOMATICALLY
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

package.json
sirreal: Good callout!

[The v28 changelog reads:](https://github.com/WordPress/gutenberg/blob/eea7183c6b156610cd3daa538e98a8b4b5388c2e/packages/scripts/CHANGELOG.md#2800-2024-05-31)

> **Breaking Changes**
> Note If you're using @wordpress/scripts for building JS scripts to target WordPress 6.5 or earlier, you should not upgrade to this version and continue using @wordpress/scripts@27.

I think the site uses latest WordPress, so if it's on WP 6.6, it should work if that comment is accurate. Do you know if it's WordPress 6.6 or 6.7 that will use the new JSX runtime?

@ryelle can you confirm the site runs on latest WordPress release (6.6.x at this time)?

source/wp-content/themes/wporg-developer-2023/src/dashicons-page/render.php
sirreal: I moved this into the block's `viewStyle`. As a follow-up when removing dashicons shortcode, the stylesheet can move into the block source.

source/wp-content/themes/wporg-developer-2023/src/dashicons-page/render.php
sirreal: I changed this for the store. I did not modify the block name, let me know if you'd like me to use another name for the block.

sirreal:
I've addressed feedback and fixed lints, this is ready for another review.

package.json
gziolo: Right, 6.6 is fine, too. I assume that it's sartisfied here. However, I wanted to ensure we don't see any surprises when deploying.

ryelle: WordPress.org runs trunk & (usually) the latest Gutenberg version (the exception is the Support Forums due to conflicts with [blocks-everywhere](https://github.com/Automattic/blocks-everywhere), so we used a pinned version there). So we should be set with this version change. It also worked fine on my sandbox, which matches production versions.

source/wp-content/themes/wporg-developer-2023/src/dashicons-page/render.php
ryelle: Nope, the block name is fine, just the stores. Thanks!

source/wp-content/themes/wporg-developer-2023/src/dashicons-page/view.js
```diff
@@ -0,0 +1,104 @@
...
+		get iconSectionLabel() {
+			return IAPI.getContext().section.label;
+		},
+
+		get eachIcon() {
+			return config.icons[ IAPI.getContext().icon ];
+		},
+
+		get style() {
+			const cleanedValue = state.filter.replace( /[^a-zA-Z0-9]/, '' );
```
ryelle: This only replaces the first matching instance, so `s.ite` works but `s.i.t.e` doesn't.

I would also suggest allowing `-`, since some icons use that as part of the name. For example, [menu-alt](https://developer.wordpress.org/resource/dashicons/?icon=dashicons-menu-alt) can't be found by searching "menu-alt".

```suggestion
			const cleanedValue = state.filter.replace( /[^a-zA-Z0-9-]/g, '' );
```

source/wp-content/themes/wporg-developer-2023/src/dashicons-page/render.php
```diff
@@ -0,0 +1,255 @@
...
+	);
+	foreach ( $section['icons'] as $k => $v ) {
+		$icons[ $k ]                 = $v;
+		$icons[ $k ]['sectionLabel'] = $section['label'];
+		$icon_section['icons'][]     = $k;
+	}
+	$icons_sections[] = $icon_section;
+}
+
+$selected_icon = $_GET['icon'] ?? '';
```
ryelle: This line flags some security lint issues, but it's OK since the value is matched against the icon list. I'd add a comment disabling the check to avoid those false errors.

```suggestion
$selected_icon = $_GET['icon'] ?? ''; // phpcs:ignore WordPress.Security.ValidatedSanitizedInput -- Value checked against icon safelist.
```

sirreal:
Abandoning although there was good work and potential here.

PR #533, opened 2024-07-25
User notes voting: Fix selector

cbravobernal:
Before this PR, AJAX requests were not working. I'm not sure if I will have the bandwidth to fix it, so, meanwhile, at least with this fix it will work.


cbravobernal:
Closed as I was able to remove also jQuery in https://github.com/WordPress/wporg-developer/pull/532

PR #535, opened 2024-07-26
Search filters: Implement with Interactivity API

sirreal:
A full implementation of the search-filters block in Interactivity API.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
 	"license": "GPL-2.0-or-later",
 	"private": true,
 	"dependencies": {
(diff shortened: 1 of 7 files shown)


sirreal:
This is fully working and could be merged. I don't think it's a significant improvement over the existing implementation.

An advantage: The PHP rendering does seem easier to understand to me.
A drawback: This introduces a dependency on Interactivity API. The block previously had no dependencies.

I'll close, but pinging @ryelle in case you think this is a change worth landing.

PR #537, opened 2024-07-29
Autocomplete: Use script module and modern approaches

sirreal:
- No jQuery
- Use Script Modules
- [Use Script Modules data passing](https://make.wordpress.org/core/2024/05/06/proposal-server-to-client-data-sharing-for-script-modules/)
- Use REST endpoint for request

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/inc/autocomplete.php
+++ b/source/wp-content/themes/wporg-developer-2023/inc/autocomplete.php
@@ -20,14 +20,37 @@ class DevHub_Search_Form_Autocomplete {
 	 * @access public
 	 */
 	public function init() {
-
-		add_action( 'wp_ajax_autocomplete', array( $this, 'autocomplete_data_update' ) );
-		add_action( 'wp_ajax_nopriv_autocomplete', array( $this, 'autocomplete_data_update' ) );
+		add_action(
+			'rest_api_init',
+			function () {
+				register_rest_route(
+					'wporg-developer/v1',
+					'/autocomplete',
+					array(
+						'methods'             => 'POST',
+						'callback'            => array( $this, 'autocomplete_rest_handler' ),
+						'permission_callback' => '__return_true',
+						'schema' => function () {
+							return array(
+								'html' => array(
+									'description' => 'The HTML to process.',
+									'type' => 'string',
+									'required' => true,
+								),
+								'quirksMode' => array(
+									'description' => 'Process the document in quirks mode.',
+									'type' => 'boolean',
+								),
+							);
+						},
+					)
+				);
+			}
+		);
 
 		// Enqueue scripts and styles.
 		add_action( 'wp_enqueue_scripts', array( $this, 'scripts_and_styles' ), 11 );
 	}
-
 
 	/**
 	 * Enqueues scripts and styles.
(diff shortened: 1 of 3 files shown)


source/wp-content/themes/wporg-developer-2023/inc/autocomplete.php
```diff
@@ -37,97 +60,95 @@ public function init() {
...
 
 		// Handbook searches don't have autocomplete.
-		if ( function_exists( 'wporg_is_handbook' ) && wporg_is_handbook() ) {
-			return;
-		}
-
+		/*if ( function_exists( 'wporg_is_handbook' ) && wporg_is_handbook() ) {*/
+		/*	return;*/
+		/*}*/
+		/**/
```
sirreal: Don't remove this.

```suggestion
		if ( function_exists( 'wporg_is_handbook' ) && wporg_is_handbook() ) {
			return;
		}

```

source/wp-content/themes/wporg-developer-2023/inc/autocomplete.php
```diff
@@ -20,15 +20,38 @@ public function __construct() {
...
+									'description' => 'The HTML to process.',
+									'type' => 'string',
+									'required' => true,
+								),
+								'quirksMode' => array(
+									'description' => 'Process the document in quirks mode.',
+									'type' => 'boolean',
+								),
+							);
+						},
```
sirreal: No schema.

```suggestion
```

sirreal:
Rebased onto current trunk (clean — trunk hasn't touched these files). Adversarial review (AI-assisted) found the following; capturing here before further work. Note that CI signal on this PR is meaningless: `lint:js` only covers `src/` (this code is in `js/`) and PHP linting is disabled, so everything below sails through green.

**Blockers**

- `js/autocomplete.js:142` — `form.elements.namedItem( 'post_type[]' )` is `null` on every page except `/search/` (only `patterns/search-content.php` gets the injected checkboxes via the `wporg-filtered-search-form` class). The `TypeError` fires after `processing = true` and outside the `try/finally`, so `processing` is wedged `true` forever — the first keystroke permanently kills autocomplete on nearly every page. Also: with exactly one matching element `namedItem` returns the element itself, which isn't iterable. Use `form.querySelectorAll( 'input[name="post_type[]"]:checked' )` and move `processing = true` inside the `try`.
- `js/autocomplete.js:93` — the `replace` callback calls `searchfield.val( text )` on a raw `HTMLInputElement` (leftover jQuery). Selecting a suggestion throws, so replace/navigate never runs. Should be `searchfield.value = text`.

**Major**

- `js/autocomplete.js:72,81` — multi-word filter port bug: `words = _input.split( ' ' ).length` but `_input` stays a string, so `_input[ i ]` indexes *characters*, not words. "custom post" matches any title containing `c` and `u`.
- `js/autocomplete.js:126-131` — double-instantiation race: input events during the dynamic-import round trip pass both the `!processing` and `!awesome` guards and construct a second `Awesomplete` on the same input (nested wrappers, duplicate listeners).
- `js/autocomplete.js:34,111` — the old `.always()` stale-query refetch is gone; `search` is assigned and never read. Typing during an in-flight request leaves stale suggestions with no self-correction.
- `inc/autocomplete.php:134` — empty result set returns `WP_REST_Response( null )` → client does `response.posts` on `null` → unhandled rejection on every no-match query. Return an empty object instead.
- `inc/autocomplete.php:110-119` — the handler intersects requested types with `get_parsed_post_types()`, so the old `command` post-type autocomplete path can never work (and command pages have no `post_type[]` checkboxes to send anyway).

**Minor**

- No REST `args` schema / request validation; malformed JSON bodies raise warnings on `$req_json['nonce']`/`['post_types']`.
- The nonce design defeats itself: `permission_callback => '__return_true'` + an anonymous nonce has zero CSRF value, but under full-page caching the baked-in nonce expires and every request 403s. Worth dropping the nonce and making this a cacheable GET.
- No fetch error handling — network failures/non-OK responses escape as unhandled rejections.
- Unconditional `el.focus()` after the async import steals focus if the user moved on.
- `js/awesomplete.min.js` is dead (zero references) — delete.
- Module IDs (`@wporg-developer/*`) predate the convention trunk established in #567 (`wporg-developer-php-code-snippet`) — worth aligning.
- Live region (`role=status`) is now created on first keystroke instead of page load; on-demand live regions are unreliably announced by some screen readers.

The script-module architecture itself checks out: registration/dynamic-import dependency shape, `script_module_data_{id}` usage (correctly anticipated the WP 6.7 API), old handle cleanup, and the Awesomplete class port (keyboard model, blur/mousedown ordering, no new XSS) are all fine.

sirreal:
Stale stream of work, abandoning.

PR #547, opened 2025-02-20
Handbooks: Add SCF

kraftbj:
See https://meta.trac.wordpress.org/ticket/7917
See https://github.com/WordPress/secure-custom-fields/issues/35
Adding SCF as a handbook for the developer site. Links to it will come after validating the import and cleaning up the initial doc set.


dd32 (approved):
Code looks fine, can we rename it to `import-sfc.php` though?

I know `rest-api.php` doesn't, but most others do.

PR #551, opened 2025-12-12
Update Composer packages causing failures

desrosj:
This updates the `wporg/wporg-mu-plugins` and `wporg/wporg-parent-2021` dependencies in Composer.

When running `yarn setup:tools` for the first time, the command will fail due to the pinned commit hashes being removed from these repositories.

It seems like this can happen through rebases, but it's not clear why this is happening for these two repositories.

This is a more targeted change than #549, which seems to run `composer update` without limiting the command to specific dependencies. I am only seeing these two repositories causing failures. The other dependencies should likely be upgraded more intentionally.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -2679,12 +2679,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/WordPress/wporg-mu-plugins.git",
-                "reference": "a79757142e75d5339bffb82de45f5746d5dbe390"
+                "reference": "823e5de983acee45d1512ad73fa33e97223521e4"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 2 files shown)


desrosj:
Looks to also be related to #514.

desrosj:
While working on something else, I noticed that build.yml has the following line:
`composer install || composer update wporg/*`

I've added this `||` command to the `setup:tools` script, which should solve the problem for any `wporg-*` repository.

dd32:
Another option is just to delete the composer.lock file..

> This updates the wporg/wporg-mu-plugins and wporg/wporg-parent-2021 dependencies in Composer.

These hashes will be deleted on every commit to those repo's, as the build branch is rewritten on every merge, as the latest `build` commit changes to be rebased on top of it.

PR #567, opened 2026-06-08
Render parsed DocBlock PHP snippets with php-snippet

adamziel:
Renders PHP examples parsed from DocBlocks as runnable WordPress Playground snippets in their original positions in code reference descriptions.

WordPress/phpdoc-parser#258 removes `php interactive` fenced examples from `long_description`, stores the structured snippets and setup Blueprints as post metadata, and leaves exact `<!-- wp-parser-code-snippet:N -->` placeholders behind. Without a theme-side renderer, those examples disappear after import. This PR replaces the placeholders with `<php-snippet>` elements, emits only the setup Blueprints they reference, preserves Blueprint JSON object shapes through post meta, and loads the web component only when a snippet is rendered. A snippet whose declared Blueprint is missing or malformed remains visible without a Run action. Metadata imported without a placeholder is appended rather than dropped.

A code reference page showing three runnable PHP examples with expected output

## Testing

Import parser data from WordPress/phpdoc-parser#258, then open a code reference page containing standalone, inline-Blueprint, and shared-Blueprint examples. Confirm each snippet appears where its fence was, valid snippets run and show their expected output, empty and nested-empty Blueprint objects remain JSON objects, and a snippet with an unusable Blueprint has no Run action.

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
@@ -114,13 +114,21 @@ function get_description_content( $post_id ) {
 	}
 
 	if ( $has_snippets ) {
-		enqueue_php_code_snippet_script();
+		wp_enqueue_script_module(
+			'wporg-developer-php-code-snippet',
+			PHP_CODE_SNIPPET_SCRIPT_URL,
+			array(),
+			null
+		);
 	}
 
 	// Emit the referenced setup Blueprint <script> tags once, up front.
 	foreach ( array_keys( $used_setup_blueprints ) as $name ) {
 		if ( isset( $setup_blueprints[ $name ] ) ) {
-			$output .= render_blueprint_script( get_setup_blueprint_id( $post_id, $name ), $setup_blueprints[ $name ] );
+			$output .= render_php_code_snippet_blueprint_script(
+				get_php_code_snippet_setup_blueprint_id( $post_id, $name ),
+				$setup_blueprints[ $name ]
+			);
 		}
 	}
 
@@ -167,18 +175,6 @@ function get_description_content( $post_id ) {
 }
 
 /**
- * Enqueue the web component that renders and runs PHP snippets.
- */
-function enqueue_php_code_snippet_script() {
-	wp_enqueue_script_module(
-		'wporg-developer-php-code-snippet',
-		PHP_CODE_SNIPPET_SCRIPT_URL,
-		array(),
-		null
-	);
-}
-
-/**
  * Render parser snippet placeholders in HTML comments.
  *
  * @param string $description      Description HTML.
@@ -190,18 +186,25 @@ function enqueue_php_code_snippet_script() {
  * @return string
  */
 function render_php_code_snippet_placeholders( $description, $post_id, $snippets, $setup_blueprints, &$used_blueprints, &$placed ) {
-	$processor = new Code_Snippet_Placeholder_Processor( $description );
+	$processor = new PHP_Code_Snippet_Placeholder_Processor( $description );
 
 	while ( $processor->next_token() ) {
 		if ( \WP_HTML_Tag_Processor::COMMENT_AS_HTML_COMMENT !== $processor->get_comment_type() ) {
 			continue;
 		}
 
-		$index = get_php_code_snippet_placeholder_index( $processor->get_modifiable_text() );
-		if ( null === $index ) {
+		$comment_text = $processor->get_modifiable_text();
+		$prefix       = ' wp-parser-code-snippet:';
+		if ( ! str_starts_with( $comment_text, $prefix ) || ! str_ends_with( $comment_text, ' ' ) ) {
 			continue;
 		}
 
+		$index = substr( $comment_text, strlen( $prefix ), -1 );
+		if ( ! ctype_digit( $index ) ) {
+			continue;
+		}
+
+		$index = (int) $index;
 		if ( ! isset( $snippets[ $index ] ) || ! is_array( $snippets[ $index ] ) ) {
 			$processor->replace_current_token( '' );
 			continue;
@@ -221,28 +224,6 @@ function render_php_code_snippet_placeholders( $description, $post_id, $snippets
 	}
 
 	return $processor->get_updated_html();
-}
-
-/**
- * Return the parser snippet placeholder index from exact HTML comment text.
- *
- * @param string $comment_text HTML comment text.
- * @return int|null
- */
-function get_php_code_snippet_placeholder_index( $comment_text ) {
-	$prefix = ' wp-parser-code-snippet:';
-	$suffix = ' ';
-
-	if ( ! str_starts_with( $comment_text, $prefix ) || ! str_ends_with( $comment_text, $suffix ) ) {
-		return null;
-	}
-
-	$index = substr( $comment_text, strlen( $prefix ), -strlen( $suffix ) );
-	if ( ! ctype_digit( $index ) ) {
-		return null;
-	}
-
-	return (int) $index;
 }
 
 /**
@@ -268,9 +249,15 @@ function render_php_code_snippet( $post_id, $index, $snippet, $setup_blueprints,
 		return '';
 	}
 
-	$code       = $snippet['code'];
+	$code      = $snippet['code'];
+	$post_slug = get_post_field( 'post_name', $post_id );
+	if ( ! $post_slug ) {
+		$post_slug = 'example';
+	}
+
+	$inline_blueprint_id = 'wporg-code-snippet-blueprint-' . absint( $post_id ) . '-' . absint( $index + 1 );
 	$attributes = array(
-		'name' => get_php_code_snippet_name( $post_id, $index ),
+		'name' => $post_slug . '-' . ( $index + 1 ) . '.php',
 	);
 
 	if ( array_key_exists( 'blueprint', $snippet ) ) {
@@ -279,10 +266,10 @@ function render_php_code_snippet( $post_id, $index, $snippet, $setup_blueprints,
 			isset( $setup_blueprints[ $snippet['blueprint'] ] ) &&
 			is_array( $setup_blueprints[ $snippet['blueprint'] ] )
 		) {
-			$attributes['blueprint']                 = get_setup_blueprint_id( $post_id, $snippet['blueprint'] );
+			$attributes['blueprint']                 = get_php_code_snippet_setup_blueprint_id( $post_id, $snippet['blueprint'] );
 			$used_blueprints[ $snippet['blueprint'] ] = true;
 		} elseif ( is_array( $snippet['blueprint'] ) ) {
-			$attributes['blueprint'] = get_inline_blueprint_id( $post_id, $index );
+			$attributes['blueprint'] = $inline_blueprint_id;
 		} else {
 			// Keep snippets with unusable setup visible without offering a Run action that cannot succeed.
 			$attributes['runnable'] = 'false';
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -108,3 +170,222 @@ function get_description_content( $post_id ) {
...
+	$output .= '</div>';
+
+	return $output;
+}
+
+/**
+ * Enqueue the web component that renders and runs PHP snippets.
+ */
+function enqueue_php_code_snippet_script() {
+	wp_enqueue_script(
```
sirreal: This should use [`wp_enqueue_script_module()`](https://developer.wordpress.org/reference/functions/wp_enqueue_script_module/).

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -108,3 +170,222 @@ function get_description_content( $post_id ) {
...
+	if ( 'wporg-developer-php-code-snippet' !== $handle ) {
+		return $tag;
+	}
+
+	return sprintf(
+		'<script type="module" src="%s"></script>' . "\n",
+		esc_url( $src )
+	);
+}
+
```
sirreal: This filter shouldn't be necessary.

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -4,7 +4,10 @@
 use function DevHub\get_description;
 use function DevHub\get_see_tags;
 
+const PHP_CODE_SNIPPET_SCRIPT_URL = 'https://playground.wordpress.net/php-code-snippet.js';
+
 add_action( 'init', __NAMESPACE__ . '\init' );
+add_filter( 'script_loader_tag', __NAMESPACE__ . '\add_php_code_snippet_script_type', 10, 3 );
```
sirreal: Shouldn't be necessary, the script should be registered as a module directly.

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -67,12 +70,71 @@ function get_description_content( $post_id ) {
...
+		function ( $matches ) use ( $post_id, $snippets, $setup_blueprints, &$used_setup_blueprints, &$placed ) {
+			$index = (int) $matches[1];
+			if ( ! isset( $snippets[ $index ] ) || ! is_array( $snippets[ $index ] ) ) {
+				return '';
+			}
+			$placed[ $index ] = true;
+			return render_php_code_snippet( $post_id, $index, $snippets[ $index ], $setup_blueprints, $used_setup_blueprints );
+		},
+		(string) $description
+	);
```
sirreal, line 94: This is probably fine in virtually all cases, but we could use the Tag Processor (we'll need a subclass) for this.

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -67,12 +70,71 @@ function get_description_content( $post_id ) {
...
+	}
+
+	$used_setup_blueprints = array();
+	$placed                = array();
+
+	// Render each snippet in place, where the parser left its placeholder
+	// ( <!-- wp-parser-code-snippet:N --> ), so snippets stay between the
+	// surrounding prose instead of collapsing to the end of the description.
+	$description = preg_replace_callback(
+		'/<!--\s*wp-parser-code-snippet:(\d+)\s*-->/',
```
sirreal: Seems like we can require this exact form, no need to allow broader whitespace:

```suggestion
		'/<!-- wp-parser-code-snippet:(\d+) -->/',
```

(Aside: I'd love to use processing instructions here `<?wp-parser-code-snippet N>`, added to HTML recently and supported in WordPress 7.1 HTML API, but since I'd like to land this _for_ 7.1 docs perhaps we shouldn't.)

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -108,3 +170,222 @@ function get_description_content( $post_id ) {
...
+ * Render a setup Blueprint script tag.
+ *
+ * @param string       $id        Script tag ID.
+ * @param array|string $blueprint Blueprint data.
+ * @return string
+ */
+function render_blueprint_script( $id, $blueprint ) {
+	if ( is_array( $blueprint ) ) {
+		$blueprint = wp_json_encode( $blueprint );
+	}
```
sirreal, line 328: This _has_ to be an array, doesn't it? [Use preferred JSON encode flags for script tags:](https://sirre.al/2025/08/06/safe-json-in-script-tags-how-not-to-break-a-site/)


```suggestion
	if ( is_array( $blueprint ) ) {
		$blueprint = wp_json_encode( $blueprint, JSON_HEX_TAG | JSON_UNESCAPED_SLASHES );
	}
```

If it's not an array, that seems like a failure condition and it's unclear what should be done with the dependent snippets.

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -108,3 +170,222 @@ function get_description_content( $post_id ) {
...
+
+	return sprintf(
+		'<script id="%s" type="application/json">%s</script>',
+		esc_attr( $id ),
+		escape_script_data( (string) $blueprint )
+	);
```
sirreal: [Use `wp_get_inline_script_tag()`](https://developer.wordpress.org/reference/functions/wp_get_inline_script_tag/):


```suggestion
	return wp_get_inline_script_tag( (string) $blueprint, array( 'type' => 'application/json' ) );
```

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -108,3 +170,222 @@ function get_description_content( $post_id ) {
...
+
+/**
+ * Escape content inside script tags while preserving ordinary PHP/HTML text.
+ *
+ * @param string $content Script tag content.
+ * @return string
+ */
+function escape_script_data( $content ) {
+	return str_ireplace( '</script', '<\/script', $content );
+}
```
sirreal: Let's make it clear this is only suitable for the PHP source we're embedding in script tags:


```suggestion

/**
 * Escape PHP source test for embedding in script tags.
 *
 * @param string $content Script tag content.
 * @return string
 */
function escape_php_source_script_contents( $content ) {
	return str_ireplace( '</script', '<\/script', $content );
}
```

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -108,3 +170,222 @@ function get_description_content( $post_id ) {
...
+	}
+
+	$output = '';
+
+	if ( isset( $attributes['blueprint'] ) && $attributes['blueprint'] === get_inline_blueprint_id( $post_id, $index ) ) {
+		$output .= render_blueprint_script( $attributes['blueprint'], $snippet['blueprint'] );
+	}
+
+	$output .= '<php-snippet' . render_php_code_snippet_attributes( $attributes ) . '>';
+	$output .= '<script type="application/x-php">' . escape_script_data( $code ) . '</script>';
```
sirreal: Use [`wp_get_inline_script_tag()`](https://developer.wordpress.org/reference/functions/wp_get_inline_script_tag/):

```suggestion
	$output .= wp_get_inline_script_tag( escape_php_source_script_contents( $code ), array( 'type' => 'application/x-php' ) );
```

Here we should encode the data because it can't be automatically encoded safely and could reject.

One bit of feedback would be to use `type="application/x-php+json"` instead. We know exactly how to encode and decode JSON so we can safely embed it in HTML. This same technique applies for the expected output as well.

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -108,3 +170,222 @@ function get_description_content( $post_id ) {
...
+
+	if ( array_key_exists( 'expected_output', $snippet ) ) {
+		$output .= '<script type="text/expected-output">' . escape_script_data( $snippet['expected_output'] ) . '</script>';
```
sirreal: We should use `wp_get_inline_script_tag()` here.

Again, I'd consider an `…+json` type which allows JSON encoding making it easy and clear how to escape.

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -108,3 +170,222 @@ function get_description_content( $post_id ) {
...
+		}
+	}
+
+	$output = '';
+
+	if ( isset( $attributes['blueprint'] ) && $attributes['blueprint'] === get_inline_blueprint_id( $post_id, $index ) ) {
+		$output .= render_blueprint_script( $attributes['blueprint'], $snippet['blueprint'] );
+	}
+
+	$output .= '<php-snippet' . render_php_code_snippet_attributes( $attributes ) . '>';
```
sirreal: Lets use the tag processor to set attributes. [The way Core does script tags is a good example.](https://github.com/WordPress/wordpress-develop/blob/6172b6687a36a070598ec8c2caff5d6eb32d8225/src/wp-includes/script-loader.php#L3026-L3049)

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -108,3 +170,222 @@ function get_description_content( $post_id ) {
...
+ */
+function render_php_code_snippet_attributes( $attributes ) {
+	$output = '';
+
+	foreach ( $attributes as $name => $value ) {
+		$output .= sprintf( ' %s="%s"', esc_attr( $name ), esc_attr( $value ) );
+	}
+
+	return $output;
+}
```
sirreal: Let's not stitch together HTML attribute spans like this (use tag processor like I mentioned in another comment).

sirreal (commented):
Awesome stuff here! I've left some feedback.

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
adamziel: Done in 076382d

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
adamziel: Done in 076382d

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
adamziel: Done in 076382d

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
adamziel: Good call! I'm adding support to Playground in https://github.com/WordPress/wordpress-playground/pull/4101, but it won't get deployed until next week.

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
adamziel: done

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -108,3 +165,270 @@ function get_description_content( $post_id ) {
...
+
+/**
+ * Return a stable script ID for a reusable setup Blueprint.
+ *
+ * @param int    $post_id Post ID.
+ * @param string $name    Blueprint name.
+ * @return string
+ */
+function get_setup_blueprint_id( $post_id, $name ) {
+	$slug = sanitize_title( $name );
```
adamziel: this one removes valid unicode sequences, I don't like it

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
adamziel: done

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
adamziel: good call! done

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
adamziel: Let's do it post 7.1!

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
sirreal: Nice 👍 We can follow-up to use `+php` after that's released in Playground.

sirreal (approved):
This is a nice step. I've been unable to manually test but I think this is a good change and safe because it requires other changes to have an impact on production. I understand you've verified it's working correctly.

PR #578, opened 2026-08-07
Add extension points for Code Reference source links

### LINKED ISSUE #577: Add extension points for Code Reference source links
## Problem

We are working on a WordPress Core pull-request preview that generates the Code Reference from an exact `wordpress-develop` commit and makes it available in WordPress Playground.

Source links in that preview must point to the code that was parsed. For a pull request, that means a URL such as:

```text
https://github.com/WordPress/wordpress-develop/blob/<exact-commit-sha>/src/<source-file>#L<start>-L<end>
```

The theme currently builds GitHub and Trac URLs using `get_current_version()`:

- `DevHub\get_github_source_file_link()` links to the corresponding GitHub version/tag.
- `DevHub\get_source_file_link()` links to the corresponding Core Trac tag.
- The `wporg/code-reference-source` block always renders both links.

That works on the production site, but it cannot accurately represent documentation generated from an arbitrary commit. In particular, Trac has no equivalent URL for a commit from a pull request or fork. Hiding or rewriting these links after rendering, with CSS or output-buffered HTML replacement, would be fragile.

## Desired changes

Please add narrowly scoped extension points that allow an environment to:

1. Filter the fully constructed URL returned by `DevHub\get_github_source_file_link()`.
2. Suppress the Trac link in the `wporg/code-reference-source` block before its markup is rendered.

The GitHub URL filter should receive enough context to identify the documented item, preferably its post ID in addition to the default URL. Filtering the helper itself is important because it is also used outside the source block, including Code Reference usage information.

The exact hook names are open to discussion. APIs along these lines would meet the need:

```php
$url = apply_filters(
	'wporg_developer_github_source_file_link',
	$url,
	$post_id
);
```

```php
$show_trac_link = apply_filters(
	'wporg_developer_show_trac_source_link',
	true,
	$post_id
);
```

The pull-request preview can then use a small MU plugin to replace the GitHub repository/ref with the repository and exact commit that were parsed, and disable the Trac link. The generated file path and line fragment should remain intact.

## Backward compatibility

Without any filters installed, the production Code Reference should render exactly as it does today. This request does not propose changing the production destination of either source link or removing Trac from the production site.

## Acceptance criteria

- The default GitHub and Trac source links are unchanged.
- A plugin can replace every URL produced by `get_github_source_file_link()` with an exact-commit URL.
- The GitHub URL filter receives the documented item’s post ID or equivalent context.
- A plugin can prevent the Code Reference source block from emitting Trac-link markup.
- The GitHub link continues to include the correct source path and line-number fragment.
- The extension points are covered by tests where practical.

This will let CI validate that representative class, method, function, hook, and filter pages all link back to the exact Core commit used to generate the preview.

sirreal:
Fixes #577

Useful for documentation previews on WordPress/wordpress-develop trunk and PRs. Allows for documentation links to point to the right place and to remove trac documentation links (which will be invalid, PR commits will only exist on GitHub).

See https://github.com/WordPress/wordpress-develop/pull/12954.

Two filters. Zero behavior change unless you install one.

**`devhub-github-source-file-link`** — filters the full URL returned by `DevHub\get_github_source_file_link()`, with the post ID. Runs before `esc_url()`, and also fires when there is no source file (empty string), so consumers see every call. The helper is filtered, not the block, because usage-info links go through the same helper. Filtering only the block would miss them.

**`devhub-show-trac-source-link`** — checked in the `wporg/code-reference-source` block before the Trac anchor is built. Return `false` and the markup is never emitted. No CSS hiding, no output-buffer rewriting — that approach is fragile and wrong. Trac simply has no URL for a commit from a fork, so a preview environment needs to drop the link, not disguise it.

One adjacent fix: both helpers read the line-number meta via `get_the_ID()` even when called with an explicit `$post_id`. That was a bug. They now use `$post_id`. Output is identical on singular pages, where the two are always equal today; it is correct now when they are not.

## Manual testing

Drop this in `wp-content/mu-plugins/source-link-test.php`:

```php
<?php
// Point GitHub source links at an exact commit, drop the Trac link.
add_filter(
	'devhub-github-source-file-link',
	function ( $url, $post_id ) {
		return preg_replace(
			'#^https://github\.com/WordPress/wordpress-develop/blob/[^/]+/#',
			'https://github.com/WordPress/wordpress-develop/blob/0123456789abcdef0123456789abcdef01234567/',
			$url
		);
	},
	10,
	2
);

add_filter( 'devhub-show-trac-source-link', '__return_false' );
```

Load any Code Reference page (function, method, hook, class). Expected: the GitHub link points at the fake commit SHA with the original path and `#L<start>-L<end>` fragment intact, in both the source block and the "Source:" usage-info line. "View on Trac" is gone. Remove the file and the page renders exactly as production does today.

## Verification

This repo has no PHP test infrastructure — CI lints JS/CSS only — so I verified end-to-end in a Playground `run-blueprint` boot with a fabricated `wp-parser-function` post (source-file term, line metas, stored source code):

- Defaults unchanged: `https://github.com/WordPress/wordpress-develop/blob/6.9/src/wp-includes/functions.php#L100-L150`, `https://core.trac.wordpress.org/browser/tags/6.9/src/wp-includes/functions.php#L100`; block renders all three buttons.
- Filtered: `https://github.com/fork/wordpress-develop/blob/0123456789abcdef0123456789abcdef01234567/src/wp-includes/functions.php#L100-L150` — path and fragment intact in the helper return value and the block markup.
- Both filters received the correct post IDs.
- Trac suppressed: the block drops the Trac link; a post without stored source code drops to "View all references" only.
- `get_source_file_link()` is untouched by the GitHub filter.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/code-source/block.php
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-source/block.php
@@ -75,23 +75,29 @@ function get_source_content( $post_id ) {
 	if ( ! empty( $source_file ) ) {
 		$source_code = post_type_has_source_code( $post_type ) ? get_source_code( $post_id ) : '';
 
-		$view_reference_button = sprintf(
+		$buttons = array();
+
+		$buttons[] = sprintf(
 			'<a href="%s">%s</a>',
 			esc_url( get_source_file_archive_link( $source_file ) ),
 			__( 'View all references', 'wporg' )
 		);
 
-		$view_trac_button = sprintf(
-			'<a href="%s">%s</a>',
-			esc_url( get_source_file_link( $post_id ) ),
-			__( 'View on Trac', 'wporg' )
-		);
+		/**
+		 * Filters whether to show the Trac source link for a parsed post type.
+		 *
+		 * @param bool $show_trac_link Whether to show the Trac source link. Default true.
+		 * @param int  $post_id        Post ID of the parsed function, method, hook, or class.
+		 */
+		$show_trac_link = apply_filters( 'wporg_developer_show_trac_source_link', true, $post_id );
 
-		$view_github_button = sprintf(
-			'<a href="%s">%s</a>',
-			esc_url( get_github_source_file_link( $post_id ) ),
-			__( 'View on GitHub', 'wporg' )
-		);
+		if ( $show_trac_link ) {
+			$buttons[] = sprintf(
+				'<a href="%s">%s</a>',
+				esc_url( get_source_file_link( $post_id ) ),
+				__( 'View on Trac', 'wporg' )
+			);
+		}
 
 		if ( ! empty( $source_code ) ) {
 			$output .= do_blocks(
@@ -103,10 +109,14 @@ function get_source_content( $post_id ) {
 				)
 			);
 
-			$output .= sprintf( '<p class="wporg-dot-link-list">%s</p>', implode( ' ', array( $view_reference_button, $view_trac_button, $view_github_button ) ) );
-		} else {
-			$output .= sprintf( '<p class="wporg-dot-link-list">%s</p>', implode( ' ', array( $view_reference_button, $view_trac_button ) ) );
+			$buttons[] = sprintf(
+				'<a href="%s">%s</a>',
+				esc_url( get_github_source_file_link( $post_id ) ),
+				__( 'View on GitHub', 'wporg' )
+			);
 		}
+
+		$output .= sprintf( '<p class="wporg-dot-link-list">%s</p>', implode( ' ', $buttons ) );
 	}
 
 	return $output;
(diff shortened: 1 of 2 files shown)


source/wp-content/themes/wporg-developer-2023/src/code-source/block.php
```diff
@@ -75,38 +75,48 @@ function get_source_content( $post_id ) {
...
-			esc_url( get_github_source_file_link( $post_id ) ),
-			__( 'View on GitHub', 'wporg' )
-		);
+		/**
+		 * Filters whether to show the Trac source link for a parsed post type.
+		 *
+		 * @param bool $show_trac_link Whether to show the Trac source link. Default true.
+		 * @param int  $post_id        Post ID of the parsed function, method, hook, or class.
+		 */
+		$show_trac_link = apply_filters( 'devhub-show-trac-source-link', true, $post_id );
```
dmsnell, line 92: if this were `apply_filters( 'devhub-show-button', true, 'trac', $post_id )` then it could also apply to the GitHub button, if that would be useful. then there wouldn’t need to be two filters.

PR #579, opened 2026-08-14
Load WordPress before interactive PHP snippets

sirreal:
Readers of Code Reference examples should see and edit only the code that documents the API, not the Playground-specific `require '/wordpress/wp-load.php';` boilerplate, while still running examples with WordPress loaded, including after editing them.

This wires the theme side of WordPress/wordpress-playground#4262:

- Prints one shared auto-prepend element from `wp_footer` on each reference page that renders snippets:
  ```html
  <script id="wporg-code-snippet-auto-prepend" type="application/x-php+json">"<?php require_once '/wordpress/wp-load.php';"</script>
  ```
  The payload is JSON-encoded with the `+json` script type, matching how the block already emits snippet code and expected output. `JSON_HEX_TAG` keeps `<` escaped so the payload stays inert in HTML. The code reference blocks render twice per request (once for the table of contents, once for the content), so the script is printed from a named `wp_footer` callback rather than inline; repeat registration is a no-op.
- Stamps `auto-prepend-script="#wporg-code-snippet-auto-prepend"` and `implicit-php-open-tag` on every `<php-snippet>` the code-description block emits, including placeholder-placed and appended-fallback snippets.
- Strips the `<?php` / `require '/wordpress/wp-load.php';` preamble from the two `class_list()` examples that shipped in WordPress 7.1, so they keep running under `implicit-php-open-tag`. WordPress/wordpress-develop#13267 removes the preamble from the docblocks themselves; the strip goes once the parser strips it or Core ships without it.

`auto-prepend-script` mirrors PHP's `auto_prepend_file` `php.ini` directive while taking a selector for an inert HTML script element. The selected script is a complete PHP file with its own opening tag. Parser-provided examples omit the opening tag, so `implicit-php-open-tag` adds it without displaying it in the editor. The shared ID is safe because the script is emitted only once per page.

Depends on WordPress/wordpress-playground#4262, deployed to playground.wordpress.net.

The Playground-handbook import path (`inc/import-playground.php`) is untouched. Those embeds reproduce Playground documentation examples authored for their own execution context.

## Testing

Import parser data containing interactive snippets, open a Code Reference page, and confirm the page source contains exactly one `wporg-code-snippet-auto-prepend` script, in the footer; every `<php-snippet>` carries the matching `auto-prepend-script` attribute and `implicit-php-open-tag`; and examples run with WordPress loaded while displaying no opening tag or `wp-load.php` boilerplate.

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
@@ -278,9 +278,9 @@ function render_php_code_snippet( $post_id, $index, $snippet, $setup_blueprints,
 
 	$inline_blueprint_id = get_php_code_snippet_blueprint_id( $post_id, 'inline:' . ( $index + 1 ) );
 	$attributes = array(
-		'name'                => $post_slug . '-' . ( $index + 1 ) . '.php',
-		'auto-prepend-script' => '#' . PHP_CODE_SNIPPET_AUTO_PREPEND_ID,
-		'php-fragment'        => '',
+		'name'                  => $post_slug . '-' . ( $index + 1 ) . '.php',
+		'auto-prepend-script'   => '#' . PHP_CODE_SNIPPET_AUTO_PREPEND_ID,
+		'implicit-php-open-tag' => '',
 	);
 
 	if ( array_key_exists( 'blueprint', $snippet ) ) {
--- a/source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
@@ -259,7 +259,16 @@ function render_php_code_snippet( $post_id, $index, $snippet, $setup_blueprints,
 		return '';
 	}
 
-	$code = wp_json_encode( $snippet['code'], JSON_HEX_TAG | JSON_UNESCAPED_SLASHES );
+	$code = $snippet['code'];
+
+	// The two class_list() examples in WordPress 7.1 load WordPress themselves.
+	// Remove once the parser strips this or Core ships without it.
+	$preamble = "<?php\nrequire '/wordpress/wp-load.php';\n";
+	if ( str_starts_with( $code, $preamble ) ) {
+		$code = substr( $code, strlen( $preamble ) );
+	}
+
+	$code = wp_json_encode( $code, JSON_HEX_TAG | JSON_UNESCAPED_SLASHES );
 	if ( ! is_string( $code ) ) {
 		return '';
 	}
--- a/source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
@@ -280,7 +280,7 @@ function render_php_code_snippet( $post_id, $index, $snippet, $setup_blueprints,
 	$attributes = array(
 		'name'                  => $post_slug . '-' . ( $index + 1 ) . '.php',
 		'auto-prepend-script'   => '#' . PHP_CODE_SNIPPET_AUTO_PREPEND_ID,
-		'implicit-php-open-tag' => '',
+		'implicit-php-open-tag' => true,
 	);
 
 	if ( array_key_exists( 'blueprint', $snippet ) ) {
--- a/source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
@@ -7,10 +7,10 @@ use function DevHub\get_see_tags;
 const PHP_CODE_SNIPPET_SCRIPT_URL = 'https://playground.wordpress.net/php-code-snippet.js';
 
 /**
- * ID of the shared WordPress run-before script tag. The script is identical
+ * ID of the shared WordPress auto-prepend script tag. The script is identical
  * for every snippet, so a single element per page serves them all.
  */
-const PHP_CODE_SNIPPET_RUN_BEFORE_ID = 'wporg-code-snippet-run-before';
+const PHP_CODE_SNIPPET_AUTO_PREPEND_ID = 'wporg-code-snippet-auto-prepend';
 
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -351,6 +356,28 @@ function render_php_code_snippet_blueprint_script( $id, $blueprint ) {
...
+ *
+ * @param int $post_id Post ID.
+ * @return string
+ */
+function render_php_code_snippet_bootstrap_script( $post_id ) {
+	return wp_get_inline_script_tag(
+		"<?php require_once '/wordpress/wp-load.php';",
+		array(
+			'id'   => get_php_code_snippet_bootstrap_id( $post_id ),
+			'type' => 'application/x-php',
```
sirreal: This should be `+json` encoded.

adamziel:
@sirreal I've implemented [the Playground part](https://github.com/WordPress/wordpress-playground/pull/4262) and used `auto-prepend-script` as the attribute name to mirror the `auto_prepend_file` php.ini setting. I've also added a `implicit-php-open-tag` boolean attribute and enabled it by default in the WP.org theme. When present, you may skip the opening `<?php` from the snippet. I also took the liberty of updating this PR.

What do you think?

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -268,7 +278,9 @@ function render_php_code_snippet( $post_id, $index, $snippet, $setup_blueprints,
 
 	$inline_blueprint_id = get_php_code_snippet_blueprint_id( $post_id, 'inline:' . ( $index + 1 ) );
 	$attributes = array(
-		'name' => $post_slug . '-' . ( $index + 1 ) . '.php',
+		'name'                  => $post_slug . '-' . ( $index + 1 ) . '.php',
+		'auto-prepend-script'   => '#' . PHP_CODE_SNIPPET_AUTO_PREPEND_ID,
+		'implicit-php-open-tag' => '',
```
sirreal: We can do the boolean attribute like this
```suggestion
		'implicit-php-open-tag' => true,
```

sirreal:
Love it @adamziel

adamziel:
Yay! The related Playground PR landed, I'll hold on with deploying playground.wordpress.net until WCUS ends though—it seems fine for our purposes here.

sirreal:
[I tested and it seems like the feature is live.](https://software.hixie.ch/utilities/js/live-dom-viewer/?%3C!DOCTYPE%20html%3E%0A%3C!DOCTYPE%20html%3E%0A%3Cscript%20type%3Dmodule%20src%3D%22https%3A%2F%2Fplayground.wordpress.net%2Fphp-code-snippet.js%22%3E%3C%2Fscript%3E%0A%3Cscript%20id%3D%22load-wordpress%22%20type%3D%22application%2Fx-php%22%3E%0A%3C%3Fphp%20require_once%20%27%2Fwordpress%2Fwp-load.php%27%3B%0Aecho%20%22initialized!%5Cn%22%3B%0A%3C%2Fscript%3E%0A%0A%3Cphp-snippet%0A%20%20%20%20name%3D%22site-title.php%22%0A%20%20%20%20auto-prepend-script%3D%22%23load-wordpress%22%0A%20%20%20%20implicit-php-open-tag%0A%3E%0A%20%20%20%20%3Cscript%20type%3D%22application%2Fx-php%22%3E%0Aecho%20get_bloginfo(%20%27name%27%20)%3B%0A%20%20%20%20%3C%2Fscript%3E%0A%3C%2Fphp-snippet%3E)

I'll do some final testing but I think this can land.

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -351,6 +363,44 @@ function render_php_code_snippet_blueprint_script( $id, $blueprint ) {
...
+ * Render the shared WordPress auto-prepend script tag for PHP code snippets.
+ *
+ * Snippets reference this element from their `auto-prepend-script` attribute
+ * so it runs before the visible example, mirroring PHP's auto_prepend_file
+ * directive. JSON encoding matches the snippet code payloads: JSON_HEX_TAG
+ * escapes `<` so the payload stays inert in HTML.
+ *
+ * Only the first call renders the tag, so a page with several code
+ * descriptions still holds a single element with this ID.
+ *
```
sirreal:
```suggestion
```

sirreal:
This doesn't work for me. I see the global bootstrap script being added to the block content but it's not present in the final, served HTML.

I'm debugging.

sirreal:
Table of contents renders the content - the script tag was produced correctly - but this render is thrown away because the ToC is not interested in all the content, only headings.

Then the content is actually rendered but the static variable indicates the script tag was already rendered, so it's not produced in the content that's actually rendered to the page.

sirreal:
Fixed in 69dd57399b9c22917c8d46e6d1394bc090260891.

The block now registers a named `wp_footer` callback instead of printing inline. The script is rendered once in the page where necessary.

Now with the committed 7.1 including `<?php require…`, we'll need to strip that. Otherwise we get this warning when running the snippet:

> &lt;php-snippet implicit-php-open-tag&gt; code must not start with a PHP opening tag.

sirreal:
d13131d873aaff3f0348d109f805b01ad62c3aca will strip the WordPress load code from snippets.

This is working across the 7.1 and 7.2+ expected syntax.

Here's this working on the existing site with the 7.1 docs:

https://github.com/user-attachments/assets/97f7540d-4caf-437a-8fa0-753cd846d33f

And here it is running against a 7.2+ expected snippet based on https://github.com/WordPress/wordpress-develop/pull/13267:


https://github.com/user-attachments/assets/07facc6a-822f-4626-a715-2a0d77890f78

**Before** this change, the current snippets looked like this, including the `<?php require '…';` inside the snippet text:



Screenshot 2026-08-25 at 16 31 32


---

This seems ready, I'll plan to merge soon.

PR #580, opened 2026-08-14
Live snippets: Load WordPress implicitly

adamziel:
Readers of Code Reference examples should see and edit only the code that documents the API—not the Playground-specific `require '/wordpress/wp-load.php';` boilerplate—while still running examples with WordPress loaded, including after editing them.

This wires the theme side of WordPress/wordpress-playground#4261:

- Emits one shared run-before element per reference page that renders snippets, alongside the existing setup Blueprint scripts:
  ```html
  <script id="wporg-code-snippet-run-before" type="application/x-php+json">"\u003C?php require_once '/wordpress/wp-load.php';"</script>
  ```
  The payload is JSON-encoded with the `+json` script type, matching how the block already emits snippet code and expected output. `JSON_HEX_TAG` keeps `<` escaped so the payload remains inert in HTML.
- Stamps `run-before="#wporg-code-snippet-run-before"` and `php-fragment` on every `<php-snippet>` emitted by the code-description block, including placeholder-placed and appended-fallback snippets.

The shared script is a complete PHP file with its own opening tag. Parser-provided examples remain fragments, so `php-fragment` adds their opening tag without displaying it in the editor.

This depends on WordPress/wordpress-playground#4262 and should land after that component change.

The Playground-handbook import path (`inc/import-playground.php`) remains untouched. Those embeds reproduce Playground documentation examples authored for their own execution context.

## Testing

Import parser data containing interactive snippets, open a Code Reference page, and confirm the page source contains exactly one `wporg-code-snippet-run-before` script; every `<php-snippet>` carries the matching `run-before` attribute and `php-fragment`; and examples run with WordPress loaded while displaying no opening tag or `wp-load.php` boilerplate.

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
@@ -7,10 +7,10 @@ use function DevHub\get_see_tags;
 const PHP_CODE_SNIPPET_SCRIPT_URL = 'https://playground.wordpress.net/php-code-snippet.js';
 
 /**
- * ID of the shared WordPress bootstrap script tag. The bootstrap is identical
+ * ID of the shared WordPress run-before script tag. The script is identical
  * for every snippet, so a single element per page serves them all.
  */
-const PHP_CODE_SNIPPET_BOOTSTRAP_ID = 'wporg-code-snippet-bootstrap';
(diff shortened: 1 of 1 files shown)


adamziel:
Closing in favor of #579, which now includes the `run-before` and `php-fragment` updates.

PR #581, opened 2026-08-26
Interactive snippets: render the snippet source as a code block

sirreal:
Add a `<pre><code>` fallback inside `<php-snippet>` so the source is readable before the Playground module loads, or when JavaScript is unavailable. The element's shadow root hides the fallback after initialization.

Depends on #595. Without it, `wpautop()` encloses snippet placeholders in paragraphs. The fallback's `<pre>` start tag closes the paragraph, causing the HTML parser to place the fallback outside `<php-snippet>`.

`render_php_snippet()` builds the markup from one template using the HTML API. `set_modifiable_text()` preserves literal character references such as `&lt;egg&gt;` in the displayed source. The expected-output script is omitted when no output is specified; an empty string remains valid expected output.

Validation:

- Sandbox HTML for `WP_HTML_Tag_Processor::class_list()` has the fallback inside `<php-snippet>`, with text matching the JSON source.
- Four snippet placements were checked in a browser with the Playground module loaded and blocked, along with escaping round trips and nested content filters.
- PHP syntax, whitespace checks, and 15 template-rendering cases passed, covering source and attribute escaping and absent versus empty expected output.

Implemented with Claude Code and reviewed by me; template refactor and description edits with Codex.

### DIFF
--- a/source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
+++ b/source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
@@ -268,8 +268,8 @@ function render_php_code_snippet( $post_id, $index, $snippet, $setup_blueprints,
 		$code = substr( $code, strlen( $preamble ) );
 	}
 
-	$code = wp_json_encode( $code, JSON_HEX_TAG | JSON_UNESCAPED_SLASHES );
-	if ( ! is_string( $code ) ) {
+	$json_encoded_code = wp_json_encode( $code, JSON_HEX_TAG | JSON_UNESCAPED_SLASHES );
+	if ( ! is_string( $json_encoded_code ) ) {
 		return '';
 	}
 
@@ -318,9 +318,17 @@ function render_php_code_snippet( $post_id, $index, $snippet, $setup_blueprints,
 		$output .= render_php_code_snippet_blueprint_script( $attributes['blueprint'], $snippet['blueprint'] );
 	}
 
-	$snippet_output = '<php-snippet>';
+	/*
+	 * A `<pre>` start tag closes an open `<p>`, and the description renders
+	 * snippets inside one: `wpautop()` puts a bare placeholder comment in a
+	 * paragraph. Without a wrapper the parser carries the code block alone out
+	 * of the paragraph, tearing it off the element it belongs to. A `<div>`
+	 * closes the paragraph the same way but takes the whole snippet with it,
+	 * so the code block stays where it was written.
+	 */
+	$snippet_output = '<div class="wporg-php-snippet"><php-snippet>';
 	$snippet_output .= wp_get_inline_script_tag(
-		$code,
+		$json_encoded_code,
 		array( 'type' => 'application/x-php+json' )
 	);
 
@@ -331,7 +339,8 @@ function render_php_code_snippet( $post_id, $index, $snippet, $setup_blueprints,
 		);
 	}
 
-	$snippet_output .= '</php-snippet>';
+	$snippet_output .= render_php_code_snippet_source( $code );
+	$snippet_output .= '</php-snippet></div>';
 
 	$tags = new \WP_HTML_Tag_Processor( $snippet_output );
 	if ( $tags->next_tag( 'php-snippet' ) ) {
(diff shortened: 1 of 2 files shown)


source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -331,6 +331,7 @@ function render_php_code_snippet( $post_id, $index, $snippet, $setup_blueprints,
 		);
 	}
 
+	$snippet_output .= render_php_code_snippet_source( $code );
 	$snippet_output .= '</php-snippet>';
```
sirreal: We're still interpolating HTML here. We should work with a single template and use HTML API functionality (set_modifiable_text, set_attribute…) to build the snippet.

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -312,38 +300,63 @@ function render_php_code_snippet( $post_id, $index, $snippet, $setup_blueprints,
...
+ * @return string
+ */
+function render_php_snippet( $code, $attributes, $expected_output = null ) {
+	$html = new PHP_Code_Snippet_Processor(
+		<<<'HTML'
+		<php-snippet>
+			<script type="application/x-php+json"></script>
+			<script type="text/expected-output+json"></script>
+			<pre class="wp-block-code"><code class="language-php">placeholder</code></pre>
+		</php-snippet>
```
sirreal, line 332: We don't need the expected output if there's no expected output.

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -312,38 +300,63 @@ function render_php_code_snippet( $post_id, $index, $snippet, $setup_blueprints,
...
+			if ( $is_expected_output && null === $expected_output ) {
+				$html->replace_current_token( '' );
+				continue;
+			}
 
-	$tags = new \WP_HTML_Tag_Processor( $snippet_output );
-	if ( $tags->next_tag( 'php-snippet' ) ) {
-		foreach ( $attributes as $name => $value ) {
-			$tags->set_attribute( $name, $value );
+			$json = wp_json_encode( $is_expected_output ? $expected_output : $code, JSON_HEX_TAG | JSON_UNESCAPED_SLASHES );
```
sirreal: Add unescaped unicode and unescaped line terminators.

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -312,38 +300,63 @@ function render_php_code_snippet( $post_id, $index, $snippet, $setup_blueprints,
...
+	);
 
-	$snippet_output .= '</php-snippet>';
+	while ( $html->next_tag() ) {
+		if ( 'PHP-SNIPPET' === $html->get_tag() ) {
+			foreach ( $attributes as $name => $value ) {
+				$html->set_attribute( $name, $value );
+			}
+		} elseif ( 'SCRIPT' === $html->get_tag() ) {
+			$is_expected_output = 'text/expected-output+json' === $html->get_attribute( 'type' );
```
sirreal: This handles the code script and the expected output script. That's a bit awkward but I guess it works.

adamziel: Not awkward at all, it reads very natural to me. The only way to make it better I can think of is @dmsnell's templating engine idea.

source/wp-content/themes/wporg-developer-2023/src/code-description/block.php
```diff
@@ -312,38 +300,63 @@ function render_php_code_snippet( $post_id, $index, $snippet, $setup_blueprints,
...
+ * @param array       $attributes      Snippet element attributes.
+ * @param string|null $expected_output Expected output, or null when unspecified.
+ * @return string
+ */
+function render_php_snippet( $code, $attributes, $expected_output = null ) {
+	$html = new PHP_Code_Snippet_Placeholder_Processor(
+		<<<'HTML'
+		<php-snippet>
+			<script type="application/x-php+json"></script>
+			<script type="text/expected-output+json"></script>
```
sirreal: This gives up `wp_get_inline_script_tag()` which may include some useful behaviors and is filterable. We should construct the HTML using tags that provides.

adamziel (approved):

sirreal:
Follow-up, not for this PR: print the Blueprint `<script>` tags from `wp_footer` instead of inside the content.

The code-description block renders inside `the_content` at priority 4, so its output goes through `do_shortcode` and the other content filters. The snippet elements have to stay in the content, and their string payloads are escaped for that. The Blueprint scripts do not need to be there: the element resolves `blueprint` lazily by id over the whole document, and the auto-prepend script already prints from `wp_footer` for the same reason. Register each Blueprint by id when the block renders and print them once in the footer. That takes them out of the content filters and off the Markdown output pages.

dmsnell:
Screenshot 2026-09-23 at 9 05 07 PM

thank you for restoring the non-JS page renders!

dmsnell:
actually @sirreal it seems like the expected output has been stripped away now, and we only see the code snippet. that is unfortunate. the screenshot I left should have `// Outputs: "free <egg> lang-en "` following the code.

the inline docs inside the IDE also are still pretty broken. because of the un-indenting the example is still harder to separate from the explanatory text, and the IDE rendering treats it as inline code rather than a code block.

Screenshot 2026-09-23 at 9 07 41 PM

this did work before the interactive code example changes.

sirreal:
> the screenshot I left should have `// Outputs: "free <egg> lang-en "` following the code.

Remember that the live site uses the 7.1 version of the code, where the expected output was in its own code fence. We could create another PR to add that as well, but it should be fixed in 7.2 by changing the example format to use the trailing comment style.

dmsnell:
@sirreal why is another code fence pruned from the output? A number of examples have that, I believe.

Also the IDE issue is pretty important too. That’s not just PHPStorm, but also the alignment in the code. I know this is a work in progress, but it’s ideal to get this stuff fixed before this gets used everywhere and all the docblocks break.

PR #590, opened 2026-09-14
Neutralize shortcode syntax in the buffered notes list

mcliwanow:
## Why

The Code Reference comments block buffers the rendered User Contributed Notes and returns them into `the_content` at priority 4 (`filter_code_content`), so core parses that buffer a second time on the way out: `do_blocks` at priority 9 and `do_shortcode` at priority 11. #582 handled the first of those by neutralising block delimiters in the buffer. Brackets were left alone, so a note containing shortcode-shaped text was expanded by the site-wide shortcode registry at render time, even though the note's own `comment_text` pass (`DevHub_User_Submitted_Content::do_note_shortcodes()`) only ever expands `code`, `php`, `js` and `css`.

Example: a note with `[caption width=1 caption="x"]see example[/caption]` renders a `<figure>` with a `<figcaption>` on trunk, where the note's own pass had left that text as-is.

## What changed

- `src/code-comments/block.php` encodes `[` and `]` in the buffered note list as `&#91;` and `&#93;`, right after the block-delimiter step, and the comment above it now names both parsers.
- Notes whose text contains shortcode syntax other than the four note shortcodes now render that text literally instead of expanding it. Code blocks are unaffected: by the time the buffer is built `do_note_shortcodes()` has already expanded `[code]`, `[php]`, `[js]` and `[css]`, so any bracket still present is literal text (including brackets inside the code itself, e.g. `$a[0]`, which now carry the entity in the HTML and render the same).

**Smaller things:** `DevHub_User_Submitted_Content::do_note_shortcodes()` now snapshots `$shortcode_tags`, narrows it to the four note shortcodes, and puts the snapshot back in a `finally`. The old loop only re-added the tags it had removed, so a throwing callback, or anything that changed the registry mid-pass, left it in a different state than it found it.

## Testing

The repo has no PHP test suite for the theme, so this was checked by hand in wp-env against trunk and this branch in the same session, with the web container restarted between the two so opcache could not serve the other copy. Four approved notes on `reference/functions/wp_insert_post/`, by a user with no role on the site:

| note | trunk | branch |
|---|---|---|
| `[caption width=1 caption="CAPMARKER-A"]see example[/caption]` | 1 `wp-caption-text` figcaption | 0, the bracketed text shows literally |
| `[code]echo $a[0] . "CODEMARKER-B";[/code]` | code block | code block, identical on screen |
| `Literal brackets [BRACKMARKER-C] and array[0]` | literal | literal |
| `[php]$x = "PHPMARKER-D";[/php]` | code block | code block |

Logged-in pass, same branch, as the test account with a note on the page: the per-note feedback editor, textarea, quicktags toolbar and hidden fields render inside the section, the served section contains no `<script>` and no JSON, and a feedback reply containing `[BRACKET-FB]` and `$a[1]` submitted through the form is stored with the brackets and renders with them after approval.

Also checked that no theme JavaScript reads the note code blocks' `innerHTML` (the highlighter reads text content, which decodes the entities), and that the second consumer of `filter_code_content`, the `wporg/table-of-contents` block, only regexes headings out of the content and is not used on reference templates anyway.

Lint: `phpcs` on both files with the WordPress standard reports nothing on the changed lines (the pre-existing docblock and short-array findings on untouched lines remain).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bor0:
Reviewed. This looks right to me — the buffer reaches the later `the_content` filters with no literal `[` left, and I walked the filters that run in between to confirm none of them decode numeric entities. The two in the theme that do rewrite entities, `fix_double_encoding()` and `DevHub_CLI::fix_formatting()`, are both gated on post type and don't run here.

Three things, none blocking:

**Restoring the registry.** The `finally` is a good change. Since you're in there, a whole-snapshot swap is exact where the current re-add is partial — if anything reached during `do_shortcode()` removes one of the four allowed tags or registers a new one, the loop doesn't put things back the way it found them:

```php
$allowed  = array( 'code', 'php', 'js', 'css' );
$original = $GLOBALS['shortcode_tags'];

$GLOBALS['shortcode_tags'] = array_intersect_key( $original, array_flip( $allowed ) );

try {
	$text = do_shortcode( $text );
} finally {
	$GLOBALS['shortcode_tags'] = $original;
}
```

Shorter, and drops both loops.

**One more pass on the test matrix.** Every measurement in the table is an anonymous request, but the buffer isn't the same shape when a note-capable user is logged in — `wporg_developer_list_notes()` adds a `wp_editor()` instance, the quicktags container and the hidden fields per note, and all of that is now inside the `str_replace`. I think it's fine (`tinymce => false`, and the editor's settings JS prints in the footer, outside `ob_get_clean()`), so this is confirm-don't-suspect: load the page logged in as the test account with a reply editor open and check the form still renders and a note with a `[` in it still submits.

That implies a constraint worth writing down next to the `str_replace`: nothing in this buffer may emit inline JS or a JSON island, since that's the one context where the entities wouldn't decode. Nothing does today, and there's no test that would catch it if that changed.

**Follow-up, not for this PR.** The render callback is now carrying a hand-maintained list of parser syntaxes to neutralize — delimiters in #582, brackets here — and that list is only as complete as whatever parses `the_content` today. The durable version is to stop round-tripping rendered output through the content parsers at all: leave a placeholder at priority 4 and substitute the rendered notes after both parsers have run. That'd also move the buffer out from under `wpautop`/`wptexturize`, which is a rendering change and needs its own testing pass, so it wants its own issue. Your own results hint at it — the curly quotes in the branch column are `wptexturize` working on note text that's now correctly left alone.

Also worth an issue: there's no regression guard for any of this. A single assertion that `render()`'s output contains no raw `[` would freeze the invariant, and this surface has now been touched three times with hand-verification each time.

Detail on the ordering argument and the filter walk is in the tracking issue.

mcliwanow:
Thanks. Both taken in `c04e377`:

- `do_note_shortcodes()` now snapshots `$shortcode_tags`, narrows it with `array_intersect_key()`, and puts the snapshot back in the `finally`. Both loops are gone. The two assignments carry a `phpcs:ignore` for `WordPress.WP.GlobalVariablesOverride`, since overriding that global is the point. There's also a line noting that the narrowing is global for the duration of the pass, so `do_blocks()` calls made from the allowed callbacks see the same four tags.
- The constraint is stated in the comment above the `str_replace`: nothing in the buffer may emit inline JavaScript or a JSON island.

Logged-in pass done on the branch, as the test account with a note on the page: the feedback editor, its textarea, the quicktags toolbar (4 buttons) and the hidden fields all render inside the section, the served section has no `<script>` and no JSON, no console errors, and a feedback reply containing `[BRACKET-FB]` and `$a[1]` submitted through the form, was stored with the brackets intact, and renders with them after approval. Added to the Testing section.

The round-trip follow-up and the regression guard I'll leave for separate issues rather than grow this one.

bor0:
Re-reviewed at `c04e377` — read the whole diff again rather than just the delta. **LGTM, no blocking findings.**

The registry swap is right, and I checked it for defects rather than just matching it against what I'd suggested:

- `$original = $GLOBALS['shortcode_tags']` is a value copy even though the global is reference-marked (every core `global $shortcode_tags;` puts that zval in a reference set, and assigning a plain variable from a reference dereferences and copies). So `$original` can't be aliased by anything `do_shortcode()` does, and the restore is a genuine restore.
- Writing through `$GLOBALS[...]` reaches the same zval core's `add_shortcode()`/`remove_shortcode()` mutate, so the narrowing is visible to core exactly as the old `remove_shortcode()` loop was. Nothing is lost by bypassing the wrappers — neither fires a hook, and `add_shortcode()`'s only extra work is `_doing_it_wrong()` validation that doesn't apply to putting tags back.
- `array_intersect_key` preserves ordering, `get_shortcode_regex()` rebuilds from `array_keys()` every call with no cache, so there's no stale-regex path.
- Both `phpcs:ignore` annotations name the right sniff, and it's the pattern itself rather than a workaround.

One honest note, since I'm the one who asked for this change: the snapshot is exact in one direction and lossy in the other. Where the old loop *preserved* a shortcode registered during the pass, the snapshot *discards* it. Both versions have a mirror-image flaw and this is the better one to have — but I checked whether anything actually hits it, and every `add_shortcode()` in the theme runs at file load or on `init`, well before any `do_shortcode()`: six in `inc/shortcodes.php`, four in `DevHub_Formatting::do_init()`, one in `DevHub_Playground_Importer::init()`, one in `inc/shortcode-dashicons.php`. Twelve, none lazy. No regression. Worth remembering this function exists if some unrelated shortcode ever goes missing on a reference page.

The logged-in verification is better than what I asked for, in the way that matters — slicing the served HTML from an in-page `fetch()` measures the bytes as sent rather than the decoded DOM, which is the only measurement that can tell "the entity is there" apart from "the browser already turned it back into a bracket." Zero `<script>` and zero JSON in the section confirms the constraint the new comment states against the real logged-in buffer, not by reading `wp_editor()`'s arguments. And the submit-approve-render round trip covers the path end to end.

Re-checked the parts of my last review the new commit could have invalidated — statement ordering in `render()`, the post-type gates on the two entity-rewriting filters, the `do_blocks()` reachability the new comment documents. All still hold.

The two follow-ups are unchanged and belong in their own issues, not here. Nothing should hold this merge.

bor0 (approved):
Approved at `c04e377`.

Both review items were taken and the verification gap closed. The registry swap holds up on a defect-level read — the snapshot is a true value copy despite the global being reference-marked, so the `finally` restore is a real restore — and the logged-in check measures the served bytes rather than the decoded DOM, which is the only measurement that actually proves the invariant.

Detail in the two review comments above. The two follow-ups (ending the round trip through the content parsers, and a regression assertion over `render()`'s output) belong in their own issues and shouldn't hold this.

